PR #98, opened 2023-11-01
python publish workflow

Carbon225:


djstrong (changes requested):
It is in subdirectory

djstrong:
Add also authors to toml

Carbon225:
@djstrong ok?

djstrong (approved):
we need to try

PR #112, opened 2023-11-07
mock get_nft_metadata responses

djstrong:


djstrong:
Now all tests run 6 s instead of 90 s - it is good for development. However, I think that non-mock tests should be run in Github Actions. What do you think?

api/README.md
```diff
@@ -89,11 +89,17 @@ To enable this, you will need to set an environment variable
...
 ```
 
 ### Using the AWS Lambda handler
 
 NameGuard includes a handler for [Amazon AWS Lambda](https://aws.amazon.com/lambda/). It is available in the `nameguard.lambda` module. You can use it to create a Lambda function that will respond to HTTP requests. It uses the [mangum](https://mangum.io) library.
 
 Check out the included [Dockerfile](./Dockerfile) for an example of how to build a Lambda container image.
+
+### Disable monkeypatch tests
+
```
Carbon225, line 102: explain?

djstrong: done

Carbon225 (approved):

PR #126, opened 2023-11-13
Error 404 page

Y3drk:


djstrong:
This button doesn't work.

notrab (approved):

apps/website/src/app/components/Footer.tsx
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import { TwitterIcon } from "@/app/components/icons/TwitterIcon";
+import { GithubHoverableLogo } from "@/app/components/icons/GithubHoverableLogo";
+
+export const Footer = () => {
+  return (
+    <footer className="z-10 bg-white relative bot-0 w-full h-fit flex flex-col gap-3 py-5 items-center self-stretch gt_mobile:flex-row gt_mobile:justify-between gt_mobile:py-6 gt_mobile:px-28 gt_mobile:self-stretch border-t border-gray-200">
+      <p className="not-italic font-normal text-gray-500 text-sm gt_mobile:text-footer_text gt_mobile:font-light">
+        © NameHash Labs. All Rights Reserved
```
notrab, line 9:
```suggestion
        &copy; NameHash Labs. All Rights Reserved
```

apps/website/src/app/components/icons/TwitterIcon.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+export const TwitterIcon = () => {
```
notrab, line 3: Let's update all of the icons to allow props so we can spread on `className` instead of hardcoding them here.

We can do this in a separate PR that touches all icons.

apps/website/src/app/layout.tsx
```diff
@@ -48,7 +52,12 @@ export default function RootLayout({
 }) {
   return (
     <html lang="en">
-      <body className={inter.variable}>{children}</body>
+      <body className={inter.variable}>
+        <Header />
+        {children}
+        <Footer />
+        <NameGuardWrapper />
```
notrab, line 59: I added this here so that the header/footer works across all pages as well as the `SearchModal`/`SettingsModal`

apps/website/tailwind.config.ts
```diff
@@ -36,9 +36,11 @@ const config: Config = {
           "radial-gradient(50% 50% at 50% 50%, rgba(255, 255, 255, 0.8) 0%, #FFF 100%), linear-gradient(0deg, #FFAF00 0%, #F112D9 32.29%, #4C3FA0 70.83%, #2ED3C6 95.83%)",
         in_touch_background:
           "radial-gradient(50% 50% at 50% 50%, rgba(255, 255, 255, 0.6) 0%, #FFF 100%), linear-gradient(0deg, #4C3FA0 0%, #F112D9 32.29%, #2ED3C6 58.85%, #FFAF00 100%)",
+        error404_background:
```
notrab, line 39: We could probably move these to inline at some point. Non-blocking for this PR though.

apps/website/src/app/components/NameGuardWrapper.tsx
```diff
@@ -0,0 +1,12 @@
+"use client";
+
+import { SearchModal, SettingsModal } from "@namehash/nameguard-react";
+
+export function NameGuardWrapper() {
```
notrab, line 5: Right now the NameGuard React pacakge relies on a headlessui float dependency that has to be `use client` so I created this. Importing it as dynamic with ssr disabled didn't resolve it.

I hope to remove the dependency for the Tooltip in another PR as well so we can reduce the number of dependencies needed.

apps/website/src/app/components/Footer.tsx
Y3drk: Will do that in other PR today, just to not forget about it.

apps/website/src/app/components/icons/TwitterIcon.tsx
Y3drk: Yes, definitely, but I'd do it either when we need to reuse them in slightly other configuration or near the end of things, when functionalities are ready

apps/website/tailwind.config.ts
Y3drk: Yes. 
I think a good moment would be when I'll reverse the prop drilling and make all LP components static like you suggested at some point. 
I kept saving those in the config because I thought they were the same all over that page and it would be easy to reuse them just as a green one and a purple one, but they seem to differ slightly each time I put them in a new component and that pretty much defeats the point of doing it.

PR #132, opened 2023-11-14
Figma adjustments + minor fixes

Y3drk:


apps/website/src/app/atoms/icons/EmailIcon.tsx
```diff
@@ -1,4 +1,4 @@
-import React from "../../../../node_modules/react";
+import * as React from "react";
```
notrab, line 1: We're in Next.js context so we could probably remove this line altogether (across all files)

PR #142, opened 2023-11-17
run unmocked tests only on merge

djstrong:


Byczong (approved):

apohllo:
This change runs the tests only when the PR was merged. To my understanding this better works for things like automatic deploy. But here we want to prevent the merge if the optional tests are not passing. We could do that e.g. by providing checking the commit message:

https://stackoverflow.com/questions/63619329/github-action-get-commit-message

So I we are ready for the merge we could add a message stating that the branch is ready, so the additional checks would be run.

There's also a way to run tests if the PR is selected for merge, but this requires a **merge queue** which is available for the enterprise plan only.

djstrong:
We can just leave it to run always.

djstrong:
We decided to run unmocked tests always.

PR #145, opened 2023-11-17
fake eth name checkpoint with provided fields

djstrong:


api/nameguard/nameguard.py
```diff
@@ -350,6 +351,14 @@ async def secure_primary_name(self, address: str, network_name: str) -> SecurePr
                                        primary_name_status=status,
                                        nameguard_result=nameguard_result)
 
+    async def fake_eth_name_check_fields(self, network_name, contract_address, token_id, investigated_fields: dict[str,str]) -> FakeEthNameCheckResult:
+        contract_address = contract_address.lower()
+        
+        # title = investigated_fields['title']  # mandatory if ENS contract; ENS name
+        
+        #TODO: return only impersonating fields
```
djstrong, line 359: @lightwalker-eth For now, we have been returning all fields obtained from Alchemy. It doesn't make sense in the new endpoint because we would have to return the same dict as provided by client. We can return only fields which are impersonating ETH name. Do we want different response for these endpoints?

api/nameguard/nameguard.py
```diff
@@ -350,6 +351,14 @@ async def secure_primary_name(self, address: str, network_name: str) -> SecurePr
                                        primary_name_status=status,
                                        nameguard_result=nameguard_result)
 
+    async def fake_eth_name_check_fields(self, network_name, contract_address, token_id, investigated_fields: dict[str,str]) -> FakeEthNameCheckResult:
+        contract_address = contract_address.lower()
+        
+        # title = investigated_fields['title']  # mandatory if ENS contract; ENS name
```
djstrong, line 357: @lightwalker-eth In the new endpoint a client needs to provide ETH name under `title` key of `investigated_fields` dictionary if contract_address is ENS, because we return NameGuard report.

djstrong: @lightwalker-eth It generates a problem when we want to return fields which are faking ENS. If user also send `fields` dict with `title` key.

api/nameguard/nameguard.py
djstrong: So we will return only fields which are faking ENS names. However, we break the loop if the first is faking field encountered so there always be at most only one key in response.

api/nameguard/web_api.py
```diff
@@ -273,6 +273,38 @@ async def fake_eth_name_check_get(
...
 
+class FakeETHNameCheckFieldsRequest(BaseModel):
+    network_name: NetworkName
+    contract_address: str = Path(examples=['0x495f947276749ce646f68ac8c248420045cb7b5e'],
+                                 description='Contract address for the NFT contract (ERC721 and ERC1155 supported).')
+    token_id: str = Path(examples=['61995921128521442959106650131462633744885269624153038309795231243542768648193'],
+                         description='The ID of the token (in hex or decimal format).')
+    title: str = Field(description='Title of the NFT. For ENS names it have to be the name.')
+    fields: dict[str, str] = Field(
+        description='Fields with values which will be investigated (e.g. title, collection name, metadata) whether they look like fake .eth ENS name. Do not use `title` key.')
```
djstrong, line 284: What to do if user provides `title` as key of `fields`?

api/tests/test_api.py
```diff
@@ -657,18 +732,267 @@ def test_fake_eth_name_check(test_client, api_version, contract_address, token_i
...
+          'title': 'dot.this.averageman.eth'},
+         FakeEthNameCheckStatus.AUTHENTIC_ETH_NAME),
+        # NameWrapper https://rarible.com/token/0xd4416b13d2b3a9abae7acd5d6c2bbdbe25686401:34762977820481521209114130776556072772965907316729597364642457029530388725237
+        # pytest.param('0x495f947276749ce646f68ac8c248420045cb7b5e',
+        #              '12254154752654038808579574986506730093213650301129873673484321000757900869633',
+        #              FakeEthNameCheckStatus.IMPERSONATED_ETH_NAME,
+        #              marks=pytest.mark.xfail(reason='also delisted on opensea')),
+        # https://rarible.com/token/0x495f947276749ce646f68ac8c248420045cb7b5e:12254154752654038808579574986506730093213650301129873673484321000757900869633
+        ('0xd4416b13d2b3a9abae7acd5d6c2bbdbe25686401',
+         '0x37bf77d30d63cbf9ddad6b3c161522c53dcdcd8177b6177c83835c5ea69a7f8f',
```
djstrong, line 898: This token_id does not match the name. Should we check this or remove token_id as argument?

api/nameguard/web_api.py
```diff
@@ -273,6 +273,38 @@ async def fake_eth_name_check_get(
     token_id = validate_token_id(token_id)
     return await ng.fake_eth_name_check(network_name=network_name, contract_address=contract_address, token_id=token_id)
 
+class FakeETHNameCheckFieldsRequest(BaseModel):
+    network_name: NetworkName
+    contract_address: str = Path(examples=['0x495f947276749ce646f68ac8c248420045cb7b5e'],
+                                 description='Contract address for the NFT contract (ERC721 and ERC1155 supported).')
+    token_id: str = Path(examples=['61995921128521442959106650131462633744885269624153038309795231243542768648193'],
```
djstrong, line 280: We are not using token_id in the new endpoint. We can use it to check if it matches ENS name.

api/nameguard/nameguard.py
```diff
@@ -381,26 +390,45 @@ async def fake_eth_name_check(self, network_name, contract_address, token_id) ->
...
+        return await self._fake_eth_name_check(network_name, contract_address, title, investigated_fields)
+        
+    async def _fake_eth_name_check(self, network_name, contract_address, title: str, fields: dict[str,str]) -> FakeEthNameCheckResult:
+        investigated_fields = {'title': title}
+        if 'title' in fields: del fields['title']
+        investigated_fields.update(fields)
+        
         
         if contract_address in ens_contract_adresses:
             if title == '':  # the name has never been registered
```
djstrong, line 411: This was Alchemy thing

api/nameguard/nameguard.py
```diff
@@ -350,6 +351,14 @@ async def secure_primary_name(self, address: str, network_name: str) -> SecurePr
                                        primary_name_status=status,
                                        nameguard_result=nameguard_result)
 
+    async def fake_eth_name_check_fields(self, network_name, contract_address, token_id, title: str, investigated_fields: dict[str,str]) -> FakeEthNameCheckResult:
```
djstrong, line 354: New endpoint does not return UNKNOWN_NFT

api/nameguard/nameguard.py
djstrong: Sure, we can, but ens_cure is quite expensive (n^2).

api/nameguard/nameguard.py
djstrong: What should empty title (name) mean for ENS contracts? Now we return UNKNOWN_NFT.

api/nameguard/nameguard.py
```diff
@@ -381,39 +389,69 @@ async def fake_eth_name_check(self, network_name, contract_address, token_id) ->
...
+                investigated_fields['title'] = unknown_name
+        
+       
+        return await self._fake_eth_name_check(network_name, contract_address, investigated_fields)
         
+    async def _fake_eth_name_check(self, network_name, contract_address, fields: dict[str,str]) -> FakeEthNameCheckResult:
+
         if contract_address in ens_contract_adresses:
+            if 'title' not in fields:
+                raise Exception('title not in fields') #TODO
```
djstrong, line 405: Checking existence of `title` key only for ENS contracts.

api/nameguard/nameguard.py
djstrong: Ok.

api/nameguard/nameguard.py
```diff
@@ -381,39 +387,70 @@ async def fake_eth_name_check(self, network_name, contract_address, token_id) ->
                 investigated_fields['.'.join(keys)] = name
             except KeyError:
                 pass
+
+        if contract_address in ens_contract_adresses:
+            if ALCHEMY_UNKNOWN_NAME.match(title):
+                unknown_name = f"[{res_json['id']['tokenId'][2:]}].eth"
+                investigated_fields['title'] = unknown_name
+            elif title == '': # the name has never been registered
```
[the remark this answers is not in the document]
djstrong: Yes, that's how Alchemy is saying that.

PR #154, opened 2023-11-23
refine messages

djstrong:


api/nameguard/checks/label/punycode.py
```diff
@@ -5,14 +5,14 @@
 
 STATUS = CheckStatus.WARN
 
-L_MESSAGE_PASS = 'This label is Punycode compatible'
-N_MESSAGE_PASS = 'All labels in this name are Punycode compatible'
+L_MESSAGE_PASS =  'This label is Punycode compatible'
```
djstrong, line 8: Here needs to stay word `label` because we have different check for punycode name compatibility.

PR #164, opened 2023-11-25
optimized bulk inspect names

djstrong:


api/nameguard/nameguard.py
```diff
@@ -120,6 +139,9 @@ async def inspect_name(self, network_name: NetworkName, name: str, resolve_label
         if resolve_labelhashes:
             name = await resolve_all_labelhashes_in_name_querying_labelhashes(network_name, name)
 
+        if bulk_mode and simple_name(name):
+            return consolidated_report_from_simple_name(name)
```
djstrong, line 143: When in bulk_mode and the name is simple then instantly return positive report.

api/nameguard/nameguard.py
```diff
@@ -92,6 +94,23 @@ def nested_get(dic, keys):
...
+
+def consolidated_report_from_simple_name(name: str) -> ConsolidatedNameGuardReport:
+    return ConsolidatedNameGuardReport(
+        name=name,
+        namehash=namehash_from_name(name),
+        normalization=Normalization.NORMALIZED,
+        rating=Rating.PASS,
+        risk_count=0,
+        highest_risk=None,
+    )
```
djstrong, line 112: @lightwalker-eth @Carbon225 Please review the regex and fields in the report - is it always true for any name matching the regex?
TODO: Now I see that Punycode check can fail for too long name.

djstrong: NameWrapper label is <=63

PR #180, opened 2023-12-04
remove checkout in GA

djstrong:


Carbon225:
I think so. No need for this option anywhere (?)
The badge should not run on PRs anyway so it should not matter btw.

djstrong (commented):
remoce checkout in add_badge

Carbon225 (approved):

PR #198, opened 2023-12-21
intro page for api reference

notrab:


notrab:
@lightwalker-eth it would be good to merge this too.

PR #234, opened 2024-02-19
Rename ens webfont font family [sc-24299]

notrab:


djstrong:
So this solves the issue when some Inter is remotely loaded by other import, right? So we should prefix all our web fonts?

notrab:
@djstrong yeah this resolves an issue if you use or load Inter in addition to this library.

PR #238, opened 2024-03-24
feat: separate css build

notrab:
* Wait for `@namehash/tailwind-config` to be published so we can import that dependency from the internal UI library we have.

### DIFF
--- a/packages/nameguard-react/src/styles.css
+++ b/packages/nameguard-react/src/styles.css
@@ -1,0 +1,3 @@
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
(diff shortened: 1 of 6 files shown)


packages/nameguard-react/src/styles.css
```diff
@@ -0,0 +1,3 @@
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
```
FrancoAguzzi, line 3: Blank line needed. Seems to be the only update needed in this PR. Great work with the prefix config 👍🏼

FrancoAguzzi (approved):

notrab:
I'm going to close this, but keep in mind we will need to bring this back at some point!

PR #240, opened 2024-03-26
Update string length in ETHRegistrar

djstrong:


notrab:
@djstrong we removed nameparser from this repo today. It now lives at https://github.com/namehash/ens-utils

PR #247, opened 2024-04-04
refactor: replace types usage by enums

FrancoAguzzi:


apps/website/package.json
```diff
@@ -14,7 +14,7 @@
     "@heroicons/react": "2.0.18",
     "@namehash/ens-utils": "1.3.0",
     "@namehash/ens-webfont": "workspace:*",
-    "@namehash/nameguard": "workspace:*",
+    "@namehash/nameguard": "0.1.3",
```
notrab, line 17: @FrancoAguzzi is this intentional? The website should be able to install from the monorepo?

notrab (approved):
Aside from the changes to the version in `package.json` for nameguard/react, this looks good to me. It meets the goals we discussed in Slack for using enums instead of types so we can remove the logic in namehash/kit.

PR #256, opened 2024-04-26
chore: refactor react components to use Rating

FrancoAguzzi:


packages/nameguard-react/src/components/icons/ErrorShield.tsx
```diff
@@ -0,0 +1,25 @@
+import { ShieldSize } from "../Report/Shield";
+import { ErrorShieldLarge } from "./ErrorShieldLarge";
+import { ErrorShieldMedium } from "./ErrorShieldMedium";
+import { ErrorShieldSmall } from "./ErrorShieldSmall";
+import { ErrorShieldMicro } from "./ErrorShieldMicro";
+
+import React from "react";
+
+export const ErrorShield = ({
```
lightwalker-eth, line 9: I don't believe we should call this an "ErrorShield" or an "ErrorShieldIcon" because the icon doesn't really visually express the idea of an error.

Instead, this icon represents uncertainty. We don't know what the actual rating is.

Suggest renaming this UI component and the ideas related to it to "UnknownShieldIcon".

packages/nameguard-react/src/components/NameShield/index.tsx
```diff
@@ -1,50 +1,46 @@
...
+import { LoadingShield } from "../icons/LoadingShield";
+import { ErrorShield } from "../icons/ErrorShield";
 
 type NameShieldProps = {
   data?: ConsolidatedNameGuardReport;
   children?: ReactNode;
   disableHover?: boolean;
   size?: ShieldSize;
-} & React.ComponentProps<typeof Shield>;
+  error?: string;
```
lightwalker-eth, line 19: Why is error a string and what possible values might it have? Appreciate if we could refine and/or document this. It seems we're actually just using this as a boolean??

packages/nameguard-react/src/components/NameBadge/index.tsx
```diff
@@ -1,22 +1,19 @@
...
 import cc from "classcat";
 
 import { NameShield } from "../NameShield";
-import { Shield, ShieldSize } from "../Report/Shield";
 import { ShieldError } from "../Shield/Error";
+import { LoadingShield, ShieldSize } from "../..";
 
-type NameBadgeProps = {
+interface NameBadgeProps {
   name?: string;
```
lightwalker-eth, line 10: Can you help me understand why we have both a `name` and a `data` param here?

If we have a `name` param, then why does the loading state for a NameBadge not show the `name`, as I assume we would always have a `name` value, even if `data` is loading?

I'm really confused.

Alternatively, perhaps we should totally remove the `name` param and just use the `name` field that's returned in `data`.

Either way, I'm confused.

examples/with-nextjs/src/app/components/ImpersonationShield.tsx
```diff
@@ -38,7 +25,12 @@ export function ImpersonationShield({ data }: ImpersonationShieldProps) {
...
 
-  const textClass = cc(["font-semibold", textColor(nameguard_result?.rating)]);
+  if (!nameguard_result) return <LoadingShield />;
+
+  const textClass = cc([
+    "font-semibold",
+    getNameGuardRatingTextColors(nameguard_result.rating),
+  ]);
 
   const Icon = nameguard_result?.rating === Rating.warn ? WarnIcon : AlertIcon;
```
lightwalker-eth, line 35: Why do we have the ? after `nameguard_result` here? Same idea for the rest of the code here in this function.

lightwalker-eth (changes requested):
@FrancoAguzzi Thanks for your updates. Reviewed and shared feedback :+1:

packages/nameguard-react/src/components/NameBadge/index.tsx
notrab: If memory serves me correctly, This badge is used on the marketing site which can be used when the data is available. Name is a placeholder/fallback and is later replaced with `data`.

packages/nameguard-react/src/components/Report/Shield.tsx
```diff
@@ -1,26 +1,35 @@
...
+    case Rating.alert: {
+      return "text-red-600 hover:text-red-700";
+    }
+    case Rating.pass: {
+      return "text-emerald-600 hover:text-emerald-800";
+    }
+    case Rating.warn: {
+      return "text-yellow-500 hover:text-yellow-600";
+    }
+    default: {
```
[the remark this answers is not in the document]
notrab: The `default` applies whenever it;s not one of the other three states. Since this is the default colour, we need  it.

packages/nameguard-react/src/components/NameShield/index.tsx
```diff
@@ -1,50 +1,46 @@
...
+} & React.ComponentProps;
 
 export function NameShield({
```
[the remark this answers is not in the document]
FrancoAguzzi: Great that you've raised this topic. It seems to me that the ReportShield (old Report/Shield.tsx) is not previewed in NameGuard figma specs. This is the difference between ReportShield and NameShield. I suggest we remove ReportShield and update NameShield name to ReportShield. @notrab any ideas around the shared infos above?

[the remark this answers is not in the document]
FrancoAguzzi: Uploading Gravação de Tela 2024-04-29 às 13.48.58.mov…

packages/nameguard-react/src/components/Report/Shield.tsx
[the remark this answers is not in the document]
FrancoAguzzi: Hey! It is really a decision to be made. 

The places we currently use this `text-gray-400` are:

1. Where we display the number of risks encountered in the Ui... e.g. `3 risks detected`
2. Texts around the application... e.g. `You can contact us using any XMTP chat app, including:`
3. Icons around the application such as `XMarkIcon` and `MagnifyingGlassIcon`

For me, the usage of this function to pass along `text-gray-400` depends on wether we want to reuse it not only for `Rating` typed prop but also for `CheckResultCode` typed prop. 

IMO we shouldn't use this function for returning the text color for non-Rating prop. CheckResultCode is not used in NameHash, it is more used in NameGuard and thus, within few NameGuard core features (Only the point 1 above)

@lightwalker-eth can you give me the final decision voice regarding this?

packages/nameguard-react/src/components/NameShield/index.tsx
FrancoAguzzi: It is really just a boolean used for `UnknownShieldIcon` rendering... Updated the prop type 👍🏼

packages/nameguard-react/src/components/NameShield/index.tsx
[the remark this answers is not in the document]
notrab: I don't have the full context of the discussion about this but at the time of creating these, there were components which are specific to our website report, and Nameguard library itself, which is probably where some of the confusion is.

These components were created in the space of a week, 4 months ago. So there's likely a better way to do it now, but keep in mind that the website and report required different visuals for the same rating.

packages/nameguard-react/src/components/Report/Shield.tsx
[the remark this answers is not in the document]
lightwalker-eth: @notrab @FrancoAguzzi The `getNameGuardRatingTextColors` function takes only one param: a `Rating`. There are only 3 possible `Rating` values: https://github.com/namehash/nameguard/blob/main/packages/sdk/src/index.ts#L80-L95

Therefore we should respect the type system. There's no other "default" case here. We should not add "default" cases where they should be impossible to reach.

@notrab Nothing here is for a `CheckResultCode`. If you think another function should be created for colors for these, please suggest one in a PR. Thanks

packages/nameguard-react/src/components/NameShield/index.tsx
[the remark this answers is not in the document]
FrancoAguzzi: Keeping this in mind I went to review Website's implementation and there is no usage to the what was known as ReportShield, currently, thus, I have deleted this file and named the current NameShield as ReportShield, which is a more aligned name for this component since it presents a ShieldIcon + Report details in a tooltip.

packages/nameguard-react/src/components/NameBadge/index.tsx
FrancoAguzzi: Captura de Tela 2024-04-29 às 18 40 41

Hey gents, I am facing the following error in NameGuard website view. Not sure when why this started to appear but I rolled back some versions and identified it started to cause local dev. issue from v1.3.0

I will investigate it deeper tomorrow

I could not fix anything related to NameBadge today because of this issue that prevented me from proper testing the component behavior

lightwalker-eth: > If memory serves me correctly, This badge is used on the marketing site which can be used when the data is available. Name is a placeholder/fallback and is later replaced with data.

@notrab That would make sense to me, but the `name` value isn't being used as a placeholder during loading. It would be nice if that was fixed / changed.

lightwalker-eth: @FrancoAguzzi I think you might have some commits for this PR that you haven't pushed to GitHub yet?

@FrancoAguzzi As for your question, it seems the NameGuard website uses ENS Utils v1.3.0 as defined here: https://github.com/namehash/nameguard/blob/main/apps/website/package.json#L15 So not sure where the v1.4.0 issue comes from?

Breaking changes are frequently being made on new ENS Utils releases as the library is still not so mature and a number of refactorings remain to be needed.

It looks like your error is related to `date-fns` which is a dependency of ENS Utils for some time now. I'm not sure what version it was added as a dependency, but NPM install should theoretically handle everything there for us?

@FrancoAguzzi Happy to assist but will need more info to be able to help.

notrab: @lightwalker-eth sounds good. There are lots of components that compose other components on the marketing site. When we started there was a clear split between `nameguard-react` and `website` folders, but when we integrated NameGuard into Namehash repo, we had to bring some components across to nameguard-react so they could be embedded for our use case only.

It might make sense we add those components to our own `namehash-utils` package, but let's not do that now.

packages/nameguard-react/src/components/ReportShield/index.tsx
```diff
@@ -4,35 +4,35 @@ import cc from "classcat";
...
+import { UnknownShieldIcon } from "../icons/UnknownShieldIcon";
 
 type NameShieldProps = {
   data?: ConsolidatedNameGuardReport;
   children?: ReactNode;
   disableHover?: boolean;
-  size?: ShieldSize;
-  error?: string;
+  size?: ShieldIconSize;
+  error: boolean;
```
lightwalker-eth, line 19:
```suggestion
  hadLoadingError: boolean;
```

packages/nameguard-react/src/components/NameBadge/index.tsx
lightwalker-eth: @notrab Thanks for the background there. For now I'm happy to keep components inside `nameguard-react` that might be needed on the `website` folder (at least where that's logical).

packages/nameguard-react/src/components/Report/ShieldIcon.tsx
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  return components[componentName];
+};
+
+function capitalizeFirstLetter(string: string) {
+  return string.charAt(0).toUpperCase() + string.slice(1);
+}
+
+export const ShieldIcon = ({
```
lightwalker-eth, line 96:
```suggestion
export const RatingIcon = ({
```

We should name things based on the ideas they represent, not on the specific visual representation of those ideas: (ex: a shield).

packages/nameguard-react/src/components/NameBadge/index.tsx
lightwalker-eth: @FrancoAguzzi Please add additional examples to the "rating-shields" page for this UI component. We need to see all variations of this UI component when `name` isn't defined, and all variations when `name` is defined.

lightwalker-eth (changes requested):
@FrancoAguzzi Thanks for your updates. Reviewed and shared feedback :+1:

packages/nameguard-react/src/components/Report/ShieldIcon.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { WarnShieldMediumIcon } from "../icons/WarnShieldMediumIcon";
+import { AlertShieldMediumIcon } from "../icons/AlertShieldMediumIcon";
+import { PassShieldSmallIcon } from "../icons/PassShieldSmallIcon";
+import { WarnShieldSmallIcon } from "../icons/WarnShieldSmallIcon";
+import { AlertShieldSmallIcon } from "../icons/AlertShieldSmallIcon";
+import { AlertIcon as AlertShieldMicroIcon } from "../icons/AlertIcon";
+import { WarnIcon as WarnShieldMicroIcon } from "../icons/WarnIcon";
+import { PassIcon as PassShieldMicroIcon } from "../icons/PassIcon";
+
+export function getNameGuardRatingHoverTextColors(rating: Rating) {
```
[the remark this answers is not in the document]
FrancoAguzzi: Good idea

packages/nameguard-react/src/components/Report/ShieldIcon.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { PassShieldLargeIcon } from "../icons/PassShieldLargeIcon";
+import { WarnShieldLargeIcon } from "../icons/WarnShieldLargeIcon";
+import { AlertShieldLargeIcon } from "../icons/AlertShieldLargeIcon";
+import { PassShieldMediumIcon } from "../icons/PassShieldMediumIcon";
+import { WarnShieldMediumIcon } from "../icons/WarnShieldMediumIcon";
+import { AlertShieldMediumIcon } from "../icons/AlertShieldMediumIcon";
+import { PassShieldSmallIcon } from "../icons/PassShieldSmallIcon";
+import { WarnShieldSmallIcon } from "../icons/WarnShieldSmallIcon";
+import { AlertShieldSmallIcon } from "../icons/AlertShieldSmallIcon";
+import { AlertIcon as AlertShieldMicroIcon } from "../icons/AlertIcon";
```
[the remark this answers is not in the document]
FrancoAguzzi: No reason

packages/nameguard-react/src/index.ts
```diff
@@ -7,16 +7,16 @@ export { useChatModalStore } from "./stores/chat";
 export { useSearchStore } from "./stores/search";
 
 export { NameBadge } from "./components/NameBadge";
```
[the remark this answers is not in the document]
FrancoAguzzi: I think it is a good idea

apps/website/src/app/page.tsx
```diff
@@ -1222,7 +1224,9 @@ function ListSectionBadge(props: ListBadgeProps) {
 function ChecksSection() {
   const checkCircle = (
     <div className="w-12 h-12 flex justify-center items-center flex-shrink-0 p-[14px] bg-white rounded-[40px] border border-gray-200">
-      <CheckCircleIcon className="text-emerald-600 w-5 h-5" />
+      <CheckCircleIcon
+        className={cc(["w-5 h-5", getNameGuardRatingTextColors(Rating.pass)])}
```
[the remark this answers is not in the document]
FrancoAguzzi: Created one `checkResultCodeTextColor` function

packages/nameguard-react/src/components/Shield/UnknownShield.tsx
```diff
@@ -1,24 +1,24 @@
...
 
-type ShieldErrorProps = {
+type UnknownShieldProps = {
   children?: ReactNode;
-  size?: ShieldSize;
+  size?: ShieldIconSize;
 } & React.ComponentProps<typeof ShieldIcon>;
 
-export const ShieldError = ({
+export const UnknownShield = ({
```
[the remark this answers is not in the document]
FrancoAguzzi: This component shows an Unknown Shield + Report in Tooltip. Renamed it

lightwalker-eth (changes requested):
@FrancoAguzzi Many great improvements here. 🙌 Reviewed and shared feedback :+1:

packages/nameguard-react/src/components/Report/Check.tsx
```diff
@@ -77,7 +79,10 @@ function icon(code: CheckResultCode): React.ReactNode {
     case CheckResultCode.info:
       return (
         <svg
-          className="w-5 h-5 fill-current text-gray-400 hover:text-gray-500 transition cursor-pointer"
+          className={cc([
+            "w-5 h-5 fill-current text-gray-400 hover:text-gray-500 transition cursor-pointer",
```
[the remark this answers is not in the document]
FrancoAguzzi: No!

apps/website/src/app/page.tsx
```diff
@@ -1225,7 +1225,7 @@ function ChecksSection() {
   const checkCircle = (
     <div className="w-12 h-12 flex justify-center items-center flex-shrink-0 p-[14px] bg-white rounded-[40px] border border-gray-200">
       <CheckCircleIcon
-        className={cc(["w-5 h-5", getNameGuardRatingTextColors(Rating.pass)])}
+        className={cc(["w-5 h-5", ratingTextColor(Rating.pass)])}
```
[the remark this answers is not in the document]
FrancoAguzzi: Yes!

packages/nameguard-react/src/components/Report/Check.tsx
```diff
@@ -30,8 +27,7 @@ function icon(code: CheckResultCode): React.ReactNode {
         <svg
```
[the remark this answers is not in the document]
FrancoAguzzi: Sounds great

packages/nameguard-react/src/components/Report/Check.tsx
```diff
@@ -93,7 +98,10 @@ function icon(code: CheckResultCode): React.ReactNode {
...
       return (
         <svg
-          className="w-[12px] h-[10px] fill-current text-gray-400 hover:text-gray-500 transition cursor-pointer"
+          className={cc([
+            "w-[12px] h-[10px] fill-current transition cursor-pointer",
+            checkResultCodeTextColor(CheckResultCode.skip, true),
+          ])}
           viewBox="0 0 12 10"
           fill="none"
           xmlns="http://www.w3.org/2000/svg"
```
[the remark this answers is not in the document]
FrancoAguzzi: Hi @lightwalker-eth, thanks for these ideas, I have acted on them and on all the ones above this one as well.

The one thing I did not act was on adding `CheckIcon` to `/rating-shields` since I am yet facing a build issue on website repo, related to NameGuard React bundling. Once I have this fixed I will update `/rating-shields` with all our important components variations and send it to your review and feedback 👍🏼👨🏼‍💻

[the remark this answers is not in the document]
FrancoAguzzi: Hi @lightwalker-eth ! `CheckIcon` was included with its different states in `/rating-shields` page 😉

lightwalker-eth (changes requested):
@FrancoAguzzi Hey, many great updates here. Still more for us to improve. Reviewed and shared feedback.

packages/nameguard-react/src/components/ReportBadge/index.tsx
```diff
@@ -1,22 +1,19 @@
...
+import { LoadingShieldIcon, RatingIconSize } from "../..";
 
-type NameBadgeProps = {
+interface ReportBadgeProps {
   name?: string;
   data?: ConsolidatedNameGuardReport;
   onClick?: MouseEventHandler<HTMLButtonElement>;
-  error?: string;
-};
+  error?: boolean;
```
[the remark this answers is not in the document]
FrancoAguzzi: Updated it to `hadLoadingError`

packages/nameguard-react/src/components/ReportShield/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const { title, subtitle, rating, risk_count, highest_risk } = data;
+
+  const textClass = cc(["font-semibold mb-1", ratingTextColor(rating)]);
+
+  if (disableHover) {
+    return <RatingIcon rating={rating} size={size} {...props} />;
+  }
+
+  return (
+    <Tooltip trigger={<RatingIcon rating={rating} size={size} {...props} />}>
```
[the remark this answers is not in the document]
FrancoAguzzi: There are two components that replicate this code:

`ReportIcon` and `UnknownReportShield` 

After an analysis I realize it would not be valuable to make these 1: logic complexity would increase a lot and code lines that are reused by both of these are less than 5

I decided to go forward with these separated, then

Let me know if anything 😉

apps/website/src/app/rating-shields/page.tsx
```diff
@@ -20,59 +23,301 @@ export default function RatingShieldsPage() {
...
-          </div>
 
-          <div className="flex items-center justify-center">
-            <Shield name="lightwalker.eth" />
-          </div>
-          <div className="flex items-center justify-center">
-            <Shield name="culturecafé.eth" />
+            <div className="flex items-center justify-center">
+              <ReportShield
+                data={{
```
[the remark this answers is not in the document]
FrancoAguzzi: Created one helper function to centralize this data passing 👍🏼

apps/website/src/app/rating-shields/page.tsx
```diff
@@ -20,59 +23,301 @@ export default function RatingShieldsPage() {
...
+                onClick={() => alert("Clicked")}
+              />
+            </div>
           </div>
-          <div className="flex items-center justify-center"></div>
-          <div className="flex items-center justify-center">
-            <ShieldIcon status={CheckResultCode.info} />
+          <div className="grid grid-cols-6 gap-x-6 py-5">
+            <div className="flex items-center font-mono">
+              <pre>{"<ReportShield> \n Hello World \n </ReportShield>"}</pre>
```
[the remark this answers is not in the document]
FrancoAguzzi: Hey! Now both `ReportIcon` and `ReportBadge` have default and custom onClick handlers!

apps/website/src/app/docs/rating-icon/page.tsx
```diff
@@ -0,0 +1,106 @@
...
+        <h1 className="justify-center flex font-bold text-2xl mb-4">
+          {"<"}RatingIcon {"/>"} documentation
+        </h1>
+
+        <div className="grid grid-cols-5 gap-x-6 py-5 text-center font-medium mt-12">
+          <div></div>
+          <div>Rating.pass</div>
+          <div>Rating.warn</div>
+          <div>Rating.alert</div>
+          <div>Loading Rating</div>
```
lightwalker-eth, line 23: What about Unknown? Suggest we create a distinct table for these as they are a distinct UI component (not a `RatingIcon`).

Also, there should be a distinct table for these too because they also are not `RatingIcon` from a technical perspective.

CleanShot 2024-05-07 at 17 01 39

apps/website/src/app/docs/check-icon/page.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+import { CheckResultCode } from "@namehash/nameguard";
+import { CheckIcon } from "@namehash/nameguard-react";
+
+export default function RatingShieldsPage() {
+  return (
+    <div className="py-12 max-w-7xl mx-auto px-6 space-y-12">
+      <div className="divide-y divide-gray-100">
+        <h1 className="justify-center flex font-bold text-2xl mb-4">
+          {"<"}CheckResultCode {"/>"} documentation
```
lightwalker-eth, line 11:
```suggestion
          {"<"}CheckIcon {"/>"} documentation
```

apps/website/src/app/docs/report/page.tsx
```diff
@@ -0,0 +1,232 @@
...
+            <div className="flex items-center justify-center">
+              <ReportIcon name="notrab.eth" hadLoadingError={true} />
+            </div>
+            <div className="flex items-center justify-center">
+              <ReportIcon name="notrab.eth" />
+            </div>
+          </div>
+          <div className="grid grid-cols-6 gap-x-6 py-5">
+            <div className="flex items-center font-mono">
+              <pre>{"<ReportIcon onClick />"}</pre>
```
lightwalker-eth, line 54: Could we make it more clear that this is a custom onClick override? My concern is that someone might think that the variation without a `onClick` passed should not be clickable. But that's not true. All cases are clickable, it's just that this case has an `onClickOverride`.

apps/website/src/app/docs/report/page.tsx
```diff
@@ -0,0 +1,232 @@
...
+          </div>
+          <div className="flex items-center justify-center">
+            <ReportBadge name="loading.eth" />
+          </div>
+        </div>
+
+        <div className="grid grid-cols-6 gap-x-6 py-5">
+          <div className="flex items-center">
+            <div className="flex items-center font-mono">
+              <pre>{"<ReportBadge onClick />"}</pre>
```
lightwalker-eth, line 139: Could we make it more clear that this is a custom onClick override? My concern is that someone might think that the variation without a `onClick` passed should not be clickable. But that's not true. All cases are clickable, it's just that this case has an `onClickOverride`.

apps/website/src/app/docs/report/page.tsx
```diff
@@ -0,0 +1,232 @@
...
+          {"<"}ReportIcon {"/>"} documentation
+        </h1>
+
+        <div className="divide-y divide-gray-100">
+          <div className="grid grid-cols-6 gap-x-6 py-5 text-center font-medium">
+            <div></div>
+            <div>Success</div>
+            <div>Warn</div>
+            <div>Alert</div>
+            <div>Unknown</div>
```
lightwalker-eth, line 25: Why don't the "Unknown" `ReportIcon` have tooltips? The "Unknown" `ReportBadge` have a tooltip and the `ReportIcon` should work exactly the same.

Additionally, these should all have "clicky-hand" cursors as they are all clickable and they all should open when clicked.

Even if we have an "Unknown" result inside this UI component, clicking on it should work correctly no matter what.

apps/website/src/app/docs/report/page.tsx
```diff
@@ -0,0 +1,232 @@
...
+            <div>Loading</div>
```
lightwalker-eth, line 26: We should ensure the following is true for all the "Loading" states for both `ReportIcon` and `ReportBadge`:

* They should always have a clicky-hand cursor. They are all clickable and they should all open when clicked. Even if we are still loading the result inside this UI component clicking on it should work correctly no matter what.

apps/website/src/app/docs/report/page.tsx
```diff
@@ -0,0 +1,232 @@
...
```
lightwalker-eth, line 99: Why isn't the icon in the loading states for `ReportBadge` softly animated? The loading state of the icon for `ReportIcon` is. `ReportBadge` should do the same.

apps/website/src/app/docs/report/page.tsx
lightwalker-eth, line 25:
```suggestion
            <div>hasLoadingError</div>
```

apps/website/src/app/docs/report/page.tsx
```diff
@@ -0,0 +1,232 @@
...
+          {"<"}ReportBadge {"/>"} documentation
+        </h1>
+
+        <div className="divide-y divide-gray-100">
+          <div className="grid grid-cols-6 gap-x-6 py-5 text-center font-medium">
+            <div></div>
+            <div>Success</div>
+            <div>Warn</div>
+            <div>Alert</div>
+            <div>Unknown</div>
```
lightwalker-eth, line 98:
```suggestion
            <div>hasLoadingError</div>
```

apps/website/src/app/docs/report/page.tsx
lightwalker-eth, line 99:
```suggestion
            <div>no data</div>
```

apps/website/src/app/docs/report/page.tsx
lightwalker-eth, line 26:
```suggestion
            <div>no data</div>
```

packages/nameguard-react/src/components/ReportIcon/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  size?: RatingIconSize;
+} & React.ComponentProps;
+
+declare global {
+  interface Window {
+    location: Location;
+  }
+}
+
+export function ReportIcon({
```
lightwalker-eth, line 27: I believe each `ReportIcon` should also take a `name` param with all the same logic and rules for that as we have in `ReportBadge`. Whenever we create a `ReportIcon` we should already know what name we're creating it for.

This is important based on the rules for how `ReportIcon` should always be clickable no matter if they are loading or hadLoadingError.

packages/nameguard-react/src/components/ReportBadge/index.tsx
```diff
@@ -0,0 +1,85 @@
+import { type ConsolidatedNameGuardReport } from "@namehash/nameguard";
+import React from "react";
+import cc from "classcat";
+
+import { ReportIcon } from "../ReportIcon/index";
+import { RatingLoadingIcon, RatingIconSize } from "../..";
+import { UnknownReportShield } from "../UnknownReportShield/UnknownReportShield";
+
+interface ReportBadgeProps {
+  name: string;
```
lightwalker-eth, line 10: This UI component should always validate that if `data` is not null that `data.name === name`. If this isn't true then it should throw an Error.

This idea should also be documented in /** ... **/ comments above the UI component.

packages/nameguard-react/src/components/ReportBadge/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+          size={RatingIconSize.small}
+          className={"cursor-pointer"}
+        />
+      </button>
+    );
+
+  return (
+    <button className={buttonAndCursorClass} onClick={onClickHandler}>
+      <span className="text-black text-sm leading-5 ens-webfont">
+        {data.beautiful_name || data.name}
```
lightwalker-eth, line 66: Ah, there's a problem here.

It's possible that data.name is different than data.beautiful_name. This is the difference between the normalized version of a name and the beautiful version of a name. For example, consider the name: "ξsk3nder.eth". The current logic will have the name being displayed suddenly change once the data is loaded.

The fix to this is to stop passing `name` as a string and instead pass it as an `ENSName` instead as defined in `ens-utils`. Then this UI component will can always show `name.displayName` where we show the name to the user. For cases in these UI components where we generate a link, we should instead be using `name.name`.

You can use the `buildENSName` function in `ens-utils` to convert from name as a string to name as an `ENSName`.

packages/nameguard-react/src/components/Report/RatingIcon.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  return components[componentName];
+};
+
+function capitalizeFirstLetter(string: string) {
+  return string.charAt(0).toUpperCase() + string.slice(1);
+}
+
+export const RatingIcon = ({
```
lightwalker-eth, line 68: Please note how `RatingIcon` is supposed to have a hover state for all cases except for loading.

https://www.figma.com/file/Kt4nUHyZJ5NuPCSZZkXmPL/NameGuard?type=design&node-id=395-3753&mode=design&t=wrzBFLo8iSoUN5bc-4

CleanShot 2024-05-07 at 17 46 24

Could we please add these? Thank you :+1:

apps/website/src/app/docs/report/page.tsx
```diff
@@ -0,0 +1,232 @@
...
+            <div>Loading</div>
+          </div>
+        </div>
+
+        <div className="grid grid-cols-6 gap-x-6 py-5">
+          <div className="flex items-center">
+            <div className="flex items-center font-mono">
+              <pre>{"<ReportBadge />"}</pre>
```
lightwalker-eth, line 106: Could you please also create a variation with a very long name for us to test with? For example, please consider this case from Figma:

CleanShot 2024-05-07 at 17 49 15

https://www.figma.com/file/Kt4nUHyZJ5NuPCSZZkXmPL/NameGuard?type=design&node-id=390-3364&mode=design&t=wrzBFLo8iSoUN5bc-4

lightwalker-eth (changes requested):
@FrancoAguzzi Hey some really super updates here. Nice work! I've completed another detailed review and shared additional suggestions for feedback :+1:

packages/nameguard-react/src/components/Report/CheckResultCard.tsx
```diff
@@ -1,12 +1,14 @@
 import React from "react";
 import { CheckResult } from "@namehash/nameguard";
-import { Check } from "./Check";
+import { CheckIcon } from "./CheckIcon";
 
 export function CheckResultCard(props: CheckResult) {
+  if (!props) return <></>;
```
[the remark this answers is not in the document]
FrancoAguzzi: Sorry, this is a debugging piece of code that I left in the commit without need.

packages/nameguard-react/src/components/Report/RatingIcon.tsx
FrancoAguzzi: There is one misunderstanding above: the artboard mentioned describes `ReportIcon` and not `RatingIcon`

`RatingIcon` represents three different states: Rating.pass, Rating.warn and Rating.alert
`ReportIcon` renders `RatingIcon` and, in case of a failed rating status or a loading status, displays the Unknown state or the Loading state.

apps/website/src/app/docs/rating-icon/page.tsx
FrancoAguzzi: Hi @lightwalker-eth !

I've defined the following tables:

`RatingIcon` table (`/docs/rating-icon`):

Captura de Tela 2024-05-08 às 17 35 47


`ReportIcon` table (`/docs/report`):

Captura de Tela 2024-05-08 às 17 34 47

packages/nameguard-react/src/components/Report/RatingIcon.tsx
FrancoAguzzi: Hover states added 😉

packages/nameguard-react/src/utils/text.ts
```diff
@@ -2,7 +2,7 @@ import { CheckResultCode, Rating } from "@namehash/nameguard";
 
 export function checkResultCodeTextColor(
   check: CheckResultCode,
-  isHover = false
+  isHover = false,
 ): string {
   let color = "";
```
lightwalker-eth, line 7: Better to just make this `let color: string | null = null;` or something like that.

packages/nameguard-react/src/utils/text.ts
```diff
@@ -23,7 +23,8 @@ export function checkResultCodeTextColor(
         : "text-yellow-500";
       return color;
```
lightwalker-eth, line 24: There's no need to repeat `return color` on each case. Just put that once after the whole switch statement.

apps/website/src/app/docs/report/page.tsx
```diff
@@ -109,75 +256,99 @@ export default function RatingShieldsPage() {
...
-const getReportExampleData = (rating: Rating): ConsolidatedNameGuardReport => {
+const getExampleReportName = (
+  rating?: Rating,
+  compWithError = false,
+): ENSName => {
+  if (compWithError) return buildENSName("notrab.eth");
+
+  if (!rating) return buildENSName("loading.eth");
+
+  switch (rating) {
```
lightwalker-eth, line 341: Better to call `buildENSName(getExampleReportData(rating).name)` or something like that.

apps/website/src/app/docs/report/page.tsx
```diff
@@ -109,75 +256,99 @@ export default function RatingShieldsPage() {
...
           </div>
 
           <div className="flex items-center justify-center">
             <ReportBadge
-              name="lightwalker.eth"
-              data={getReportExampleData(Rating.pass)}
-              onClick={() => alert("Clicked")}
+              ensName={getExampleReportName(Rating.pass)}
+              data={getExampleReportData(Rating.pass)}
+              onClickOverride={() => alert("Clicked")}
```
lightwalker-eth, line 297: Please see my other comment about how the signature of the `onClickOverride` function should change to include the `ENSName` that was clicked on.

Please update the logic so that these "alert" calls are dynamic to show the specific name that was clicked on.

lightwalker-eth (changes requested):
@FrancoAguzzi Thanks for updates. Reviewed and shared feedback :+1:

packages/nameguard-react/src/components/ReportIcon/index.tsx
```diff
@@ -25,33 +40,69 @@ declare global {
...
+      window.location.href = `https://nameguard.io/inspect/${encodeURIComponent(
+        ensName.name,
+      )}`;
+    }
+  };
+
   if (hadLoadingError) {
-    return <RatingUnknownIcon size={size} {...props} />;
+    return (
+      <UnknownReportShield
```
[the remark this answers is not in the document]
FrancoAguzzi: Yes

packages/nameguard-react/src/components/ReportShield/index.tsx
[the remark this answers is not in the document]
lightwalker-eth: @FrancoAguzzi Thanks I appreciate that! Let's please rename `UnknownReportShield` though. Would `UnknownReportIcon` be good? Appreciate your advice :+1:

packages/nameguard-react/src/components/ReportBadge/index.tsx
```diff
@@ -62,24 +101,45 @@ export function ReportBadge({
...
+      <div className="invisible absolute left-0 top-0 pointer-events-none">
+        <div ref={nameRef}>{ensName.displayName}</div>
+      </div>
+
+      {/*
+          Below HTML is the rendered badge and tooltip, being the
+          last one displayed only when needed. When is it needed?
+          Whenever the ensName displayed is longer than the badge itself.
+        */}
+      {showTooltip ? (
```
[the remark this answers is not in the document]
FrancoAguzzi: I've briefly updated this piece of code. `Tooltip`'s trigger elm is slightly different from `ReportBadge`'s displayed name.

apps/website/src/app/components/HeroCarousel.tsx
```diff
@@ -37,18 +37,24 @@ const examples = [
...
     <div className="w-[200%] group flex flex-nowrap justify-center items-center space-x-1 animate-carousel group-hover:pause-on-hover">
       {data?.results?.map((report, index) => (
-        <HeroReportBadge key={`carousel-item-${index}`} data={report} />
+        <HeroReportBadge
+          key={`carousel-item-${index}`}
+          ensName={ensNames[index]}
+          data={report}
+        />
       ))}
       {data?.results?.map((report, index) => (
```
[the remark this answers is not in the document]
FrancoAguzzi: I think this was added in the past in order to avoid duplicated names in the visible ones list... Probably this names list grew with time and now this duplication of comps is not anymore needed

apps/website/src/app/docs/report/page.tsx
```diff
@@ -0,0 +1,491 @@
...
+
+  return buildENSName(getExampleReportData(rating, withLongName).name);
+};
+
+const getExampleReportData = (
+  rating: Rating,
+  withLongName = false,
+): ConsolidatedNameGuardReport => {
+  if (withLongName) {
+    switch (rating) {
```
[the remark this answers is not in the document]
FrancoAguzzi: It can return the same thing from the other switch case below, thus, I don't think it is necessary

packages/nameguard-react/src/components/ReportBadge/index.tsx
```diff
@@ -1,93 +1,170 @@
...
 }
 
-export function ReportBadge({ name, onClick, data, error }: ReportBadgeProps) {
+/*
+  Below number is a measure of the maximum width that the ensName 
+  should have inside ReportBadge. If the ensName displayed is longer 
+  than this number, the badge will truncate the ensName and display a 
+  tooltip on hover that shows the full ensName. This number is measured in pixels.
+*/
+const MAX_ENSNAME_DISPLAY_WIDTH = 200;
```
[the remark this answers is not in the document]
FrancoAguzzi: Nope, we can allow this to be a prop passed, great idea @lightwalker-eth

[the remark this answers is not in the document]
FrancoAguzzi: Captura de Tela 2024-05-13 às 13 20 11
Captura de Tela 2024-05-13 às 13 20 01

FrancoAguzzi:
Hi @lightwalker-eth , I have reintroduced the duplicated "data" mapping to badges in `HeroCarousel` because I saw this is needed for a good Ui experience when seeing badges rolling... Otherwise:

Captura de Tela 2024-05-13 às 13 50 46

I also introduced docs to specify this in code 👍🏼

lightwalker-eth (changes requested):
@FrancoAguzzi reviewed and shared feedback

packages/nameguard-react/src/components/icons/RatingUnknownIcon.tsx
```diff
@@ -8,16 +8,21 @@ import React from "react";
...
   size,
-  ...props
-}: { size: RatingIconSize } & React.ComponentProps) => {
+  isNotInteractive = false,
+  className = "",
+}: {
+  size: RatingIconSize;
+  isNotInteractive?: boolean;
+} & React.ComponentProps) => {
+  console.log();
```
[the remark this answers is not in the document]
FrancoAguzzi: 💯

apps/website/src/app/docs/report/page.tsx
```diff
@@ -380,63 +376,19 @@ export default function RatingShieldsPage() {
 const getExampleReportName = (
   rating?: Rating,
   compWithError = false,
```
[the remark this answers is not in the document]
FrancoAguzzi: Function updated

packages/nameguard-react/src/components/Search/SearchEmptyState.tsx
```diff
@@ -122,6 +122,9 @@ export const SearchEmptyState = () => {
             {isLoading &&
               examples.map((e, index) => (
                 <ReportBadge
```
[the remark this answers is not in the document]
FrancoAguzzi: @lightwalker-eth are you suggesting to have this auto-scrolling carousel inside this modal?

**auto-scrolling carousel:**
Captura de Tela 2024-05-14 às 09 45 14
Captura de Tela 2024-05-14 às 09 45 14


**modal:**
Captura de Tela 2024-05-14 às 09 44 58
Captura de Tela 2024-05-14 às 09 44 58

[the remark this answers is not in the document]
lightwalker-eth: @FrancoAguzzi Thanks for asking. No I wasn't suggesting that. I thought a `HeroReportBadge` doesn't have anything specific inside of it for a carousel. So I was asking why couldn't we use it inside the carousel AND in the search empty state?

packages/nameguard-react/src/components/Search/SearchEmptyState.tsx
```diff
@@ -36,11 +36,7 @@ export const SearchEmptyState = () => {
   const { openModal } = useSearchStore();
   const exampleNames = examples.map((n) => n.name);
 
-  const {
-    data,
-    isLoading,
-    error: hadLoadingError,
```
lightwalker-eth: Not sure if I understand this change. Can you please confirm that we will continue handling the case of a loading error correctly?

packages/nameguard-react/src/components/Report/ResultCodeIcon.tsx
```diff
@@ -0,0 +1,97 @@
+import React from "react";
+import { Tooltip } from "../Tooltip/Tooltip";
+import { CheckResultCode, Rating } from "@namehash/nameguard";
+import { CheckResultCodePassIcon } from "../icons/CheckResultCodePassIcon";
+import { CheckResultCodeWarnIcon } from "../icons/CheckResultCodeWarnIcon";
+import { CheckResultCodeAlertIcon } from "../icons/CheckResultCodeAlertIcon";
+import { CheckResultCodeInfoIcon } from "../icons/CheckResultCodeInfoIcon";
+import { CheckResultCodeSkipIcon } from "../icons/CheckResultCodeSkipIcon";
+
+function text(code: CheckResultCode | Rating) {
```
lightwalker-eth, line 10: I'm strongly against using the type system this way. Let's please totally separate `CheckResultCode` from `Rating`.

The test if code is good or not requires a higher standard than just asking the question: Does it work? We also need to pass the test of: Is it maintainable? Is it easy for other developers to understand and use correctly without bugs being created?

For me, this fails those tests.

Also important: a `CheckResultCode` is conceptually different from a `Rating`. We should avoid smashing these ideas together as if they are the same thing. They are not.

packages/nameguard-react/src/components/ReportBadge/index.tsx
```diff
@@ -118,7 +118,12 @@ export function ReportBadge({
         Whenever the ensName displayed is longer than MAX_ENSNAME_DISPLAY_WIDTH.
       */}
       {displayFullNameInTooltip === null ? (
-        <div className="h-2.5 my-[5px] w-[90px] mr-1.5 bg-gray-100 rounded-lg"></div>
+        <div
+          style={{ maxWidth: maxEnsNameDisplayWidth }}
+          className="h-2.5 my-[5px] mr-1.5 bg-gray-100 rounded-lg text-transparent"
```
lightwalker-eth, line 123: Why text-transparent?

packages/nameguard-react/src/components/Report/NoGraphemesWarning.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 import cc from "classcat";
-import { CheckIcon } from "./CheckIcon";
+import { ResultCodeIcon } from "./ResultCodeIcon";
```
lightwalker-eth, line 3: `ResultCodeIcon` is too generic. It sounds like it could be an icon for any kind of result code. But this icon is specifically for the purpose of visualizing the result from a NameGuard check. Ex: a `CheckResultCode`.

I'm ok to rename this component to `CheckResultIcon` or `CheckResultCodeIcon`.

Let's take a step back. What is NameGuard doing? It's important to understand otherwise all these names are just meaningless noise.

NameGuard performs a series of "checks" on names, labels, and graphemes. Each of these "checks" produces a result in the form of a `CheckResultCode`. Each `CheckResultCode` can be visualized in a UI using a `CheckResultIcon`.

NameGuard also provides "ratings" on names, labels, and graphemes. The `Rating` assigned to a name, label, or grapheme is a function of all the results of all the checks performed on that name, label, or grapheme.

A check result is fundamentally different from a rating. Let's please completely avoid smashing these ideas together.

packages/nameguard-react/src/components/Report/GraphemeListItem.tsx
```diff
@@ -48,14 +48,13 @@ export function GraphemeListItem({ item }: GraphemeListItemProps) {
               <RiskCounter count={item.risk_count} />
             </div>
 
-            {item.highest_risk && (
```
lightwalker-eth: Please double check this change. Thanks

lightwalker-eth (changes requested):
@FrancoAguzzi Hey thanks for your updates. Reviewed and shared feedback.

apps/website/src/app/docs/check-icon/page.tsx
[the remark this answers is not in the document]
FrancoAguzzi: Seems like git mv does not resolve this because I was proudly using it imagining that you'd see it working but in PR changes preview it really doesn't do any difference... This is the difference I see in GitHub Desktop App:

Captura de Tela 2024-05-14 às 16 28 20


I used the command in all name changes you've reviewed this last time 🤔

packages/nameguard-react/src/components/Report/CheckIcon.tsx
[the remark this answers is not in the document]
FrancoAguzzi: https://github.com/namehash/nameguard/pull/256#discussion_r1600544784

packages/nameguard-react/src/components/Report/NoGraphemesWarning.tsx
FrancoAguzzi: @lightwalker-eth we then need an idea to create `CheckResultIcon`s for objects that provide a `null` `highest_risk`, which is exactly the case described in one of your loom videos where `Looks good` didn't have an icon representation. Before this PR was created, a plain string was used to inform "pass" or "info", for example, for the `CheckResultIcon` code and this is why it used to work "without problems". When `CheckResultCode` was introduced as `code` type the things broke because `CheckResultCode`  is only provided when `highest_risk` is defined, otherwise, we only have a `Rating` (from `ConsolidatedNameGuardReport.rating` to trust on). I don't see how we can give support to all use cases without:

1. Making `CheckResultIcon` receive either a `CheckResultCode` or a `Rating` OR
2. Creating two dedicated components, one for `CheckResultCode` and one for `Rating`, to render the icons that currently `CheckResultIcon` already renders

packages/nameguard-react/src/components/Report/ResultCodeIcon.tsx
FrancoAguzzi: https://github.com/namehash/nameguard/pull/256#discussion_r1600551012

packages/nameguard-react/src/components/Report/GraphemeListItem.tsx
FrancoAguzzi: https://github.com/namehash/nameguard/pull/256#discussion_r1600551012

packages/nameguard-react/src/components/Search/SearchEmptyState.tsx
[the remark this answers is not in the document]
FrancoAguzzi: Ah, yes!

apps/website/src/app/docs/check-icon/page.tsx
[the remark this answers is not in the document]
lightwalker-eth: @FrancoAguzzi Ok nice, thank you! Really appreciate that. Maybe there's some issue with GitHub in their PR review interface.

packages/nameguard-react/src/index.ts
```diff
@@ -7,8 +7,9 @@ export { SettingsModal } from "./components/SettingsModal/SettingsModal";
 
 export { ReportBadge } from "./components/ReportBadge";
 export { ReportIcon } from "./components/ReportIcon";
-export { ResultCodeIcon } from "./components/Report/ResultCodeIcon";
+export { CheckResultCodeIcon } from "./components/Report/CheckResultCodeIcon";
 export { RatingIcon, RatingIconSize } from "./components/Report/RatingIcon";
+export { NGWebsiteReportBadge } from "./components/Report/NGWebsiteReportBadge";
```
lightwalker-eth, line 12: This UI component should not be in the `nameguard-react` package. That package is for everyone who might want to build a React UI using NameGuard. This UI component is specific to the NameGuard website and therefore it should be moved.

packages/nameguard-react/src/components/Report/CheckResultCodeIcon.tsx
```diff
@@ -77,17 +83,19 @@ function icon(
...
  */
-export function ResultCodeIcon({
+export function CheckResultCodeIcon({
   code,
+  onClick = () => {},
   isInteractive = false,
   className = "",
 }: {
+  onClick?: () => void;
   code: CheckResultCode | Rating;
```
lightwalker-eth, line 93: Let's please ensure we remove any code that doesn't strictly separate `CheckResultCode` from `Rating`. These are never to be treated like they are the same thing.

packages/nameguard-react/src/components/Report/CheckResultCodeIcon.tsx
```diff
@@ -28,6 +28,7 @@ function text(code: CheckResultCode | Rating) {
 function icon(
   code: CheckResultCode | Rating,
```
lightwalker-eth, line 29: Let's please ensure we remove any code that doesn't strictly separate `CheckResultCode` from `Rating`. These are never to be treated like they are the same thing.

apps/website/src/app/components/HeroCarousel.tsx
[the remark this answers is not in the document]
lightwalker-eth: @FrancoAguzzi If this isn't needed anymore then let's please remove it. If it still is needed please add a comment explaining why it is needed.

lightwalker-eth (changes requested):
Still issues to fix.

lightwalker-eth (changes requested):

packages/nameguard-react/src/index.ts
FrancoAguzzi: @lightwalker-eth `NGWebsiteReportBadge` is used both in https://nameguard.io/ and in https://nameguard.io/ modal. The first leaves inside `apps/website` while the second, is a component inside `@namehash/nameguard-react`. We cannot import components from `apps/website` to `@namehash/nameguard-react` and this is why I had to move `NGWebsiteReportBadge` to `@namehash/nameguard-react`. For next step, I'd recommend:

1. Providing this `NGWebsiteReportBadge` for importing from `@namehash/nameguard-react` as part of the package utilities OR;
2. Using `NGWebsiteReportBadge` in apps/website and using similar implementation but not same component in `@namehash/nameguard-react`'s `SearchEmptyState` component;

Let me know your thoughts around this 👍🏼

lightwalker-eth: @FrancoAguzzi Ah you’re right I forgot about that.

Let’s keep it inside ‘nameguard-react’. Also, given the important correction you shared, we should probably rename the component to something more generic. I only suggested a name that was specific for the NG website due to my misunderstanding.

packages/nameguard-react/package.json
```diff
@@ -25,7 +25,7 @@
     "@headlessui-float/react": "0.11.4",
     "@headlessui/react": "1.7.17",
     "@heroicons/react": "2.0.18",
-    "@namehash/ens-utils": "latest",
+    "@namehash/ens-utils": "1.9.0",
```
notrab, line 28: @FrancoAguzzi this can now be `workspace:^` since it's part of the monorepo.

packages/nameguard-react/package.json
```diff
@@ -25,7 +25,7 @@
...
     "@namehash/ens-webfont": "latest",
```
notrab, line 29: Same for this, let's set this version to `workspace:^`

packages/nameguard-react/package.json
```diff
@@ -25,7 +25,7 @@
...
     "@namehash/nameguard": "latest",
```
notrab, line 30: Same for this, let's set this version to `workspace:^`

apps/website/src/app/components/HeroCarousel.tsx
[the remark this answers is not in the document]
lightwalker-eth: @FrancoAguzzi An open question for you here.

packages/nameguard-react/src/components/Report/CheckResultCard.tsx
```diff
@@ -1,12 +1,14 @@
 import React from "react";
-import { CheckResult } from "@namehash/nameguard";
+import { CheckResult, CheckResultCode } from "@namehash/nameguard";
 import { CheckResultCodeIcon } from "./CheckResultCodeIcon";
 
 export function CheckResultCard(props: CheckResult) {
+  const status = props ? props.status : CheckResultCode.pass;
```
lightwalker-eth, line 6: This suggests our type system is being violated. It should be impossible for `props` not not be defined when this function is called. The param is not identified as being optional and I don't believe it should be optional. It should be the responsibility of the caller of this function to handle the possibility that they don't have a `CheckResult` and to manually pass a value such as `CheckResultCode.pass` as needed.

lightwalker-eth (changes requested):
@FrancoAguzzi We're getting really close! Reviewed and shared feedback :+1:

apps/website/src/app/docs/check-result-code-icon/page.tsx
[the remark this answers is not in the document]
lightwalker-eth: @notrab Yes the plan is to transition to that in the future. For now the goal was to confirm we're getting these UI components right. Let's not grow the PR with any storybook use. Let's come back to this soon though :+1:

lightwalker-eth (commented):
Latest commit looks good :+1:

apps/website/src/app/docs/check-result-code-icon/page.tsx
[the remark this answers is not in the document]
lightwalker-eth: @notrab Created a shortcut story to track the goal of following up on your suggestion. I'll mark this thread as Resolved as working to merge this PR asap.

https://app.shortcut.com/ps-web3/story/25015/transition-nameguard-ui-component-examples-to-storybook

lightwalker-eth (approved):
Approved for merge :+1:

apps/website/src/app/docs/check-result-code-icon/page.tsx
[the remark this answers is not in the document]
notrab: Amazing, thank you @lightwalker-eth

PR #258, opened 2024-05-01
bump ens-utils version

notrab:


lightwalker-eth:
No longer needed in new monorepo strategy

PR #259, opened 2024-05-10
Run Python tests only on changes in its directory

djstrong:


djstrong:
@Carbon225 @notrab What do you think?

PR #263, opened 2024-05-16
Refine big merge

lightwalker-eth:


apps/examples.nameguard.io/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "nextjs-sdk",
+  "name": "nameguard-examples",
```
notrab, line 2: This is an example, we can get rid of this and version.

lightwalker-eth: This is useful during the build process.

CleanShot 2024-05-16 at 16 12 57

lightwalker-eth: Removed version :+1:

apps/examples.nameguard.io/postcss.config.cjs
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   plugins: {
     tailwindcss: {},
-    autoprefixer: {},
-  },
-}
+    autoprefixer: {}
+  }
+}
```
FrancoAguzzi, line 6: Blank line needed

FrancoAguzzi (approved):

apps/examples.nameguard.io/postcss.config.cjs
lightwalker-eth: Interesting. What creates this need? I added one as requested but don't see what it changes. If some different fix is needed please feel welcome to action that directly in this PR :+1:

PR #266, opened 2024-05-17
Centralize color logic

lightwalker-eth:


FrancoAguzzi (approved):

packages/nameguard-react/src/components/Report/Banner.tsx
```diff
@@ -7,7 +7,7 @@ import { RatingIcon, RatingIconSize } from "./RatingIcon";
 import { RatedBox } from "../RatedBox/RatedBox";
 import { ReportChangesApplied } from "./ReportChangesApplied";
 import { ReportFormattedDisplayName } from "./ReportFormattedName";
-import { ratingTextColor } from "../../utils/text";
+import { ratingTextColor } from "../../utils/colors";
```
notrab, line 10: We should create an alias for `@` if it doesn't already exist so we can remove the relative imports once and for all.

PR #276, opened 2024-05-22
chore: explicitly describe RatingLoadingIcon size when rendering it

FrancoAguzzi:


lightwalker-eth (commented):
@FrancoAguzzi Changes here look good. Can you please prepare a changeset as part of this PR? This can be a nice practice opportunity for us.

FrancoAguzzi:
@lightwalker-eth I'd love to but I can't! 😃⬇️

Changeset is not tied to the `apps/` folder, but it is to `packages/` folder

In this PR I only changed one file from within `apps/examples.nameguard.io` which means no Changeset files can be generated for later version bump!

Image below reflects what we see in terminal after a `pnpm changeset` call:
Captura de Tela 2024-05-23 às 11 28 36

Indeed I want to practice Changeset usage! 

No doubts I will soon be able to do it in some of our PRs 🚀🙌🏼

lightwalker-eth (approved):
@FrancoAguzzi You're right! Thanks for your good comment. Reviewed and approved :+1:

PR #281, opened 2024-05-30
feat: [SC-24992] Create DisplayedName and export it from NG React pkg…

FrancoAguzzi:
… | Create doc. page


apps/nameguard.io/src/app/docs/displayed-name/page.tsx
```diff
@@ -0,0 +1,152 @@
+"use client";
+
+import { ENSName, buildENSName } from "@namehash/ens-utils";
+import { DisplayedName } from "@namehash/nameguard-react";
+
+export default function DisplayedNamePage() {
```
lightwalker-eth, line 6: Ideally we move all these examples to the storybook. But until we do that please ensure the `/docs` page includes a link to each of these types of pages you've created. Thanks.

packages/nameguard-react/src/components/DisplayedName/DisplayedName.tsx
```diff
@@ -0,0 +1,98 @@
...
+        Below HTML is the rendered text and tooltip, being the
+        tooltip only shown when it is needed, on mouse hover. When is it needed?
+        Whenever the ensName displayed is longer than MAX_ENSNAME_DISPLAY_WIDTH.
+      */}
+      {displayFullNameInTooltip && displayTooltipWhenNameClamps ? (
+        <Tooltip
+          trigger={
+            <div
+              style={{ maxWidth: maxEnsNameDisplayWidth }}
+              className="cursor-pointer text-black text-sm leading-5 ens-webfont truncate pr-1"
```
lightwalker-eth, line 78: I don't think there should be "cursor-pointer" here. Just because there's a tooltip doesn't mean it's a link. Only interactive links should have "cursor-pointer".

While you update this, please ensure everything in the UX continues to work correctly on `ReportBadge`

lightwalker-eth (changes requested):
@FrancoAguzzi Thanks for the updates here. Reviewed and shared feedback :+1:

packages/nameguard-react/src/components/DisplayedName/DisplayedName.tsx
```diff
@@ -0,0 +1,98 @@
...
+            >
+              {displayName}
+            </div>
+          }
+        >
+          <span className="text-white text-sm leading-5 ens-webfont">
```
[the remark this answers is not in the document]
FrancoAguzzi: Maybe a max-width with line breaking? 

I don't think we should add ellipsis to the tooltip text

Line breaking would limit the width of the tooltip, making it grow in height whenever the text is longer than what fits in the informed width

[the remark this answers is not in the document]
FrancoAguzzi: @lightwalker-eth I have added a `maxTooltipWidth` prop that is by default 400 and that can be changed... Here is how the default prop looks like:

Captura de Tela 2024-06-04 às 19 17 56

FrancoAguzzi:
`<Text />` component was created and `/docs` and `/docs/displayed-name` pages were updated

packages/nameguard-react/src/components/DisplayedName/DisplayedName.tsx
```diff
@@ -1,98 +1,37 @@
...
-  maxEnsNameDisplayWidth?: number;
+  name: ENSName;
+  stylingClasses?: string;
+  maxDisplayWidth?: number;
+  maxTooltipWidth?: number;
   displayUnnormalizedNames?: boolean;
   displayTooltipWhenNameClamps?: boolean;
 }
 
-const DEFAULT_MAX_ENSNAME_DISPLAY_WIDTH = 200;
```
lightwalker-eth: I feel like we should keep a custom default for this inside `DisplayedName` that is distinct from any default inside `Text`.

packages/nameguard-react/src/components/Text/Text.tsx
```diff
@@ -0,0 +1,107 @@
...
+  stylingClasses?: string;
+  string: string;
+}
+
+const DEFAULT_MAX_DISPLAY_WIDTH = 200;
+const DEFAULT_MAX_TOOLTIP_WIDTH = 400;
+
+export const Text = ({
+  string,
+  stylingClasses = "",
```
lightwalker-eth, line 18: Should we rename this to `textStylingClasses` as these are applied to the text?

packages/nameguard-react/src/components/ReportBadge/index.tsx
```diff
@@ -35,7 +35,7 @@ interface ReportBadgeProps {
     than this number, the badge will truncate the ensName and display a 
     tooltip on hover that shows the full ensName. This number is measured in pixels.
   */
-  maxEnsNameDisplayWidth?: number;
+  maxDisplayWidth?: number;
 }
 
 const DEFAULT_MAX_ENSNAME_DISPLAY_WIDTH = 200;
```
lightwalker-eth, line 41: This constant should be removed. This idea belongs inside `DisplayedName` not `ReportBadge`. Please let me know if you think I'm missing something :+1:

lightwalker-eth (changes requested):
@FrancoAguzzi Thanks for your updates. Reviewed and shared feedback :+1:

packages/nameguard-react/src/components/Text/Text.tsx
```diff
@@ -0,0 +1,107 @@
...
+            {getTextElm(
+              "text-white text-sm leading-5 break-all",
+              maxTooltipWidth,
+            )}
+          </Tooltip>
+        </>
+      ) : (
+        <>
+          {getTextElm(
+            `${stylingClasses} text-black text-sm leading-5 truncate pr-1.5`,
```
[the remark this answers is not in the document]
FrancoAguzzi: Patternized this 👍🏼

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
```diff
@@ -0,0 +1,274 @@
...
+              <pre>
+                {
+                  "<DisplayedName name\n textStylingClasses='text-purple-400' />"
+                }
+              </pre>
+            </div>
+
+            <div className="flex items-center justify-center">
+              <DisplayedName
+                textStylingClasses="text-purple-400"
```
lightwalker-eth, line 144: Can you please include an example in the docs showing what happens if we set custom text styling classes like these AND there's an overflow tooltip?

I want to confirm that the styling of the text in the tooltip will NOT be changed if we change the styling of the name itself. In other words, maybe I want the ENS name to appear in a larger font or in a different color, etc.. but then I don't want anything to change about the styling in the tooltip.

What's a good way for us to achieve this?

lightwalker-eth (changes requested):
@FrancoAguzzi Thanks for your updates! Reviewed and shared feedback

packages/nameguard-react/src/components/Text/Text.tsx
```diff
@@ -3,90 +3,88 @@ import React, { useEffect, useState, useRef } from "react";
...
+  const invisibleTextWidthTester = useRef<null | HTMLParagraphElement>(null);
 
   /* 
-    Below state is only true if the name displayed is longer 
-    than name max-width (defined by MAX_ENSNAME_DISPLAY_WIDTH)
+    Below state is only true if the text displayed 
+    is longer than text maxDisplayWidth
   */
-  const [displayFullNameInTooltip, setDisplayFullNameInTooltip] =
+  const [displayFullTextInTooltip, setDisplayFullTextInTooltip] =
```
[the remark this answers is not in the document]
FrancoAguzzi: Perfectly noted 👌🏼

packages/nameguard-react/src/components/DisplayedName/DisplayedName.tsx
```diff
@@ -0,0 +1,39 @@
...
+}: DisplayedNameProps) {
+  const showUnnormalizedName =
+    displayUnnormalizedNames &&
+    name.normalization === Normalization.unnormalized;
+  const displayName = showUnnormalizedName ? name.name : name.displayName;
+
+  return (
+    <Text
+      text={displayName}
+      textStylingClasses={textStylingClasses + " ens-webfont"}
```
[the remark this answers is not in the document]
FrancoAguzzi: Great catch 🦅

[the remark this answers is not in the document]
FrancoAguzzi: We are not applying styles from `TruncatedText` into `Tooltip`, yet

I can create one new prop to do so: it'd work just as `textStylingClasses` but being applied in tooltip content

This goes in opposite compared to [this comment](https://github.com/namehash/nameguard/pull/281#discussion_r1626175779) but I see `ens-webfont` is indeed needed in tooltip content. 
Do you think we should leave the comment behind and agree a `tooltipTextStylingClasses` is good idea?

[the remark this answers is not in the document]
lightwalker-eth: @FrancoAguzzi We need to achieve a few key goals here:

1. If the tooltip needs to use `ens-webfont`, this needs to be possible.
2. If we set a special style on the truncated text, that style should not automatically be applied to the tooltip.

Therefore it seems we need two distinct ways to configure styles.

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
lightwalker-eth: @FrancoAguzzi Not sure about adding the complexity of that enum. I think it's ok to allow UI components to be given class names that don't make sense. Maybe needs two distinct custom classname fields?

lightwalker-eth (commented):
@FrancoAguzzi Thanks for questions and updates. Reviewed and shared feedback :+1:

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
```diff
@@ -41,14 +41,14 @@ export default function DisplayedNamePage() {
             <div className="w-full flex justify-center items-center font-mono">
               <pre>
                 {
-                  "<TruncatedText name \n textStylingClasses='text-purple-400' />"
+                  "<TruncatedText name \n textStylingClasses='\nens-webfont font-black' />"
```
lightwalker-eth, line 44: why the '\n'?

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
```diff
@@ -59,12 +59,16 @@ export default function DisplayedNamePage() {
         <div className="divide-y divide-gray-100">
           <div className="grid grid-cols-2 gap-x-6 py-5">
             <div className="w-full flex justify-center items-center font-mono">
-              <pre>{"<TruncatedText name \n maxTooltipWidth={100} />"}</pre>
+              <pre>
+                {
+                  "<TruncatedText name \n textStylingClasses='\nens-webfont font-black' />"
```
lightwalker-eth, line 64: There's multiple \n here?

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
```diff
@@ -75,19 +79,14 @@ export default function DisplayedNamePage() {
         <div className="divide-y divide-gray-100">
           <div className="grid grid-cols-2 gap-x-6 py-5">
             <div className="w-full flex justify-center items-center font-mono">
-              <pre>
-                {
-                  "<TruncatedText name \n displayTooltipWhenNameOverflows={false} />"
-                }
-              </pre>
+              <pre>{"<TruncatedText name \n maxTooltipWidth={100} />"}</pre>
```
lightwalker-eth, line 82: Why the \n here?

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
```diff
@@ -110,6 +109,28 @@ export default function DisplayedNamePage() {
             </div>
           </div>
         </div>
+        <div className="divide-y divide-gray-100">
+          <div className="grid grid-cols-2 gap-x-6 py-5">
+            <div className="w-full flex justify-center items-center font-mono">
+              <pre>
+                {
+                  "<TruncatedText name \n displayTooltipWhenNameOverflows={false}\n maxDisplayWidth={200} />"
```
lightwalker-eth, line 117: Why multiple \n here?

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
```diff
@@ -138,19 +159,37 @@ export default function DisplayedNamePage() {
             <div className="w-full flex justify-center items-center font-mono">
               <pre>
                 {
-                  "<DisplayedName name\n textStylingClasses='text-purple-400' />"
+                  "<DisplayedName name\n textStylingClasses='\nens-webfont font-black' />"
```
lightwalker-eth, line 162: Why \n?

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
```diff
@@ -138,19 +159,37 @@ export default function DisplayedNamePage() {
...
               />
             </div>
           </div>
         </div>
+        <div className="divide-y divide-gray-100">
+          <div className="grid grid-cols-2 gap-x-6 py-5">
+            <div className="w-full flex justify-center items-center font-mono">
+              <pre>
+                {
+                  "<DisplayedName name\n tooltipContentStylingClasses='\nens-webfont font-black' />"
```
lightwalker-eth, line 180: Why multiple \n?

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
```diff
@@ -138,19 +159,37 @@ export default function DisplayedNamePage() {
...
-                  "<DisplayedName name\n textStylingClasses='text-purple-400' />"
+                  "<DisplayedName name\n textStylingClasses='\nens-webfont font-black' />"
                 }
               </pre>
             </div>
 
             <div className="flex items-center justify-center">
               <DisplayedName
-                textStylingClasses="text-purple-400"
+                textStylingClasses="ens-webfont font-black"
```
lightwalker-eth, line 169: Why do so many of these have "font-black" in their text styling? This comment applies across the whole page not just this line of code.

packages/nameguard-react/src/components/DisplayedName/DisplayedName.tsx
```diff
@@ -30,9 +32,10 @@ export function DisplayedName({
   return (
     <TruncatedText
       text={displayName}
-      textStylingClasses={textStylingClasses + " ens-webfont"}
       maxTooltipWidth={maxTooltipWidth}
       maxDisplayWidth={maxDisplayWidth}
+      textStylingClasses={textStylingClasses + " ens-webfont"}
+      tooltipContentStylingClasses={tooltipContentStylingClasses}
```
lightwalker-eth, line 38: Shouldn't this also add "ens-webfont"?

packages/nameguard-react/src/components/TruncatedText/TruncatedText.tsx
```diff
@@ -92,7 +94,7 @@ export const TruncatedText = ({
             }
           >
             {getTextElm(
-              "text-white text-sm leading-5 break-all",
+              `text-white text-sm leading-5 break-all ${tooltipContentStylingClasses}`,
```
lightwalker-eth, line 97: Should we use cc here?

If so, please update our code everywhere we call `getTextElem`

packages/nameguard-react/src/components/TruncatedText/TruncatedText.tsx
```diff
@@ -3,11 +3,12 @@ import React, { useEffect, useState, useRef } from "react";
 import { Tooltip } from "../..";
 
 interface TruncatedTextProps {
-  displayTooltipWhenTextOverflows?: boolean;
+  text: string;
   maxTooltipWidth?: number;
   maxDisplayWidth?: number;
   textStylingClasses?: string;
-  text: string;
+  tooltipContentStylingClasses?: string;
```
lightwalker-eth, line 10:
```suggestion
  tooltipTextStylingClasses?: string;
```

lightwalker-eth (changes requested):
@FrancoAguzzi Thanks for updates. Reviewed and shared feedback :+1:

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
FrancoAguzzi: For the string to break into a new line

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
FrancoAguzzi: Yes, is for optimizing text displaying in the screen

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
lightwalker-eth: @FrancoAguzzi What I really want to see on the demo page is that:

1. DisplayedName shows the tooltip in ens-webfont
2. Some of the DisplayedName shows custom font styling for the TruncatedText (but not the tooltip, outside of request 1 above)

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
FrancoAguzzi: Same

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
FrancoAguzzi: Same as above

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
FrancoAguzzi: Same as above

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
FrancoAguzzi: It is an example of how users can customize tooltip content and / or text content styles

packages/nameguard-react/src/components/DisplayedName/DisplayedName.tsx
FrancoAguzzi: Only if the user wants to, right?

lightwalker-eth: @FrancoAguzzi   It should ALWAYS happen inside `DisplayedName`!

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
lightwalker-eth: @FrancoAguzzi I don't understand why we're doing this here? I've never seen this before in my life it looks very strange.

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
lightwalker-eth: On who's screen? The developers screen? The user's screen?

This is really strange for me.

FrancoAguzzi: @lightwalker-eth in both, see how this affects the texts displaying:


https://github.com/namehash/nameguard/assets/49823133/cebe900b-44a0-4dce-a292-c81b69f979a7

The container width is 600px, for screens that are smaller than this the displaying might yet look broken.

For screens wider than this, the displaying is the one we see in the video

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
FrancoAguzzi: @lightwalker-eth `TruncatedText` examples were updated to use plain strings

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
FrancoAguzzi: Here I explain to you the usage of `\n` which is called:

Captura de Tela 2024-06-12 às 09 04 36

lightwalker-eth: @FrancoAguzzi I fully understand what '\n' is 😉 What I don't understand is why we're putting explicit newline characters where we are in the code. These seem like a very strange places for explicit newline characters.

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
lightwalker-eth: @FrancoAguzzi Ah, so these newline characters aren't for the actual HTML code, but rather the strings that are in the "Component usage" column?

If so that makes sense to me!

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
```diff
@@ -101,10 +107,7 @@ export default function DisplayedNamePage() {
...
               <TruncatedText
                 maxDisplayWidth={60}
-                text={
-                  getExampleDisplayedName(DisplayedNameExample.SHORT_MAX_WIDTH)
-                    .name
-                }
+                text={getExampleTruncatedText(TruncatedTextExample.LONG_TEXT)}
               />
             </div>
           </div>
```
lightwalker-eth, line 113: A few lines below this in the `<pre>` tag we should change `displayTooltipWhenNameOverflows` to `displayTooltipWhenTextOverflows`

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
FrancoAguzzi: I get it! The thing is that these are not code lines but these are code documentation lines, so these are plain strings that need to be visually displayed in the screen, thus, they are not code but strings.

apps/nameguard.io/src/app/docs/displayed-name/page.tsx
FrancoAguzzi: Exactly! 👌🏼😃

packages/nameguard-react/src/components/Report/ReportFormattedName.tsx
```diff
@@ -1,26 +1,28 @@
 import React from "react";
+import { DisplayedName } from "../DisplayedName/DisplayedName";
+import { ENSName } from "@namehash/ens-utils";
+import { Normalization } from "@namehash/nameguard";
 
 type Props = {
+  name: ENSName;
```
lightwalker-eth, line 7: The only field we should pass here is an `ENSName`. All the other fields in these Props should be removed. The other props already exist inside the provided `ENSName`.

packages/nameguard-react/src/components/Report/ReportFormattedName.tsx
```diff
@@ -1,26 +1,28 @@
...
+  beautifiedName,
   normalization,
   name,
-  displayName,
 }: Props) {
-  const displayNameDifferent =
-    normalization === "normalized" && name !== displayName;
+  const displayNameDifferentThanBeautifiedName =
+    normalization === Normalization.normalized &&
+    (name.displayName !== beautifiedName || !beautifiedName);
```
lightwalker-eth, line 19:
```suggestion
    (name.name !== beautifiedName || !beautifiedName);
```

The goal of the check here is if the displayName is different than the normalizedName

packages/nameguard-react/src/components/Report/ReportFormattedName.tsx
```diff
@@ -1,26 +1,28 @@
...
 type Props = {
+  name: ENSName;
   normalization: string;
-  name: string;
-  displayName: string;
+  beautifiedName?: string;
 };
 
 export function ReportFormattedDisplayName({
+  beautifiedName,
```
lightwalker-eth, line 13: Please see my comment above about how the props should change.

Please also note that `ENSName` does not have a `beautifiedName` field on purpose. It has a `displayName` field instead. That's because only normalized names can have a `beautifedName` but we always provide a `displayName` for all possible name values.

Please ask if any questions.

packages/nameguard-react/src/components/Report/LoadingSkeleton.tsx
```diff
@@ -26,11 +26,6 @@ export const LoadingSkeleton = ({ parsedName }: LoadingSkeletonProps) => {
               <h1 className="pt-1 text-2xl md:text-4xl text-black font-semibold md:font-bold overflow-hidden overflow-ellipsis whitespace-nowrap ens-webfont">
                 {name}
               </h1>
-              <ReportFormattedDisplayName
```
lightwalker-eth, line 29: Why have we removed this? We want the user to instantly see results as they type (during the LoadingSkeleton) without having to wait for a server response to display back to the user the name that is being inspected.

packages/nameguard-react/src/components/Report/ReportFormattedName.tsx
lightwalker-eth, line 1: Should we align the name of this file with the name of the UI component inside this file?

packages/nameguard-react/src/components/Report/Banner.tsx
```diff
@@ -34,9 +34,9 @@ export function Banner({ report, parsedName }: Props) {
             {name}
```
lightwalker-eth, line 34: We should also be using a `DisplayedName` here, but we need to enhance the `DisplayedName` UI component for this to work correctly.

We need to buildENSName(name) to get an ENSName and then pass that into the `DisplayedName`.

Currently `DisplayedName` always displays the `displayName` field. But here we have a special case where we don't want to show the `displayName` field, we want to show the literal `name` field (which may or may not be normalized or unknown, etc..)

So we need to add another prop to `DisplayedName` that if `true` displays the `name` field rather than the `displayName` field no matter what the `normalization` is.

We need to do the exact same update inside `LoadingSkeleton`.

lightwalker-eth (changes requested):
@FrancoAguzzi Thanks for updates. Reviewed and shared feedback :+1:

lightwalker-eth (changes requested):
@FrancoAguzzi Hey updates look good! Just 2 open questions.

lightwalker-eth (approved):
@FrancoAguzzi Thanks for updates :+1: Reviewed and approved! 🚀

PR #282, opened 2024-06-01
feat: [SC-24987] create LoadingReportIcon

FrancoAguzzi:


packages/nameguard-react/src/components/LoadingReportIcon/LoadingReportIcon.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { ENSName } from "@namehash/ens-utils";
+import { RatingIcon, RatingIconSize } from "../Report/RatingIcon";
+import { RatingLoadingIcon } from "../icons/RatingLoadingIcon";
+
+type LoadingReportIconProps = {
+  onClickOverride?: (ensName: ENSName) => void;
+  size?: RatingIconSize;
+} & React.ComponentProps<typeof RatingIcon>;
+
+export const LoadingReportIcon = ({
```
lightwalker-eth, line 13: Should we rename this to something such as `ReportLoadingIcon` to have more consistency with how we name `RatingLoadingIcon`?

packages/nameguard-react/src/components/LoadingReportIcon/LoadingReportIcon.tsx
```diff
@@ -0,0 +1,57 @@
...
+          <div className="flex items-center justify-between">
+            <span className="font-semibold mb-1 text-white">
+              Loading report...
+            </span>
+          </div>
+
+          <div className="text-sm text-white">
+            <button
+              className="appearance-none underline font-medium"
+              onClick={onClickHandler}
```
lightwalker-eth, line 48: Hmm.. I don't think this calls the callback correctly? How is the param getting passed to the callback?

packages/nameguard-react/src/components/LoadingReportIcon/LoadingReportIcon.tsx
```diff
@@ -0,0 +1,57 @@
+import React from "react";
+
+import { Tooltip } from "../Tooltip/Tooltip";
+import { ENSName } from "@namehash/ens-utils";
+import { RatingIcon, RatingIconSize } from "../Report/RatingIcon";
+import { RatingLoadingIcon } from "../icons/RatingLoadingIcon";
+
+type LoadingReportIconProps = {
```
lightwalker-eth, line 8: Please see my comment below about naming. Whatever we decide there, let's align the name of this type to match.

packages/nameguard-react/src/components/LoadingReportIcon/LoadingReportIcon.tsx
```diff
@@ -0,0 +1,57 @@
...
+  onClickOverride?: (ensName: ENSName) => void;
+  size?: RatingIconSize;
+} & React.ComponentProps<typeof RatingIcon>;
+
+export const LoadingReportIcon = ({
+  onClickHandler,
+  size = RatingIconSize.small,
+
+  /*
+    Props are applied to the shield icon which is the onHover trigger element
```
lightwalker-eth, line 18: Let's please update this comment to not use words such as "shield icon". Let's please use the exact names of UI components to be more clear and specific.

packages/nameguard-react/src/components/ReportIcon/index.tsx
```diff
@@ -8,11 +8,11 @@ import {
 import cc from "classcat";
 
 import { Tooltip } from "../Tooltip/Tooltip";
-import { RatingLoadingIcon } from "../icons/RatingLoadingIcon";
 import { RatingIcon, RatingIconSize } from "../Report/RatingIcon";
 import { checkResultCodeTextColor, ratingTextColor } from "../../utils/text";
 import { ENSName } from "@namehash/ens-utils";
 import { UnknownReportIcon } from "../UnknownReportIcon/UnknownReportIcon";
```
lightwalker-eth, line 14: Please see my other comments about naming standards. It seems maybe we should also rename `UnknownReportIcon` to `ReportUnknownIcon`?

packages/nameguard-react/src/components/ReportIcon/index.tsx
```diff
@@ -8,11 +8,11 @@ import {
...
+import { LoadingReportIcon } from "../LoadingReportIcon/LoadingReportIcon";
 
 type ReportShieldProps = {
   onClickOverride?: (ensName: ENSName) => void;
```
lightwalker-eth, line 18: GitHub won't let me leave this comment where I want to. This comment is actually for line 60 below where we generate the URL for a NameGuard report for an `ENSName`.

Please create a utility function for this inside the NameGuard SDK package. It should take in any `ENSName` and return a URL to inspect that name.

Then please look for opportunities to refactor existing code to call into that new utility function. Thanks.

lightwalker-eth (changes requested):
@FrancoAguzzi Hey great to see the progress here. Reviewed and shared feedback :+1:

packages/nameguard-sdk/src/utils.ts
```diff
@@ -25,3 +27,7 @@ const keccak256Regex = /^(?:0x)?[0-9a-f]{64}$/i;
 export function isKeccak256Hash(hash: string) {
   return keccak256Regex.test(hash);
 }
+
+export function getNameGuardURLForENSname(ensName: ENSName): string {
+  return `https://nameguard.io/inspect/${encodeURIComponent(ensName.name)}`;
```
notrab, line 32: I'm not a fan of this utility in general living inside the SDK, but if we are to have it, we should make sure the SDK generates this using the `endpoint` we pass the constructor:

```ts
class NameGuard {
  private endpoint: URL;
  private network: Network;
  private abortController: AbortController;

  constructor({ endpoint = DEFAULT_ENDPOINT, network = DEFAULT_NETWORK } = {}) {
    this.endpoint = new URL(endpoint);
    this.network = network;
    this.abortController = new AbortController();
  }
```

I would move this function to a public function on the `NameGuard` class that takes in a name to generate the URL.

It's good practice we do this so if we ever want to run Nameguard locally or on another URL, we have fewer constants to change.

cc @lightwalker-eth @FrancoAguzzi

FrancoAguzzi: @notrab thanks for the feedback! Very helpful towards clean coding 

@lightwalker-eth I will wait for your thoughts around how this implementation should look like before doing code updates

lightwalker-eth: I wrote a big reply but then GitHub lost it right before I submitted it due to some bug 😞 Will write a summarized version as round 2.

@notrab I agree with your feedback that this logic probably doesn't belong in the SDK. The SDK is for API-level concerns. The goal that we're discussing now is different than the `endpoint` param passed to a `NameGuard` instance. That `endpoint` param is for API-level requests, but the idea we're talking about is for URLs that end users would use to view a NameGuard report.

@notrab @FrancoAguzzi  Therefore perhaps we should move this logic into the react SDK instead somewhere? Appreciate advice on where we might move that.

lightwalker-eth (changes requested):
@FrancoAguzzi @notrab Thanks for updates. Reviewed and shared feedback :+1:

packages/nameguard-sdk/src/utils.ts
FrancoAguzzi: NameGuard React package seems a good home to this utility

FrancoAguzzi: This is **100%** related to web and it will be used by front-end logics only

lightwalker-eth: @FrancoAguzzi @notrab Ok cool! Let's proceed with moving this logic then :+1:

FrancoAguzzi: Indeed it seems interesting to provide this to devs integrating NameGuard reports... Otherwise it could live inside `apps/nameguard.io`... But devs might be interested in reusing it... At least this is what I see now, could be changed in the future if no _integrators_ adoption, to the `nameguard.io` "namehash-labs-only" folder

packages/nameguard-sdk/src/utils.test.ts
```diff
@@ -50,3 +52,18 @@ describe("isEthereumAddress", () => {
     expect(result).toBe(false);
   });
 });
+
+describe("getNameGuardURLForENSname", () => {
+  it("should return correct NameGuard inspect name page URL for notrab.eth", () => {
+    const result = getNameGuardURLForENSname(buildENSName("notrab.eth"));
+    const expectedResult = "https://nameguard.io/inspect/notrab.eth";
+
+    expect(result).toBe(expectedResult);
```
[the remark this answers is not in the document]
FrancoAguzzi: @lightwalker-eth I did a very quick research that mentioned that for string comparisons, the context we are seeing, toBe can be used while toEqual is excellent for object and other primitives comparisons, can we stick to toBe here?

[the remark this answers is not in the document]
FrancoAguzzi: https://stackoverflow.com/questions/45195025/what-is-the-difference-between-tobe-and-toequal-in-jest

packages/nameguard-react/src/utils/text.ts
```diff
@@ -63,3 +64,7 @@ export function ratingTextColor(rating: Rating, isInteractive = false): string {
 
   return color;
 }
+
+export function getNameGuardURLForENSname(ensName: ENSName): string {
```
lightwalker-eth, line 68:
```suggestion
export function viewNameReportURL(ensName: ENSName): URL {
```

1. Naming change suggested as this is the URL to view the NameGuard report for a name., not the "name guard url" or the "name guard url for ens name" as that can be confused as the API endpoint. Here we're specifically getting the URL to "viewNameReport". View being the key word here.
2. Seems nice to have this return a URL rather than a string.

I also don't think this utility function should be in `/utils/text.ts`. It should move to some other file under `/utils`. Open to ideas maybe `/utils/url.ts`

packages/nameguard-react/src/components/ReportIcon/index.tsx
```diff
@@ -56,19 +63,42 @@ export function ReportIcon({
...
+        if (onTooltipClickOverride) {
+          onTooltipClickOverride(ensName);
+        } else {
+          redirectUserToNameGuardInspectPage(ensName);
+        }
+        break;
     }
   };
 
+  const redirectUserToNameGuardInspectPage = (ensName: ENSName) => {
```
lightwalker-eth, line 85: Seems we should move this to `/utils/url.ts` ? Please see my other comment with more details about that proposed file.

packages/nameguard-react/src/components/ReportBadge/index.tsx
```diff
@@ -52,13 +62,36 @@ export function ReportBadge({
...
+        if (onTooltipClickOverride) {
+          onTooltipClickOverride(ensName);
+        } else {
+          redirectUserToNameGuardInspectPage(ensName);
+        }
+        break;
     }
   };
 
+  const redirectUserToNameGuardInspectPage = (ensName: ENSName) => {
```
lightwalker-eth, line 91: Seems we should move this to `/utils/url.ts` ? Please see my other comment with more details about that proposed file.

lightwalker-eth (changes requested):
@FrancoAguzzi Nice updates! Reviewed and shared feedback :+1:

packages/nameguard-react/src/utils/url.ts
```diff
@@ -0,0 +1,11 @@
+import { ENSName } from "@namehash/ens-utils";
+
+export function viewNameReportURL(ensName: ENSName): URL {
+  return new URL(
+    `https://nameguard.io/inspect/${encodeURIComponent(ensName.name)}`,
+  );
+}
+
+export function redirectUserToNameGuardInspectPage(ensName: ENSName) {
```
lightwalker-eth, line 9:
```suggestion
export function redirectToViewNameReportURL(ensName: ENSName) {
```

lightwalker-eth (changes requested):
@FrancoAguzzi updates look great. Just 1 small change request. Please also add a changeset for the changes made in this PR.

lightwalker-eth (approved):
@FrancoAguzzi looks nice! Thank you :+1: Reviewed and approved 🚀

lightwalker-eth (changes requested):
@FrancoAguzzi I see something broke in the tests here. Could you please investigate it? https://github.com/namehash/nameguard/actions/runs/9537548682/job/26285783558?pr=282

FrancoAguzzi:
@lightwalker-eth there was one test failing which is now fixed but looks like the error is holding tight to ens-utils workspace version... hu, will review this once again tomorrow, with more power and surely solve the solving of it

packages/ens-utils/tsconfig.json
```diff
@@ -2,8 +2,8 @@
   "compilerOptions": {
     "target": "ES2020",
     "lib": ["ES2020", "dom"],
-    "module": "ES2020",
-    "moduleResolution": "node",
+    "module": "ES6",
+    "moduleResolution": "node10",
```
lightwalker-eth, line 6: Hmm.. Is there a special reason for this? https://www.typescriptlang.org/tsconfig/#moduleResolution

CleanShot 2024-07-09 at 22 23 50

lightwalker-eth (changes requested):
@FrancoAguzzi Hey, appreciate the updates here.

Shared 1 comment with a question for you on this PR.

Also, what do you think about merging the latest changes on main into this branch or performing a rebase with main?

apps/examples.nameguard.io/package.json
```diff
@@ -10,6 +10,7 @@
   "dependencies": {
     "@headlessui-float/react": "0.11.4",
     "@headlessui/react": "1.7.17",
+    "@namehash/ens-utils": "1.12.1",
```
lightwalker-eth, line 13: @FrancoAguzzi Shouldn't we be importing this from workspace:* ? This package also lives in the monorepo

FrancoAguzzi:
@notrab @djstrong can you review this test please?

https://github.com/namehash/namekit/actions/runs/10584114051/job/29327608472?pr=282

apps/examples.nameguard.io/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module "@namehash/nameguard-react";
```
lightwalker-eth, line 1: @FrancoAguzzi Please help me understand why we added this file and this line of code.

Is this actually needed? If so, why? If it is needed, then are we doing something special to cause it to be needed in the examples.nameguard.io app? If we aren't doing anything special but still need this then shouldn't we be adding some documentation in the readme files for the related packages instructing everyone to do this when making use of the package?

apps/examples.nameguard.io/src/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module "@namehash/nameguard-react";
```
lightwalker-eth, line 1: @FrancoAguzzi Please help me understand why we added this file and this line of code.

Is this actually needed? If so, why? If it is needed, then are we doing something special to cause it to be needed in the examples.nameguard.io app? If we aren't doing anything special but still need this then shouldn't we be adding some documentation in the readme files for the related packages instructing everyone to do this when making use of the package?

packages/namekit-react/package.json
```diff
@@ -41,8 +41,10 @@
   "dependencies": {
     "@headlessui-float/react": "0.11.4",
     "@headlessui/react": "1.7.17",
+    "@namehash/ens-webfont": "workspace:*",
     "@namehash/ens-utils": "workspace:*",
-    "@namehash/ens-webfont": "workspace:*"
+    "react": "^18.2.0",
```
lightwalker-eth, line 46: @FrancoAguzzi Are you 100% sure you added `react` and `react-dom` as dependencies correctly? Shouldn't they have been added as `devDependencies` rather than `dependencies`?

I already see them listed as `devDependencies`.

What's happening here? PLEASE REVIEW YOUR OWN WORK!

packages/nameguard-react/package.json
lightwalker-eth, line 1: @FrancoAguzzi Whenever you make any changes to any build configuration files, please proactively add a comment about that change inside the PR in GitHub that explains:

1. Why that change was needed.
2. The current status of getting approval from @notrab of the change.

packages/ens-utils/.gitignore
```diff
@@ -0,0 +1,3 @@
+
+# dependencies
+/node_modules
```
lightwalker-eth, line 3: Why are we making this change? This suggests we aren't running our `pnpm` commands from the root of the monorepo. I don't see why we should do this?

packages/nameguard-react/src/components/Share/Share.tsx
```diff
@@ -16,14 +16,15 @@ import { CheckResultCode } from "@namehash/nameguard";
 import { checkResultCodeTextColor } from "../../utils/text";
 import { DisplayedName } from "../DisplayedName/DisplayedName";
 import { buildENSName } from "@namehash/ens-utils";
+import { viewNameReportURL } from "../../utils/url";
 
 type ShareProps = {
-  name?: string;
+  name: string;
```
lightwalker-eth, line 22: We should make this an `ENSName` rather than a `string`.

packages/nameguard-react/src/components/Share/Share.tsx
```diff
@@ -16,14 +16,15 @@ import { CheckResultCode } from "@namehash/nameguard";
...
 };
 
 function createTwitterLink(name: string) {
```
lightwalker-eth, line 25: All of these functions should take an `ENSName` rather than a `string`.

packages/nameguard-react/src/components/Share/Share.tsx
```diff
@@ -137,12 +136,10 @@ export function Share({ name }: ShareProps) {
 
                   <div className="min-h-[200px] flex items-center justify-center px-6 md:px-10">
                     {name && (
```
lightwalker-eth, line 138: Why are we checking here if `name` is defined? It should always be defined based on the defined props of this UI component.

lightwalker-eth (changes requested):
@FrancoAguzzi Thanks for updates. Reviewed and shared feedback

PR #284, opened 2024-06-05
feat(packages): initial namekit-react

notrab:
This stubs `@namehash/namekit-react`


packages/namekit-react/README.md
```diff
@@ -0,0 +1,55 @@
+# @namehash/nameguard-react
```
lightwalker-eth, line 1:
```suggestion
# @namehash/namekit-react
```

packages/namekit-react/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Usage
+
+These components can be used to interface with NameGuard with React.
```
lightwalker-eth, line 5:
```suggestion
These components can be used to interface NameKit with React.
```

packages/namekit-react/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Search
```
lightwalker-eth, line 7: Suggest removing the remainder of this readme until we can build it out more

packages/namekit-react/src/components/DomainCard.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from "react";
+
+export type DomainCardProps = {
+  name: string;
+};
+
+export function DomainCard({ name }: DomainCardProps) {
+  return (
+    <div>
+      Domain card for <span className="ens-webfont">{name}</span>
```
lightwalker-eth, line 10: I believe it can create too much complexity to try to separate `nameguard-react` from `namekit-react`. My assumption is that we want to merge both of these together into a single package: `namekit-react`.

If we don't do this, it gets really complicated from a dependency perspective. For example:

1. `namekit-react` needs `DisplayedName` that Franco is working on which is currently in `nameguard-react`.
2. `namekit-react` need will need the ability to show `DomainCard`, and inside `DomainCard` we need to put UI components from `nameguard-react.

My vote is to just be pragmatic and merge these packages together under the name "namekit-react". If you see it differently please let me know :+1:

lightwalker-eth (changes requested):
@notrab Thanks for the updates here. Reviewed and shared feedback :+1:

lightwalker-eth (approved):
@notrab Thanks for this update. I see there's a build error for storybook. Could you investigate and resolve? Otherwise approved :+1:

PR #286, opened 2024-06-10
Uninspected Names

djstrong:


api/nameguard/nameguard.py
```diff
@@ -142,19 +143,25 @@ def analyse_label(self, label: str):
...
+    ) -> Optional[Union[NameGuardReport, ConsolidatedNameGuardReport]]:
         """
         Inspect a name. A name is a sequence of labels separated by dots.
         A label can be a labelhash or a string.
         If a labelhash is encountered and `resolve_labelhashes` is `True`, a lookup will be performed.
         """
+
+        logger.debug(f"[inspect_name] name: '{name}'")
+
+        # if len(name) > INSPECTABLE_NAMES_LENGTH: #TODO do we need it here or limit number of labelhashes in the name?
```
djstrong, line 155: @lightwalker-eth name length limit is checked after resolving labelhashes. Should we do it before or limit number of labelhashes in a name?

api/nameguard/nameguard.py
```diff
@@ -163,6 +170,9 @@ def inspect_name_sync(
 
         logger.debug(f"[inspect_name] name: '{name}'")
 
+        if len(name) > INSPECTABLE_NAMES_LENGTH:
```
djstrong, line 173: @Carbon225 What about collection generation? Can the result be null?

api/nameguard/nameguard.py
```diff
@@ -397,7 +409,10 @@ async def secure_primary_name(self, address: str, network_name: str) -> SecurePr
         else:
             nameguard_result = await self.inspect_name(network_name, domain)
 
-            if nameguard_result.normalization == Normalization.UNNORMALIZED:
+            if nameguard_result is None:
```
djstrong, line 412: @lightwalker-eth Here we need a status for `secure_primary_name`. Do we need to add new one? Now we have:
```    NORMALIZED = 'normalized'
    NO_PRIMARY_NAME = 'no_primary_name'
    UNNORMALIZED = 'unnormalized'```

File: api/nameguard/nameguard.py (hunk 1 above)
Comment (lightwalker-eth) [reply api/nameguard/nameguard.py:155]: @djstrong Good question. Perhaps we should define a different constant for the max number of labelhashes in a name that we might work to lookup? For example: Even 5 would be more than enough for 99.999999%+ of real cases.

If we're given a name with more labelhashes then that we could skip any lookups of those labelhashes.

What do you think?

File: api/nameguard/nameguard.py (hunk 3 above)
Comment (lightwalker-eth) [reply api/nameguard/nameguard.py:412]: @djstrong Yes I suppose we need a new status for this case. Could you please suggest one? Thank you

File: packages/nameguard-sdk/src/index.ts [hunk 4]
```diff
@@ -600,6 +601,8 @@ class NameGuard {
    *   1. The labelhash of the "childmost" label of a name.
    *   2. The complete parent name of the "childmost" label.
    *
+   * Returns `null` if the `name` is longer than {INSPECTABLE_NAMES_LENGTH} characters.
```
lightwalker-eth, line 604:
```suggestion
   * Returns `null` if labels in the resolved `name` are longer than `MAX_INSPECTABLE_NAME_LENGTH` characters.
```

packages/nameguard-sdk/src/index.ts
```diff
@@ -553,7 +554,7 @@ class NameGuard {
    * Inspects the name associated with a namehash.
    *
    * NameGuard will attempt to resolve the name associated with the namehash through the ENS Subgraph.
-   * If this resolution succeeds then NameGuard will generate and return a `NameGuardReport` for the name.
+   * If this resolution succeeds then NameGuard will generate and return `null` if the `name` is longer than {INSPECTABLE_NAMES_LENGTH} characters or a `NameGuardReport` for the name.
```
lightwalker-eth, line 557:
```suggestion
   * If this resolution succeeds then NameGuard will return a `NameGuardReport` for the name. If the resolved `name` is longer than `MAX_INSPECTABLE_NAME_LENGTH` characters then NameGuard will return `null`.
```

packages/nameguard-sdk/src/index.ts
```diff
@@ -489,7 +490,7 @@ class NameGuard {
   }
 
   /**
-   * Inspects a single name with NameGuard. Provides a `NameGuardReport` including:
+   * Inspects a single name with NameGuard. Provides `null` if the `name` is longer than {INSPECTABLE_NAMES_LENGTH} characters or a `NameGuardReport` including:
```
lightwalker-eth, line 493:
```suggestion
   * Inspects a single name with NameGuard. If the `name` is longer than `MAX_INSPECTABLE_NAME_LENGTH` characters returns `null`; else returns a `NameGuardReport` including:
```

packages/nameguard-sdk/src/index.ts
```diff
@@ -446,6 +446,7 @@ const DEFAULT_ENDPOINT = "https://api.nameguard.io";
 const DEFAULT_NETWORK: Network = "mainnet";
 const DEFAULT_INSPECT_LABELHASH_PARENT = ETH_TLD;
 const MAX_BULK_INSPECTION_NAMES = 250;
+const INSPECTABLE_NAMES_LENGTH = 200;
```
lightwalker-eth, line 449:
```suggestion
const MAX_INSPECTABLE_NAME_LENGTH = 200;
```

packages/nameguard-sdk/src/index.ts
```diff
@@ -394,7 +394,7 @@ export interface NameGuardReport extends ConsolidatedNameGuardReport {
 }
 
 export interface BulkConsolidatedNameGuardReport {
-  results: ConsolidatedNameGuardReport[];
+  results: (ConsolidatedNameGuardReport | null)[];
```
lightwalker-eth, line 397: This typedef seems right

lightwalker-eth (changes requested):
@djstrong Thanks for updates. Reviewed and shared feedback

api/nameguard/nameguard.py
djstrong: `UNINSPECTABLE`

api/nameguard/nameguard.py
djstrong: `MAX_NUMBER_OF_LABELHASHES_IN_NAME = 5`

packages/nameguard-sdk/src/index.ts
lightwalker-eth, line 1: @djstrong Ah shoot I think I gave some bad advice previously to have uninspectable names return null. This creates some tricky issues for UI components because they then have to add a bunch of special cases and hardcoded values for how to represent an uninspectable result.

We need to think more about how to define the data model of these responses. Appreciate your advice. You're also welcome to setup a call where we could discuss :+1:

lightwalker-eth (changes requested):
@djstrong Thanks for the updates here! Reviewed and shared feedback :+1:

packages/nameguard-sdk/src/index.ts
djstrong: As the result should be the same for the inspect name endpoint, the other way is to return some object with status `uninspected`.

packages/nameguard-sdk/src/index.ts
```diff
@@ -394,7 +395,7 @@ export interface NameGuardReport extends ConsolidatedNameGuardReport {
...
```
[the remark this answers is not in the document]
djstrong: Yes.

The only optimization in the bulk endpoint is to gather labelhashes from all names and resolve them in one step. Then iteratively run `inspect_name` for all input names (without duplicates detection).

packages/nameguard-sdk/src/index.ts
```diff
@@ -100,7 +100,8 @@ export enum Rating {
 export type SecurePrimaryNameStatus =
   | "normalized" /** The ENS primary name was found and it is normalized. */
   | "no_primary_name" /** The ENS primary name was not found. */
-  | "unnormalized" /** The ENS primary name was found, but it is not normalized. */;
+  | "unnormalized" /** The ENS primary name was found, but it is not normalized. */
+  | "uninspectable" /** The name was not checked, because it is too long. */;
```
lightwalker-eth:
```suggestion
  | "uninspected" /** The name was not inspected, because it is too long. */;
```

api/nameguard/models/result.py
```diff
@@ -309,13 +309,13 @@ class SecurePrimaryNameStatus(str, Enum):
     * `normalized`: The ENS primary name was found and it is normalized.
     * `no_primary_name`: The ENS primary name was not found.
     * `unnormalized`: The ENS primary name was found, but it is not normalized.
-    * `uninspectable`: The name was not checked, because it is too long.
+    * `uninspected`: The name was not checked, because it is too long.
```
lightwalker-eth, line 312:
```suggestion
    * `uninspected`: The name was not inspected, because it is too long.
```

api/nameguard/nameguard.py
```diff
@@ -148,7 +148,7 @@ async def inspect_name(
         Inspect a name. A name is a sequence of labels separated by dots.
         A label can be a labelhash or a string.
         If a labelhash is encountered and `resolve_labelhashes` is `True`, a lookup will be performed.
-        Returns None if there is more labelhashes than MAX_NUMBER_OF_LABELHASHES_IN_NAME.
+        Returns None if there is more labelhashes than MAX_INSPECTABLE_NAME_UNKNOWN_LABELS.
```
lightwalker-eth, line 151:
```suggestion
        Returns None if there are more labelhashes than MAX_INSPECTABLE_NAME_UNKNOWN_LABELS.
```

api/nameguard/utils.py
```diff
@@ -7,8 +7,8 @@
 from nameguard.exceptions import InvalidNameHash, InvalidTokenID, InvalidEthereumAddress
 from nameguard.models import Rating, GenericCheckResult, Check
 
-MAX_INSPECTABLE_NAME_LENGTH = 200
-MAX_NUMBER_OF_LABELHASHES_IN_NAME = 5
+MAX_INSPECTABLE_NAME_CHARACTERS = 200  # number of characters to inspect in a name including labels and separators
```
lightwalker-eth, line 10:
```suggestion
MAX_INSPECTED_NAME_CHARACTERS = 200  # maximum number of characters to inspect in a name including labels and separators
```

Suggesting that we move away from using the term "inspectable" and instead use "inspected". This is because "inspectable" seems to suggest it is completely impossible in the abstract, while "inspected" in this context suggests that it may theoretically be possible to inspect something longer, but we just decide not to do it.

api/nameguard/utils.py
```diff
@@ -7,8 +7,8 @@
...
+MAX_INSPECTABLE_NAME_UNKNOWN_LABELS = 5  # duplicated unknown labels are counted
```
lightwalker-eth, line 11:
```suggestion
MAX_INSPECTED_NAME_UNKNOWN_LABELS = 5  # each unknown label in a name is distinctly counted, even if there are multiple instances of the same unknown label in a name.
```

Suggesting that we move away from using the term "inspectable" and instead use "inspected". This is because "inspectable" seems to suggest it is completely impossible in the abstract, while "inspected" in this context suggests that it may theoretically be possible to inspect something longer, but we just decide not to do it.

packages/nameguard-sdk/src/index.ts
```diff
@@ -603,7 +603,7 @@ class NameGuard {
    *   1. The labelhash of the "childmost" label of a name.
    *   2. The complete parent name of the "childmost" label.
    *
-   * Returns `null` if labels in the resolved `name` are longer than `MAX_INSPECTABLE_NAME_LENGTH` characters.
+   * Returns `null` if labels in the resolved `name` are longer than `MAX_INSPECTABLE_NAME_CHARACTERS` characters.
```
lightwalker-eth, line 606:
```suggestion
   * Returns `null` if labels in the resolved `name` are longer than `MAX_INSPECTABLE_NAME_CHARACTERS` characters.
```
```suggestion
   * Returns `null` if labels in the resolved `name` are longer than `MAX_INSPECTED_NAME_CHARACTERS` characters.
```

packages/nameguard-sdk/src/index.ts
```diff
@@ -556,7 +556,7 @@ class NameGuard {
    * Inspects the name associated with a namehash.
    *
    * NameGuard will attempt to resolve the name associated with the namehash through the ENS Subgraph.
-   * If this resolution succeeds then NameGuard will return a `NameGuardReport` for the name. If the resolved `name` is longer than `MAX_INSPECTABLE_NAME_LENGTH` characters then NameGuard will return `null`.
+   * If this resolution succeeds then NameGuard will return a `NameGuardReport` for the name. If the resolved `name` is longer than `MAX_INSPECTABLE_NAME_CHARACTERS` characters then NameGuard will return `null`.
```
lightwalker-eth, line 559:
```suggestion
   * If this resolution succeeds then NameGuard will return a `NameGuardReport` for the name. If the resolved `name` is longer than `MAX_INSPECTED_NAME_CHARACTERS` characters then NameGuard will return `null`.
```

packages/nameguard-sdk/src/index.ts
```diff
@@ -517,7 +517,7 @@ class NameGuard {
 
   // TODO: Document how this API will attempt automated labelhash resolution through the ENS Subgraph.
   /**
-   * Inspects up to 250 names at a time with NameGuard. Provides `null` if the `name` is longer than `MAX_INSPECTABLE_NAME_LENGTH` characters or consists of more than `MAX_NUMBER_OF_LABELHASHES_IN_NAME` labelhases; else returns a `ConsolidatedNameGuardReport` for each name provided in `names`, including:
+   * Inspects up to 250 names at a time with NameGuard. Provides `null` if the `name` is longer than `MAX_INSPECTABLE_NAME_CHARACTERS` characters or consists of more than `MAX_INSPECTABLE_NAME_UNKNOWN_LABELS` labelhases; else returns a `ConsolidatedNameGuardReport` for each name provided in `names`, including:
```
lightwalker-eth, line 520:
```suggestion
   * Inspects up to 250 names at a time with NameGuard. Provides `null` if the `name` is longer than `MAX_INSPECTED_NAME_CHARACTERS` characters or consists of more than `MAX_INSPECTED_NAME_UNKNOWN_LABELS` labelhases; else returns a `ConsolidatedNameGuardReport` for each name provided in `names`, including:
```

packages/nameguard-sdk/src/index.ts
```diff
@@ -492,7 +492,7 @@ class NameGuard {
   }
 
   /**
-   * Inspects a single name with NameGuard. If the `name` is longer than `MAX_INSPECTABLE_NAME_LENGTH` characters or consists of more than {MAX_NUMBER_OF_LABELHASHES_IN_NAME} labelhases returns `null`; else returns a `NameGuardReport` including:
+   * Inspects a single name with NameGuard. If the `name` is longer than `MAX_INSPECTABLE_NAME_CHARACTERS` characters or consists of more than {MAX_INSPECTABLE_NAME_UNKNOWN_LABELS} labelhases returns `null`; else returns a `NameGuardReport` including:
```
lightwalker-eth, line 495:
```suggestion
   * Inspects a single name with NameGuard. If the `name` is longer than `MAX_INSPECTED_NAME_CHARACTERS` characters or consists of more than {MAX_INSPECTED_NAME_UNKNOWN_LABELS} labelhases returns `null`; else returns a `NameGuardReport` including:
```

packages/nameguard-sdk/src/index.ts
```diff
@@ -101,7 +101,7 @@ export type SecurePrimaryNameStatus =
   | "normalized" /** The ENS primary name was found and it is normalized. */
   | "no_primary_name" /** The ENS primary name was not found. */
   | "unnormalized" /** The ENS primary name was found, but it is not normalized. */
-  | "uninspectable" /** The name was not checked, because it is too long. */;
+  | "uninspected" /** `isInspectableName(name) === false` therefore the name was not inspected for performance reasons */;
```
lightwalker-eth, line 104:
```suggestion
  | "uninspected" /** `willInspectName(name) === false` therefore the name was not inspected for performance reasons */;
```

packages/nameguard-sdk/src/index.ts
```diff
@@ -447,8 +447,8 @@ const DEFAULT_ENDPOINT = "https://api.nameguard.io";
 const DEFAULT_NETWORK: Network = "mainnet";
 const DEFAULT_INSPECT_LABELHASH_PARENT = ETH_TLD;
 const MAX_BULK_INSPECTION_NAMES = 250;
-const MAX_INSPECTABLE_NAME_LENGTH = 200;
-const MAX_NUMBER_OF_LABELHASHES_IN_NAME = 5;
+const MAX_INSPECTABLE_NAME_CHARACTERS = 200;
```
lightwalker-eth, line 450:
```suggestion
const MAX_INSPECTED_NAME_CHARACTERS = 200;
```

packages/nameguard-sdk/src/index.ts
```diff
@@ -447,8 +447,8 @@ const DEFAULT_ENDPOINT = "https://api.nameguard.io";
...
+const MAX_INSPECTABLE_NAME_UNKNOWN_LABELS = 5;
```
lightwalker-eth, line 451:
```suggestion
const MAX_INSPECTED_NAME_UNKNOWN_LABELS = 5;
```

api/nameguard/web_api.py
```diff
@@ -446,7 +446,7 @@ async def inspect_labelhash_post(request: InspectLabelhashRequest) -> Optional[N
     1. The labelhash of the "childmost" label of a name.
     2. The complete parent name of the "childmost" label.
 
-    Returns `null` if labels in the resolved `name` are longer than {MAX_INSPECTABLE_NAME_LENGTH} characters.
+    Returns `null` if labels in the resolved `name` are longer than {MAX_INSPECTABLE_NAME_CHARACTERS} characters.
```
lightwalker-eth, line 449:
```suggestion
    Returns `null` if labels in the resolved `name` are longer than {MAX_INSPECTED_NAME_CHARACTERS} characters.
```

api/nameguard/web_api.py
```diff
@@ -380,7 +380,7 @@ async def inspect_labelhash_get(
...
```
lightwalker-eth, line 383:
```suggestion
    Returns `null` if labels in the resolved `name` are longer than {MAX_INSPECTED_NAME_CHARACTERS} characters.
```

api/nameguard/web_api.py
```diff
@@ -322,7 +322,7 @@ async def inspect_namehash_post(request: InspectNamehashRequest) -> Optional[Nam
     NameGuard will attempt to resolve the name associated with the `request.namehash` through the ENS Subgraph,
     using the network specified in `request.network_name`.
 
-    If this resolution succeeds then NameGuard will return a `NameGuardReport` for the name. If the resolved `name` is longer than {MAX_INSPECTABLE_NAME_LENGTH} characters then NameGuard will return `null`.
+    If this resolution succeeds then NameGuard will return a `NameGuardReport` for the name. If the resolved `name` is longer than {MAX_INSPECTABLE_NAME_CHARACTERS} characters then NameGuard will return `null`.
```
lightwalker-eth, line 325:
```suggestion
    If this resolution succeeds then NameGuard will return a `NameGuardReport` for the name. If the resolved `name` is longer than {MAX_INSPECTED_NAME_CHARACTERS} characters then NameGuard will return `null`.
```

api/nameguard/web_api.py
```diff
@@ -289,7 +289,7 @@ async def inspect_namehash_get(
     NameGuard will attempt to resolve the name associated with the `namehash` through the ENS Subgraph,
     using the network specified in `network_name`.
 
-    If this resolution succeeds then NameGuard will return a `NameGuardReport` for the name. If the resolved `name` is longer than {MAX_INSPECTABLE_NAME_LENGTH} characters then NameGuard will return `null`.
+    If this resolution succeeds then NameGuard will return a `NameGuardReport` for the name. If the resolved `name` is longer than {MAX_INSPECTABLE_NAME_CHARACTERS} characters then NameGuard will return `null`.
```
lightwalker-eth, line 292:
```suggestion
    If this resolution succeeds then NameGuard will return a `NameGuardReport` for the name. If the resolved `name` is longer than {MAX_INSPECTED_NAME_CHARACTERS} characters then NameGuard will return `null`.
```

api/nameguard/web_api.py
```diff
@@ -226,7 +226,7 @@ async def bulk_inspect_names(request: BulkInspectNamesRequest) -> BulkNameGuardB
     f"""
     ## Inspects up to 250 names at a time with NameGuard.
 
-    Provides `null` if the `name` is longer than {MAX_INSPECTABLE_NAME_LENGTH} characters or consists of more than {MAX_NUMBER_OF_LABELHASHES_IN_NAME} labelhases; else returns a `ConsolidatedNameGuardReport` for each name provided in `request.names`, including:
+    Provides `null` if the `name` is longer than {MAX_INSPECTABLE_NAME_CHARACTERS} characters or consists of more than {MAX_INSPECTABLE_NAME_UNKNOWN_LABELS} labelhases; else returns a `ConsolidatedNameGuardReport` for each name provided in `request.names`, including:
```
lightwalker-eth, line 229:
```suggestion
    Provides `null` if the `name` is longer than {MAX_INSPECTED_NAME_CHARACTERS} characters or consists of more than {MAX_INSPECTED_NAME_UNKNOWN_LABELS} labelhases; else returns a `ConsolidatedNameGuardReport` for each name provided in `request.names`, including:
```

api/nameguard/web_api.py
```diff
@@ -181,7 +181,7 @@ async def inspect_name_post(request: InspectNameRequest) -> Optional[NameGuardRe
     f"""
     ## Inspects a single name with NameGuard.
 
-    If the `name` is longer than {MAX_INSPECTABLE_NAME_LENGTH} characters or consists of more than {MAX_NUMBER_OF_LABELHASHES_IN_NAME} labelhases returns `null`; else returns a `NameGuardReport` including:
+    If the `name` is longer than {MAX_INSPECTABLE_NAME_CHARACTERS} characters or consists of more than {MAX_INSPECTABLE_NAME_UNKNOWN_LABELS} labelhases returns `null`; else returns a `NameGuardReport` including:
```
lightwalker-eth, line 184:
```suggestion
    If the `name` is longer than {MAX_INSPECTED_NAME_CHARACTERS} characters or consists of more than {MAX_INSPECTED_NAME_UNKNOWN_LABELS} labelhases returns `null`; else returns a `NameGuardReport` including:
```

api/nameguard/web_api.py
```diff
@@ -131,7 +131,7 @@ async def inspect_name_get(
...
```
lightwalker-eth, line 134:
```suggestion
    If the `name` is longer than {MAX_INSPECTABLE_NAME_CHARACTERS} characters or consists of more than {MAX_INSPECTABLE_NAME_UNKNOWN_LABELS} labelhases returns `null`; else returns a `NameGuardReport` including:
```
```suggestion
    If the `name` is longer than {MAX_INSPECTABLE_NAME_CHARACTERS} characters or consists of more than {MAX_INSPECTABLE_NAME_UNKNOWN_LABELS} labelhases returns `null`; else returns a `NameGuardReport` including:
```
```suggestion
    If the `name` is longer than {MAX_INSPECTED_NAME_CHARACTERS} characters or consists of more than {MAX_INSPECTED_NAME_UNKNOWN_LABELS} labelhases returns `null`; else returns a `NameGuardReport` including:
```

lightwalker-eth (commented):
@djstrong Thanks for your updates :+1: Reviewed and shared feedback

api/nameguard/nameguard.py
```diff
@@ -142,19 +143,25 @@ def analyse_label(self, label: str):
...
-    ) -> Union[NameGuardReport, ConsolidatedNameGuardReport]:
+    ) -> Optional[Union[NameGuardReport, ConsolidatedNameGuardReport]]:
         """
         Inspect a name. A name is a sequence of labels separated by dots.
         A label can be a labelhash or a string.
         If a labelhash is encountered and `resolve_labelhashes` is `True`, a lookup will be performed.
+        Returns None if there is more labelhashes than MAX_NUMBER_OF_LABELHASHES_IN_NAME.
         """
+
+        logger.debug(f"[inspect_name] name: '{name}'")
```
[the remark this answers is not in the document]
djstrong: Not need to. This can/should be disabled in production.

djstrong:
@lightwalker-eth 
Name in uninspected:
1. If labels are going to be resolved then we count unknown labels and check if the number exceeds the limit. Unknown names are resolved.
2.  If length of the name (after optional resolving of unknown labels) exceeds the limit.

packages/nameguard-sdk/src/index.ts
```diff
@@ -499,22 +505,22 @@ class NameGuard {
    *
    * @param {string} name The name for NameGuard to inspect.
    * @param {InspectNameOptions} options The options for the inspection.
-   * @returns {Promise<NameGuardReport>} A promise that resolves with the `NameGuardReport` of the name.
+   * @returns {Promise<ConsolidatedNameGuardReport>} A promise that resolves with the `NameGuardReport` of the name.
```
djstrong, line 508: `NameGuardReport extends ConsolidatedNameGuardReport` should it be here `ConsolidatedNameGuardReport | NameGuardReport` or just `ConsolidatedNameGuardReport` is fine?

Carbon225: I would go for the union since then IDEs can suggest fields from the child class.

djstrong:
@lightwalker-eth There is a lot of problems in TS as now the API returns `ConsolidatedNameGuardReport | NameGuardReport` but I am not sure where it should be changed.

api/nameguard/provider.py
```diff
@@ -3,10 +3,16 @@
...
+provider_uris = {
+    NetworkName.MAINNET: os.environ.get('PROVIDER_URI_MAINNET'),
+    NetworkName.SEPOLIA: os.environ.get('PROVIDER_URI_SEPOLIA'),
+}
 
-async def get_nft_metadata(contract_address: str, token_id: str) -> dict:
-    url = f"{os.environ.get('PROVIDER_URI_MAINNET')}/getNFTMetadata?contractAddress={contract_address}&tokenId={token_id}&refreshCache=false"
+
+async def get_nft_metadata(network_name, contract_address: str, token_id: str) -> dict:
+    url = f'{provider_uris[network_name]}/getNFTMetadata?contractAddress={contract_address}&tokenId={token_id}&refreshCache=false'
```
lightwalker-eth, line 15: `getNFTMetadata` isn't a request that works across providers. This API is specific to Alchemy.

In the context of Ethereum, a "provider" is a system running a standard Ethereum API endpoint. Alchemy extends this with some proprietary APIs including `getNFTMetadata.

Therefore we should have separate / distinct environment variables for the Alchemy API endpoint vs the general Ethereum provider API endpoint.

We use the Alchemy API endpoint only for these NFT Metadata lookups.

We use the general Ethereum provider API endpoint for other lookups, such as what is the primary name of a given address, etc.

Please let me know if any questions :+1:

api/nameguard/models/result.py
```diff
@@ -309,7 +322,7 @@ class SecurePrimaryNameStatus(str, Enum):
     * `normalized`: The ENS primary name was found and it is normalized.
     * `no_primary_name`: The ENS primary name was not found.
     * `unnormalized`: The ENS primary name was found, but it is not normalized.
-    * `uninspected`: The name was not checked, because it is too long.
+    * `uninspected`: The name was not inspected for performance reasons.
```
lightwalker-eth, line 325:
```suggestion
    * `uninspected`: The name was exceptionally long and was not inspected for performance reasons.
```

api/nameguard/nameguard.py
```diff
@@ -143,25 +166,26 @@ def analyse_label(self, label: str):
...
     async def inspect_name(
         self, network_name: NetworkName, name: str, resolve_labelhashes: bool = True, bulk_mode: bool = False
-    ) -> Optional[Union[NameGuardReport, ConsolidatedNameGuardReport]]:
+    ) -> Union[NameGuardReport, ConsolidatedNameGuardReport]:
         """
         Inspect a name. A name is a sequence of labels separated by dots.
         A label can be a labelhash or a string.
         If a labelhash is encountered and `resolve_labelhashes` is `True`, a lookup will be performed.
-        Returns None if there is more labelhashes than MAX_INSPECTABLE_NAME_UNKNOWN_LABELS.
+        Returns ConsolidatedUninspectedNameGuardReport if name is uninspected.
```
lightwalker-eth, line 174:
```suggestion
        Returns ConsolidatedUninspectedNameGuardReport if name was exceptionally long and was not inspected for performance reasons.
```

nameguard.md
```diff
@@ -100,6 +100,8 @@ See the [NameGuard Python README](./api/README.md) for more details.
 
 12. **NameWrapper fuses**: Checks that the NameWrapper configuration of a name is safe.
 
+13. **Uninspected Name**: Checks if the name can be inspected.
```
lightwalker-eth, line 103:
```suggestion
13. **Uninspected Name**: Checks if the name is exceptionally long and would not be inspected by NameGuard for performance reasons.
```

packages/nameguard-sdk/src/index.ts
```diff
@@ -66,7 +66,10 @@ export type CheckType =
 
   // Name-level checks
   | "impersonation_risk" /** A name might be used for impersonation. */
-  | "punycode_compatible_name" /** A name is compatible with Punycode. */;
+  | "punycode_compatible_name" /** A name is compatible with Punycode. */
+  | "namewrapper_fuses" /** The NameWrapper configuration of a name is safe. */
+  | "decentralized_name" /** A name is decentralized. */
+  | "uninspected" /** A name was not inspected for performance reasons */;
```
lightwalker-eth, line 72:
```suggestion
  | "uninspected" /** A name was exceptionally long and was not inspected for performance reasons */;
```

lightwalker-eth (changes requested):
@djstrong Hey nice updates. Reviewed and shared feedback :+1: I'll follow up with additional suggestions about the data model in Slack

api/nameguard/subgraph.py
```diff
@@ -27,13 +32,18 @@
 }
 """
 
+api_key = os.environ.get('SUBGRAPH_API_KEY')
+
+headers = {'Content-Type': 'application/json', 'Authorization': f'Bearer {api_key}'}
```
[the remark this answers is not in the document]
djstrong: It works both ways, I've introduced `ENS_SUBGRAPH_URL_MAINNET` and `ENS_SUBGRAPH_URL_SEPOLIA`

packages/nameguard-sdk/src/index.ts
```diff
@@ -100,7 +103,8 @@ export enum Rating {
 export type SecurePrimaryNameStatus =
   | "normalized" /** The ENS primary name was found and it is normalized. */
   | "no_primary_name" /** The ENS primary name was not found. */
-  | "unnormalized" /** The ENS primary name was found, but it is not normalized. */;
+  | "unnormalized" /** The ENS primary name was found, but it is not normalized. */
+  | "uninspected" /** `willInspectName(name) === false` therefore the name was not inspected for performance reasons */;
```
djstrong, line 107: @lightwalker-eth `willInspectName` is hard to implement as it needs to resolve labelhashes and check length of resulted name

packages/nameguard-sdk/src/index.ts
```diff
@@ -489,7 +498,7 @@ class NameGuard {
   }
 
   /**
-   * Inspects a single name with NameGuard. Provides a `NameGuardReport` including:
+   * Inspects a single name with NameGuard. If the `name` is uninspected returns `UninspectedNameGuardReport`; else returns a `NameGuardReport` including:
```
djstrong, line 501: @lightwalker-eth Where can I define `UninspectedNameGuardReport`? It is NameGuard report with labels `null` and canonical_name `null`, rating ALERT, highest risk `uninspected`, risk count 1.

packages/nameguard-sdk/src/index.ts
lightwalker-eth: @djstrong That's fair. Thanks for feedback. I suppose we shouldn't write that function then and instead we can just have an abstract definition for the concept of what would be inspected or not and refer to that idea in our comments.

packages/nameguard-sdk/src/index.ts
lightwalker-eth: @djstrong Good question. It's a bit tricky to explain in a comment here.

Here's a simplified example of how to achieve an idea like this in TypeScript. All the below is pseudocode. I might make a typo.

Imagine we have something like the following:

```
interface AbstractMyType {
   value?: string; // NOTE: This is either `undefined` or a `string`.
   otherValue: number;
}
```

We can then also define the following types:
```
interface MyTypeWithValue implements AbstractMyType {
   value: string; // NOTE: This strips away the `undefined` case and always has a `string` value.
   otherValue: number;
}

interface MyTypeWithoutValue implements AbstractMyType {
   value?: undefined; // NOTE: This is ALWAYS undefined
   otherValue: number;
}

type MyType = MyTypeWithValue | MyTypeWithoutValue;

```

Now you can do some cool stuff with this in TypeScript. For example, you can pass in a value of `MyType` into a function. Then inside that function if you check if `value !== undefined` then the type system can know that the code following that you have a `MyTypeWithValue`.

It might be necessary to switch from defining field values as possibly being `null` to using `undefined` instead.

No worries if this is tricky. You're welcome to advance other parts of this PR and assign this step to me if you'd like :+1:

lightwalker-eth (commented):
@djstrong Good questions :+1: Replied and shared feedback

packages/nameguard-sdk/src/index.ts
lightwalker-eth: @djstrong Also, in the example above `AbstractMyType` never actually needs to be defined, but I find it nice to define that for the purpose of putting code comments on the fields in `AbstractMyType`. That way you can document all the fields in 1 place and not duplicate ideas in comments over and over again on each field of each variation of the abstract type.

djstrong: @lightwalker-eth Thank you. I implemented this:
```ts
export interface InspectedNameGuardReport extends AbstractNameGuardReport {
    labels: LabelGuardReport[];
}

export interface UninspectedNameGuardReport extends AbstractNameGuardReport {
    labels: undefined;
    canonical_name: undefined;
}

export type NameGuardReport = InspectedNameGuardReport | UninspectedNameGuardReport;
```
I don't know which is better here: `extends` or `implements`.

djstrong:
@lightwalker-eth Now, uninspected check is returned only for uninspected names. Should we return it also for all other names?

packages/nameguard-sdk/src/index.ts
lightwalker-eth: @djstrong Sure. I think you've done the right thing here with `extends` :+1:

You should update the definition of `InspectedNameGuardReport` such that it includes the following:

`canonical_name: string;`

Note here how this is not a `?string` but a `string`.

Therefore, with an `AbstractNameGuardReport` we don't know if we'll have `labels` or a `canonical_name`. With an `InspectedNameGuardReport` we're guaranteed to have both (neither can be undefined) and for an `UninspectedNameGuardReport` we're guaranteed to have neither (both must be undefined).

lightwalker-eth:
@djstrong Good question. Let's keep it the way you've built it for now that only uninspected names return the uninspected check.

Maybe we change this in the future, but happy with that for now.

packages/nameguard-sdk/src/index.ts
```diff
@@ -376,15 +376,15 @@ export interface ConsolidatedNameGuardReport extends ConsolidatedReport {
...
  * NameGuard report that contains the full results of all `checks` on all `labels` in a name.
  */
-export interface NameGuardReport extends ConsolidatedNameGuardReport {
+export interface AbstractNameGuardReport extends ConsolidatedNameGuardReport {
   /* The results of all checks performed by NameGuard on `name`. */
   checks: CheckResult[];
 
   /** Details of the inspection of all labels in `name`.
     *
     * `null` if `name` is uninspected
```
lightwalker-eth, line 385:
```suggestion
    * `undefined` if `name` was uninspected and this is an `UninspectedNameGuardReport`.
```

packages/nameguard-sdk/src/index.ts
```diff
@@ -376,15 +376,15 @@ export interface ConsolidatedNameGuardReport extends ConsolidatedReport {
...
   */
-  labels: LabelGuardReport[] | null;
+  labels?: LabelGuardReport[];
 
   /**
    * The name considered to be the canonical form of the analyzed `name`.
```
lightwalker-eth, line 390: Here's a suggestion for the line below this line. GitHub won't let me comment on the line I want to comment.

```suggestion
   * `undefined` if and only if the canonical form of `name` is considered to be undefined or `name` was uninspected and this is an ``UninspectedNameGuardReport`.
```

packages/nameguard-sdk/src/index.ts
```diff
@@ -397,9 +397,24 @@ export interface NameGuardReport extends ConsolidatedNameGuardReport {
    * `canonical_name` is guaranteed to be normalized with the exception of the case
    * where `normalization` is `unknown` and one or more labels are represented as `[labelhash]`.
    */
-  canonical_name: string | null;
+  canonical_name?: string;
+}
+
+export interface InspectedNameGuardReport extends AbstractNameGuardReport {
+    labels: LabelGuardReport[];
```
lightwalker-eth, line 404: Please see my other comment suggesting adding a definition here for the following field:

`canonical_name: string;`

packages/nameguard-sdk/src/index.ts
```diff
@@ -376,15 +376,15 @@ export interface ConsolidatedNameGuardReport extends ConsolidatedReport {
 /**
  * NameGuard report that contains the full results of all `checks` on all `labels` in a name.
  */
-export interface NameGuardReport extends ConsolidatedNameGuardReport {
+export interface AbstractNameGuardReport extends ConsolidatedNameGuardReport {
```
lightwalker-eth, line 379: Suggest we add a new field to `AbstractNameGuardReport`:

`inspected: boolean`

Then, in `InspectedNameGuardReport` define this field as:

`inspected: true`

And in `UninspectedNameGuardReport` define this field as:

`inspected: false`

Then please update the code that returns these to set the value for this field appropriately.

The benefit of this extra effort is it helps to make all the code that uses this more self documenting.

For example, a function will take as input a `NameGuardReport`. The type system will know this is either a InspectedNameGuardReport | UninspectedNameGuardReport.

Then, you can write code like the following:

```
if (report.inspected) {
  // in this code block the type system knows that `report` is an `InspectedNameGuardReport`.
} else {
  // in this code block the type system knows that `report` is an `UninspectedNameGuardReport`.
}
```

packages/nameguard-sdk/src/index.ts
```diff
@@ -397,9 +397,24 @@ export interface NameGuardReport extends ConsolidatedNameGuardReport {
...
+  canonical_name?: string;
+}
+
+export interface InspectedNameGuardReport extends AbstractNameGuardReport {
+    labels: LabelGuardReport[];
+}
+
+/**
+ * NameGuard report for a name that was exceptionally long and was not inspected for performance reasons.
+ * Fields `label` and canonical_name` are nulls, `rating` is `alert`, `highest_risk` is `uninspected`, `risk_count` is `1`.
```
lightwalker-eth, line 409:
```suggestion
 * Fields `label` and canonical_name` are undefined, `rating` is `alert`, `highest_risk` is `uninspected`, `risk_count` is `1`.
```

lightwalker-eth (changes requested):
@djstrong Hey nice updates. Reviewed and shared feedback :+1:

packages/nameguard-sdk/src/index.ts
djstrong: It still can be undefined in InspectedNameGuardReport.

packages/nameguard-sdk/src/index.ts
djstrong: @lightwalker-eth "Then please update the code that returns these to set the value for this field appropriately." - do you mean Python or TS code?

lightwalker-eth: @djstrong Ok cool. The goal of that statement is to ensure the value is set in Python code and then ultimately flows correctly into the TS code.

packages/nameguard-sdk/src/index.ts
djstrong: It is documented in AbstractNameGuard:
```ts
  /**
   * The name considered to be the canonical form of the analyzed `name`.
   *
   * `undefined` if and only if the canonical form of `name` is considered to be undefined or `name` was uninspected and this is an ``UninspectedNameGuardReport`.
   *
   * If a label is represented as `[labelhash]` in `name`,
   * the `canonical_name` will also contain the label represented as `[labelhash]`.
   *
   * `canonical_name` is guaranteed to be normalized with the exception of the case
   * where `normalization` is `unknown` and one or more labels are represented as `[labelhash]`.
   */
  canonical_name?: string;
```

api/nameguard/web_api.py
```diff
@@ -52,7 +57,15 @@ class ApiVersion(str, Enum):
 ⚠️ Beta Version: Please be aware that NameGuard is currently in beta. We appreciate your feedback as we work towards a stable v1 release. Expect changes to the API to address community feedback and improve functionality.
 
 ## Documentation
-These documentation pages focus specifically on the NameGuard API. For information on the NameGuard Library, SDK, and UI Kit, please refer to the [NameGuard GitHub repository](https://github.com/namehash/nameguard).""",
+These documentation pages focus specifically on the NameGuard API. For information on the NameGuard Library, SDK, and UI Kit, please refer to the [NameGuard GitHub repository](https://github.com/namehash/nameguard).
+
+## Glossary
+
+* uninspected name - if labels are going to be resolved then we count unknown labels and check if the number exceeds the limit `MAX_INSPECTED_NAME_UNKNOWN_LABELS`. Unknown names are resolved.
```
lightwalker-eth, line 64:
```suggestion
* uninspected name - any name that fails any of the following requirements:
  1. The count of unknown labels (distinct or non-distinct) must not exceed `MAX_INSPECTED_NAME_UNKNOWN_LABELS`.
  2. The count of characters in the name (after potential resolving of unknown labels) must not exceed the limit `MAX_INSPECTED_NAME_CHARACTERS`.
```

api/nameguard/web_api.py
```diff
@@ -52,7 +57,15 @@ class ApiVersion(str, Enum):
...
+If length of the name (after optional resolving of unknown labels) exceeds the limit `MAX_INSPECTED_NAME_CHARACTERS`.
```
lightwalker-eth, line 65: Please see the suggested change to this line in the comment above.

lightwalker-eth (changes requested):
@djstrong Hey nice updates :+1: Reviewed and shared feedback

api/nameguard/subgraph.py
```diff
@@ -27,13 +32,18 @@
 }
 """
 
+api_key = os.environ.get('SUBGRAPH_API_KEY')
```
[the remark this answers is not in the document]
djstrong: You are right!

.github/workflows/ci_sdk.yml
```diff
@@ -52,4 +52,4 @@ jobs:
         run: curl http://0.0.0.0:8000/inspect-name/mainnet/test.eth
 
       - name: Run tests
-        run: NAMEGUARD_TEST_ENDPOINT=http://0.0.0.0:8000 pnpm test -r
+        run: NAMEGUARD_API_URI=http://0.0.0.0:8000 pnpm test -r
```
lightwalker-eth, line 55: Am I correct to understand that this is passing in `NAMEGUARD_API_URI` as an environment variable to `pnpm test -r`?

If so, what do you think about updating our syntax here so that we pass in the environment variable similar to how we do it here: https://github.com/namehash/namekit/blob/main/.github/workflows/ci_sdk.yml#L56-L57

.changeset/two-horses-move.md
```diff
@@ -0,0 +1,5 @@
+---
+"@namehash/nameguard": minor
+---
+
+introduced uninspected NameGuard reports
```
lightwalker-eth, line 5:
```suggestion
Introduced uninspected NameGuard reports for performance optimization of exceptionally long names.
```

api/tests/test_api.py
```diff
@@ -225,19 +225,20 @@ def test_bulk_inspect_name_post(test_client):
...
                 'title',
+                'inspected',
             }
             for r in res_json['results']
         ]
     )
 
-    # more than 250 names
+    # more than MAX_NUMBER_OF_NAMES_IN_BULK names
     response = test_client.post('/bulk-inspect-names', json={'names': names * 126})
```
lightwalker-eth, line 235: Is this really more than `MAX_NUMBER_OF_NAMES_IN_BULK`? I see the number `126` here.

lightwalker-eth (changes requested):
@djstrong Hey super updates here :+1: Just a few very small feedbacks

packages/nameguard-sdk/src/index.ts
```diff
@@ -66,7 +66,10 @@ export type CheckType =
 
   // Name-level checks
   | "impersonation_risk" /** A name might be used for impersonation. */
-  | "punycode_compatible_name" /** A name is compatible with Punycode. */;
+  | "punycode_compatible_name" /** A name is compatible with Punycode. */
+  | "namewrapper_fuses" /** The NameWrapper configuration of a name is safe. */
+  | "decentralized_name" /** A name is decentralized. */
+  | "uninspected" /** A name was exceptionally long and was not inspected for performance reasons */;
```
lightwalker-eth, line 72: @notrab Hey, we've added 3 new potential checks here. Can you please investigate and confirm if we need to update anything in `nameguard-react`, `examples.nameguard.io` or `nameguard.io`?

packages/nameguard-sdk/src/index.ts
```diff
@@ -100,7 +103,8 @@ export enum Rating {
 export type SecurePrimaryNameStatus =
   | "normalized" /** The ENS primary name was found and it is normalized. */
   | "no_primary_name" /** The ENS primary name was not found. */
-  | "unnormalized" /** The ENS primary name was found, but it is not normalized. */;
+  | "unnormalized" /** The ENS primary name was found, but it is not normalized. */
+  | "uninspected" /** A name was exceptionally long and was not inspected for performance reasons */;
```
lightwalker-eth, line 107: @notrab Hey, we've added a new potential `SecurePrimaryNameStatus` here. Can you please investigate and confirm if we need to update anything in `nameguard-react`, `examples.nameguard.io` or `nameguard.io`?

packages/nameguard-sdk/src/index.ts
```diff
@@ -372,27 +376,50 @@ export interface ConsolidatedNameGuardReport extends ConsolidatedReport {
...
   /* The results of all checks performed by NameGuard on `name`. */
   checks: CheckResult[];
 
-  /** Details of the inspection of all labels in `name`. */
-  labels: LabelGuardReport[];
+  /** Details of the inspection of all labels in `name`.
+    *
+    * `undefined` if `name` was uninspected and this is an `UninspectedNameGuardReport`.
+  */
+  labels?: LabelGuardReport[];
```
lightwalker-eth, line 390: @notrab Hey, this field now might be undefined. Can you please investigate and confirm if we need to update anything in `nameguard-react`, `examples.nameguard.io` or `nameguard.io`?

packages/nameguard-sdk/src/index.ts
```diff
@@ -372,27 +376,50 @@ export interface ConsolidatedNameGuardReport extends ConsolidatedReport {
...
+   * `undefined` if and only if the canonical form of `name` is considered to be undefined or `name` was uninspected and this is an ``UninspectedNameGuardReport`.
    *
    * If a label is represented as `[labelhash]` in `name`,
    * the `canonical_name` will also contain the label represented as `[labelhash]`.
    *
    * `canonical_name` is guaranteed to be normalized with the exception of the case
    * where `normalization` is `unknown` and one or more labels are represented as `[labelhash]`.
    */
-  canonical_name: string | null;
+  canonical_name?: string;
```
lightwalker-eth, line 403: @notrab Hey, this field now might be undefined rather than null. Can you please investigate and confirm if we need to update anything in `nameguard-react`, `examples.nameguard.io` or `nameguard.io`?

packages/nameguard-sdk/src/index.ts
```diff
@@ -372,27 +376,50 @@ export interface ConsolidatedNameGuardReport extends ConsolidatedReport {
...
+ * NameGuard report for a name that was exceptionally long and was not inspected for performance reasons.
+ * Fields `label` and canonical_name` are undefined, `rating` is `alert`, `highest_risk` is `uninspected`, `risk_count` is `1`.
+ */
+export interface UninspectedNameGuardReport extends AbstractNameGuardReport {
+    labels: undefined;
+    canonical_name: undefined;
+    inspected: false;
+}
+
+export type NameGuardReport = InspectedNameGuardReport | UninspectedNameGuardReport;
```
lightwalker-eth, line 421: @notrab Hey, please check how we've defined the types here for `NameGuardReport`, `InspectedNameGuardReport`, and `UninspectedNameGuardReport`.

The goal for doing this was to make it so that in the case you need to support both the inspected and uninspected cases you can define a field as type `NameGuardReport`. Then, if you write some code where you do something like:

```
const myFunction = (report: NameGuardReport) => void {
   if (report.inspected) {
      // Here TypeScript will know that `report` is of type `InspectedNameGuardReport`.
   } else {
      // Here TypeScript will know that `report` is of type `UninspectedNameGuardReport`.
   }
}
```

This means we might have cases in our code where we should not define a variable as `NameGuardReport` but instead might want to define it as type `InspectedNameGuardReport`, such that the type system enforces us to first verify that the report we're passing in was indeed inspected.

Can you please investigate and confirm that all code we have in `nameguard-react`, `examples.nameguard.io` or `nameguard.io` is updated based on this concept?

lightwalker-eth (changes requested):
@djstrong @Carbon225 @notrab Hey guys, added a few more comments for you to ensure we have everything in this PR optimized for our whole platform including all packages and apps.

Appreciate your advice!

.github/workflows/ci_sdk.yml
djstrong: Yes.

api/tests/test_api.py
djstrong: Yes, `len(names)==2` so it is 252. I have added assert.

packages/nameguard-sdk/src/index.ts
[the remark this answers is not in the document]
djstrong: `ConsolidatedNameGuardReport` should also have `inspected` field so we also need separate types for inspected and uninspected report. Maybe we need some multi inheritance.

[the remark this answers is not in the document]
djstrong: The issue is resolved.

lightwalker-eth (commented):
Latest changes all look great :+1:

api/nameguard/nameguard.py
lightwalker-eth: @djstrong Hey, I missed this comment from before. Is there any remaining open question here?

djstrong: No, now we don't return null but uninspected report.

djstrong:
> Latest changes all look great 👍

@lightwalker-eth So we are waiting for answers from notrab?

packages/nameguard-sdk/src/index.ts
```diff
@@ -365,34 +369,57 @@ export interface ConsolidatedNameGuardReport extends ConsolidatedReport {
...
-   * `null` if and only if the canonical form of `name` is considered to be undefined.
+   * `undefined` if and only if the canonical form of `name` is considered to be undefined or `name` was uninspected and this is an ``UninspectedNameGuardReport`.
    *
    * If a label is represented as `[labelhash]` in `name`,
    * the `canonical_name` will also contain the label represented as `[labelhash]`.
    *
    * `canonical_name` is guaranteed to be normalized with the exception of the case
    * where `normalization` is `unknown` and one or more labels are represented as `[labelhash]`.
    */
-  canonical_name: string | null;
```
lightwalker-eth, line 393: @djstrong Two ideas here:
1. Can we please document here what cases cause `canonical_name` to be undefined?
2. When we document these cases, can we please include reference to the new `inspected` field?

packages/nameguard-sdk/src/index.ts
```diff
@@ -448,6 +475,8 @@ const DEFAULT_NETWORK: Network = "mainnet";
 const DEFAULT_INSPECT_LABELHASH_PARENT = ETH_TLD;
 export const DEFAULT_COMPUTE_NAMEGUARD_REPORT = false;
 const MAX_BULK_INSPECTION_NAMES = 250;
+export const MAX_INSPECTED_NAME_CHARACTERS = 200;  // includes label separators
+const MAX_INSPECTED_NAME_UNKNOWN_LABELS = 5;
```
lightwalker-eth, line 479: @djstrong can we make a little comment here how even duplicated instances of the same unknown label are counted each time they appear in a name

packages/nameguard-sdk/src/index.ts
```diff
@@ -490,7 +519,7 @@ export class NameGuard {
   }
 
   /**
-   * Inspects a single name with NameGuard. Provides a `NameGuardReport` including:
+   * Inspects a single name with NameGuard. If the `name` is uninspected returns `UninspectedNameGuardReport`; else returns a `InspectedNameGuardReport` including:
```
lightwalker-eth, line 522: @djstrong I think it will be nice to update this to more formally refer to the new `inspected` field rather than the abstract idea of being inspected / uninspected.

packages/nameguard-sdk/src/index.ts
```diff
@@ -515,7 +544,7 @@ export class NameGuard {
 
   // TODO: Document how this API will attempt automated labelhash resolution through the ENS Subgraph.
   /**
-   * Inspects up to 250 names at a time with NameGuard. Provides a `ConsolidatedNameGuardReport` for each name provided in `names`, including:
+   * Inspects up to 250 names at a time with NameGuard. Provides `UninspectedNameGuardReport` if the `name` is uninspected; else returns a `ConsolidatedNameGuardReport` for each name provided in `names`, including:
```
lightwalker-eth, line 547: @djstrong I think it will be nice to update this to more formally refer to the new `inspected` field rather than the abstract idea of being inspected / uninspected.

packages/nameguard-sdk/src/index.ts
```diff
@@ -554,7 +583,7 @@ export class NameGuard {
    * Inspects the name associated with a namehash.
    *
    * NameGuard will attempt to resolve the name associated with the namehash through the ENS Subgraph.
-   * If this resolution succeeds then NameGuard will generate and return a `NameGuardReport` for the name.
+   * If this resolution succeeds then NameGuard will return a `NameGuardReport` for the name. If the resolved `name` is uninspected then NameGuard will return `UninspectedNameGuardReport`.
```
lightwalker-eth, line 586: @djstrong I think it will be nice to update this to more formally refer to the new `inspected` field rather than the abstract idea of being inspected / uninspected.

packages/nameguard-sdk/src/index.ts
```diff
@@ -601,6 +630,8 @@ export class NameGuard {
    *   1. The labelhash of the "childmost" label of a name.
    *   2. The complete parent name of the "childmost" label.
    *
+   * Returns `UninspectedNameGuardReport` if the resolved `name` is uninspected.
```
lightwalker-eth, line 633: @djstrong I think it will be nice to update this to more formally refer to the new `inspected` field rather than the abstract idea of being inspected / uninspected.

packages/nameguard-sdk/src/index.ts
```diff
@@ -365,34 +369,57 @@ export interface ConsolidatedNameGuardReport extends ConsolidatedReport {
...
+}
+
+export interface InspectedNameGuardReport extends AbstractNameGuardReport {
+    labels: LabelGuardReport[];
+    inspected: true;
 }
 
+/**
+ * NameGuard report for a name that was exceptionally long and was not inspected for performance reasons.
+ * Fields `label` and canonical_name` are undefined, `rating` is `alert`, `highest_risk` is `uninspected`, `risk_count` is `1`.
```
lightwalker-eth, line 413: I believe you can remove this entire line in the comment. Instead of writing these ideas in the comment, you can explicitly identify these fields in the interface definition here.

For example, just as you defined `inspected: false;` here, I believe you can also do the same for `rating`, `highest_risk`, and `risk_count`.

packages/nameguard-sdk/src/index.ts
```diff
@@ -448,6 +475,8 @@ const DEFAULT_NETWORK: Network = "mainnet";
 const DEFAULT_INSPECT_LABELHASH_PARENT = ETH_TLD;
 export const DEFAULT_COMPUTE_NAMEGUARD_REPORT = false;
 const MAX_BULK_INSPECTION_NAMES = 250;
+export const MAX_INSPECTED_NAME_CHARACTERS = 200;  // includes label separators
```
lightwalker-eth, line 478: It's best to write comments like these in /* ... */ style above the related code. This is nice because then IDEs such as VSCode will show that comment anywhere in the code where you hover over the related definition.

packages/nameguard-sdk/src/index.ts
```diff
@@ -365,34 +369,57 @@ export interface ConsolidatedNameGuardReport extends ConsolidatedReport {
   /* The ENSIP-15 normalization status of `name` */
   normalization: Normalization;
 
-  /** Beautified version of `name` */
+  /* Beautified version of `name` */
   beautiful_name: string;
+
+  /* Whether or not the name was inspected. */
```
lightwalker-eth, line 375:
```suggestion
  /* Whether or not the name was inspected. If `false` then the name was exceptionally long and was not inspected for performance reasons */
```

api/nameguard/web_api.py
```diff
@@ -52,7 +58,16 @@ class ApiVersion(str, Enum):
...
 ## Documentation
-These documentation pages focus specifically on the NameGuard API. For information on the NameGuard Library, SDK, and UI Kit, please refer to the [NameGuard GitHub repository](https://github.com/namehash/nameguard).""",
+These documentation pages focus specifically on the NameGuard API. For information on the NameGuard Library, SDK, and UI Kit, please refer to the [NameKit GitHub repository](https://github.com/namehash/namekit).
+
+## Glossary
+
+* uninspected name - any name that fails any of the following requirements:
+  1. The count of unknown labels (distinct or non-distinct) must not exceed `MAX_INSPECTED_NAME_UNKNOWN_LABELS`.
+  2. The count of characters in the name (after potential resolving of unknown labels) must not exceed the limit `MAX_INSPECTED_NAME_CHARACTERS`.
+If name is uninspected an UninspectedNameGuardReport is returned.
```
lightwalker-eth, line 68: It would be nice to explicitly refer to the `inspected` field here.

api/nameguard/web_api.py
```diff
@@ -123,7 +138,7 @@ async def inspect_name_get(
     """
     ## Inspects a single name with NameGuard.
 
-    Provides a `NameGuardReport` including:
+    If the `name` is uninspected returns `UninspectedNameGuardReport`; else returns a `NameGuardReport` including:
```
lightwalker-eth, line 141: It would be nice to explicitly refer to the `inspected` field here.

api/nameguard/web_api.py
```diff
@@ -173,7 +188,7 @@ async def inspect_name_post(request: InspectNameRequest) -> NameGuardReport:
...
```
lightwalker-eth, line 191: It would be nice to explicitly refer to the `inspected` field here.

api/nameguard/web_api.py
```diff
@@ -218,7 +233,7 @@ async def bulk_inspect_names(request: BulkInspectNamesRequest) -> BulkNameGuardB
     """
     ## Inspects up to 250 names at a time with NameGuard.
 
-    Provides a `ConsolidatedNameGuardReport` for each name provided in `request.names`, including:
+    Provides `UninspectedNameGuardReport` if the `name` is uninspected; else returns a `ConsolidatedNameGuardReport` for each name provided in `request.names`, including:
```
lightwalker-eth, line 236: It would be nice to explicitly refer to the `inspected` field here.

api/nameguard/web_api.py
```diff
@@ -281,7 +296,7 @@ async def inspect_namehash_get(
     NameGuard will attempt to resolve the name associated with the `namehash` through the ENS Subgraph,
     using the network specified in `network_name`.
 
-    If this resolution succeeds then NameGuard will generate and return a `NameGuardReport` for the name.
+    If this resolution succeeds then NameGuard will return a `NameGuardReport` for the name. If the resolved `name` is uninspected then NameGuard will return `UninspectedNameGuardReport`.
```
lightwalker-eth, line 299: It would be nice to explicitly refer to the `inspected` field here.

api/nameguard/web_api.py
```diff
@@ -314,7 +329,7 @@ async def inspect_namehash_post(request: InspectNamehashRequest) -> NameGuardRep
     NameGuard will attempt to resolve the name associated with the `request.namehash` through the ENS Subgraph,
     using the network specified in `request.network_name`.
 
-    If this resolution succeeds then NameGuard will generate and return a `NameGuardReport` for the name.
+    If this resolution succeeds then NameGuard will return a `NameGuardReport` for the name. If the resolved `name` is uninspected then NameGuard will return `UninspectedNameGuardReport`.
```
lightwalker-eth, line 332: It would be nice to explicitly refer to the `inspected` field here.

api/nameguard/web_api.py
```diff
@@ -372,6 +387,8 @@ async def inspect_labelhash_get(
     1. The labelhash of the "childmost" label of a name.
     2. The complete parent name of the "childmost" label.
 
+    Returns `UninspectedNameGuardReport` if the resolved name is uninspected.
```
lightwalker-eth, line 390: It would be nice to explicitly refer to the `inspected` field here.

api/nameguard/web_api.py
```diff
@@ -436,6 +453,8 @@ async def inspect_labelhash_post(request: InspectLabelhashRequest) -> NameGuardR
...
```
lightwalker-eth, line 456: It would be nice to explicitly refer to the `inspected` field here.

api/nameguard/models/result.py
```diff
@@ -351,7 +379,7 @@ class SecurePrimaryNameResult(BaseModel):
         examples=['vitalik.eth'],
     )
 
-    nameguard_result: Optional[NameGuardReport] = Field(
+    nameguard_result: Optional[Union[NameGuardReport, ConsolidatedNameGuardReport]] = Field(
```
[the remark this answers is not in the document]
djstrong: It wasn't reverted.

api/nameguard/models/result.py
```diff
@@ -385,9 +413,10 @@ class FakeEthNameCheckResult(BaseModel):
 
     status: FakeEthNameCheckStatus
 
-    nameguard_result: Optional[NameGuardReport] = Field(
+    nameguard_result: Optional[Union[NameGuardReport, UninspectedNameGuardReport]] = Field(
```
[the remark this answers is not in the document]
djstrong: Yes, it is not necessary.

api/nameguard/models/result.py
```diff
@@ -235,17 +241,37 @@ class NameGuardReport(ConsolidatedNameGuardReport):
...
         examples=['vitalik.eth'],
     )
 
+    inspected: Literal[True] = Field(
+        description='Whether the name was inspected.',
+        examples=[True],
+    )
+
+
+class UninspectedNameGuardReport(NameGuardReport):
```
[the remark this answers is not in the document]
djstrong: It more of a type and validation schema.

djstrong:
I have changed `UninspectedNameGuardReport` to `NameGuardReport` with `inspected` field equal to `false`.

packages/nameguard-sdk/src/index.ts
djstrong: I don't have idea how.

djstrong:
@lightwalker-eth I don't have idea how to "refer to the new inspected field rather than the abstract idea of being inspected / uninspected". The field can't be in a condition and a result.

packages/nameguard-sdk/src/index.ts
```diff
@@ -583,13 +589,16 @@ export class NameGuard {
...
+   * If this resolution succeeds then NameGuard will return a `NameGuardReport` for the name. 
    * If this resolution fails then NameGuard will return an error.
    *
    * @param {string} namehash A namehash should be a decimal or a hex (prefixed with 0x) string.
    * @param {InspectNamehashOptions} options The options for the inspection.
-   * @returns {Promise<NameGuardReport>}  A promise that resolves with a `NameGuardReport` of the resolved name.
+   * @returns {Promise<NameGuardReport>} A promise that resolves with the `NameGuardReport` of the name. Check the `inspected` field of the result to determine if the result is an `InspectedNameGuardReport` or an `UninspectedNameGuardReport` for performance reasons in the case that the provided `name` was exceptionally long.
+   * @throws {NameGuardError} If the inspection fails due to network issues, server errors, or if the namehash cannot be resolved to a name.
+   * @throws {Error} If the provided namehash is not in a valid Keccak256 hash format.
```
lightwalker-eth, line 599:
```suggestion
   * @throws {Error} If the provided `namehash` is not in a valid Keccak256 hash format.
```

lightwalker-eth (changes requested):

lightwalker-eth (approved):

PR #288, opened 2024-06-13
feat: Price migration to `namekit-react`

FrancoAguzzi:
Related ShortCut stories:

1. [Transition <DisplayedPrice> and related components to NameKit](https://app.shortcut.com/ps-web3/story/25297/transition-displayedprice-and-related-components-to-namekit)
2. [Migrate PriceInfo and DisplayedName to NameKit React](https://app.shortcut.com/ps-web3/story/25249/migrate-priceinfo-and-displayedname-to-namekit-react)
3. [Transition PriceSymbol to `ui` repo](https://app.shortcut.com/ps-web3/story/24801/transition-pricesymbol-to-ui-repo)


apps/storybook/ postcss.config.js
lightwalker-eth, line 1: Does this file name have a space in it?

apps/storybook/package.json
```diff
@@ -2,30 +2,33 @@
   "name": "nameguard-storybook",
   "private": true,
   "scripts": {
+    "build": "npm run tailwind && storybook build",
     "dev": "storybook dev -p 6006",
-    "build": "storybook build"
+    "tailwind": "tailwindcss -i ./styles/input.css -o ./styles/output.css"
```
lightwalker-eth, line 7: Can we please document in the README.md file what this command is for and the goal it is intended to achieve?

packages/namekit-react/src/components/WethSymbol.tsx
lightwalker-eth, line 1: Please feel welcome to push back on this! Just curious. Should we move all the components that are just icons into a subfolder named /icons? Appreciate your advice :+1:

packages/namekit-react/src/components/PriceInfo.tsx
```diff
@@ -0,0 +1,80 @@
...
+          <InfoFilledIcon className="h-5 w-5 text-gray-300" />
+        </div>
+        <p className="text-sm text-gray-500 ml-3">
+          <div className="inline mr-2">{priceDescriptionElm}</div>
+        </p>
+      </div>
+    );
+};
+
+export function InfoFilledIcon(props: SVGProps<SVGSVGElement>) {
```
lightwalker-eth, line 62: Can we move this into a separate .tsx file that's dedicated just for this icon?

Also, is there any special reason we call this "InfoFilledIcon"? Why is it "Filled"? Is there any special reason not to just call it "InfoIcon"?

packages/namekit-react/src/components/EthereumCircleIcon.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { type SVGProps } from "react";
+
+interface LogoProps extends SVGProps<SVGSVGElement> {
+  isSmallIcon?: boolean;
```
lightwalker-eth, line 4: Hmm.. my first impression is this prop feels like code smell. Can you please investigate the purpose of this prop in more detail and suggest a good solution for us? Thanks

apps/storybook/tailwind.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  content: ['./**/*.{js,jsx,ts,tsx}'],
+  // Toggle dark-mode based on .dark class or data-mode="dark"
```
lightwalker-eth, line 3: Is this comment correct?

apps/storybook/styles/output.css
lightwalker-eth, line 1: Is it right for us to be adding this file into git?

packages/namekit-react/src/components/PriceSymbol.tsx
```diff
@@ -0,0 +1,88 @@
+import { PriceCurrencyFormat, Currency } from "@namehash/ens-utils";
+import { Tooltip } from "@namehash/nameguard-react";
```
lightwalker-eth, line 2: What do you think about moving `Tooltip` into NameKit react? Overall I'd prefer that NameGuard react only includes things that are truly nameguard-only.

packages/namekit-react/src/components/PriceSymbol.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+import EthereumCircleIcon from "./EthereumCircleIcon";
```
lightwalker-eth, line 4: Why does this import statement not have a {} while the others do?

packages/namekit-react/src/components/PriceSymbol.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+export enum PriceSymbology {
+  Acronym = "Acronym",
+  Symbol = "Symbol",
+}
+
+interface PriceSymbolProps {
+  currency: Currency;
+  size: PriceDisplaySize;
+  symbolFillColor?: string;
```
lightwalker-eth, line 35: Could you please document what the purpose of this `symbolFillColor` prop is for? Are we actually using it somewhere? What goal does it help to achieve?

packages/namekit-react/src/components/PriceSymbol.tsx
```diff
@@ -0,0 +1,88 @@
...
+  showTooltipDescription: boolean;
```
lightwalker-eth, line 36:
```suggestion
  describeCurrencyInTooltip: boolean;
```

packages/namekit-react/src/components/PriceSymbol.tsx
```diff
@@ -0,0 +1,88 @@
...
+};
+
+export const PriceSymbol = ({
+  size,
+  currency,
+  symbolFillColor,
+  showTooltipDescription,
+}: PriceSymbolProps) => {
+  const symbol =
+    currency === Currency.Usd ? (
```
lightwalker-eth, line 53: Could we make use of a switch statement here? I prefer to use switch where we want to ensure we have an answer for every possible enum value. with switch the type system can help to raise an error if we miss some case. Just ensure you don't use `default` in the switch. All cases should be explicit.

packages/namekit-react/src/components/PriceSymbol.tsx
```diff
@@ -0,0 +1,88 @@
...
+      />
+    ) : currency === Currency.Weth ? (
+      <WethSymbol
+        className={SymbolSize[size]}
+        fill={symbolFillColor || "inherit"}
+      />
+    ) : currency === Currency.Eth ? (
+      <EthereumCircleIcon
+        className={SymbolSize[size]}
+        fill={symbolFillColor ? symbolFillColor : "#272727"}
```
lightwalker-eth, line 75: What is special about this color value? Can we please make a named constant for it and consider any special details here in how we document the `symbolFillColor` prop?

packages/namekit-react/src/components/PriceInfo.tsx
```diff
@@ -0,0 +1,80 @@
...
+} from "@namehash/ens-utils";
+
+interface PriceInfoProps {
+  parsedName: DomainName;
+  registration: Registration;
+  // Wether to display the icon with a tooltip or the icon + the price info text directly
+  displayIconWithTooltip?: boolean;
+}
+
+export const PriceInfo = ({
```
lightwalker-eth, line 16: Suggest we make this name more specific. For example maybe `RegistrationPriceInfo`?

packages/namekit-react/src/components/PriceInfo.tsx
```diff
@@ -0,0 +1,80 @@
...
+  getPriceDescription,
+  Registration,
+  DomainName,
+} from "@namehash/ens-utils";
+
+interface PriceInfoProps {
+  parsedName: DomainName;
+  registration: Registration;
+  // Wether to display the icon with a tooltip or the icon + the price info text directly
+  displayIconWithTooltip?: boolean;
```
lightwalker-eth, line 13: Maybe we should change this from a boolean to an enum where each enum value is more self documenting and descriptive?

packages/namekit-react/src/components/DisplayedPrice.tsx
```diff
@@ -0,0 +1,213 @@
...
+  PriceSymbolPosition,
+  PriceDisplayPosition,
+  PriceDisplaySize,
+  PriceSymbology,
+} from "./PriceSymbol";
+
+import cc from "classcat";
+import { Tooltip } from "@namehash/nameguard-react";
+
+interface PriceProps {
```
lightwalker-eth, line 20:
```suggestion
interface DisplayedPriceProps {
```

packages/namekit-react/src/components/DisplayedPrice.tsx
```diff
@@ -0,0 +1,213 @@
...
+  // The price to be displayed
+  price: Price;
+  // The alternative price to be displayed (the price converted to another currency)
+  altPrice?: Price;
+  // The onClick event handler for the price text
+  onTextClick?: () => void;
+  // Wether to display the alternative price as a text instead of a tooltip
+  altPriceDisplayedAsText: boolean;
```
lightwalker-eth, line 28: Suggest converting this from a boolean into an enum where each enum value is more self documenting and descriptive.

lightwalker-eth (commented):
@FrancoAguzzi Excited to see the progress here :+1: Reviewed and shared some initial feedback

apps/storybook/ postcss.config.js
FrancoAguzzi: No!

packages/namekit-react/src/components/PriceInfo.tsx
FrancoAguzzi: Renamed and moved this component 🙂👍🏼

packages/namekit-react/src/components/PriceSymbol.tsx
FrancoAguzzi: I think it is a good idea!

packages/namekit-react/src/components/PriceSymbol.tsx
FrancoAguzzi: In storybook there is one example of this usage: we use it for displaying the symbol in the same color as the text in alternative price when gray... Both examples of a gray ethereum icon and a blue ethereum icon, which represents the ability of changing the symbol color, can be found in storybook docs, please take a look and let me know your thoughts! 😃

packages/namekit-react/src/components/PriceInfo.tsx
FrancoAguzzi: Great idea

packages/namekit-react/src/components/NamePrice.tsx
```diff
@@ -0,0 +1,82 @@
...
+  altPriceDisplaySize?: PriceDisplaySize;
+  // The size of the price display
+  priceTextDisplaySize?: PriceDisplaySize;
+  // The place to display the alternative price when displayed as text
+  altPriceDisplayPosition?: PriceDisplayPosition;
+  // Wether or not to display the name of the currency in a tooltip when its symbol is hovered
+  showCurrencyTooltipDescription?: boolean;
+}
+
+export const NamePrice = ({
```
lightwalker-eth, line 40: I'm not sure if this UI component should even exist? If we do keep it I'm open to that, but then it should be made more generic.

For example, when I test this UI component in the storybook it is always right aligned, there's a big rectangle with that highlights with a grey background on hover, there's a clicky-hand cursor, etc.. These ideas shouldn't be in this UI component if it is generic.

We need to put more special care and attention into how we build our own abstractions. We should try to avoid creating abstractions that mix ideas from too many different layers at the same time. For example here we're defining a UI component that has too much of an expectation that it is only placed inside a DomainCard.

apps/storybook/stories/Namekit/DisplayedPrice.stories.tsx
```diff
@@ -0,0 +1,299 @@
...
+  AltPriceDisplayFormat,
+  DisplayedPrice,
+  PriceDisplayPosition,
+  PriceDisplaySize,
+  PriceSymbolPosition,
+  PriceSymbology,
+} from "@namehash/namekit-react";
+import type { Meta, StoryObj } from "@storybook/react";
+
+const meta: Meta<typeof DisplayedPrice> = {
```
lightwalker-eth, line 12: Can we please make it so that Storybook is better configured to work with our types and the overall type system?

For example, how complicated is it to make it so that the controls for these are typed? For example, currency as a dropdown where only valid currency options can be input. Etc.. If this is complicated please let me know.

CleanShot 2024-06-16 at 13 05 10

apps/storybook/stories/Namekit/DisplayedPrice.stories.tsx
```diff
@@ -0,0 +1,299 @@
...
+  component: DisplayedPrice,
+  title: "Namekit/DisplayedPrice",
+};
+
```
lightwalker-eth, line 16: Is my memory correct that we also have defined the concept of a maximum displayed price in each currency? It would be great to show those examples in storybook for each currency.

Additionally, depending on the currency, there can be cases where the price is > 0 but it displays as as some minimum displayed price. It would also be great to add those examples here.

lightwalker-eth (changes requested):
@FrancoAguzzi Good to see the updates here. Reviewed and shared feedback :+1:

apps/storybook/README.md
```diff
@@ -0,0 +1,5 @@
+## NameKit React storybook documentation
+
+### Styling with TailwindCSS
+
+In `package.json`, a command called `tailwind` is defined: this command creates an `output.css` file which basically contains Tailwind CSS specs that are used whenever a Tailwind class is referenced! This `output.css` file is imported in `/.storybook/preview.js` and, this way, makes Tailwind specs available throughout the Storybook context. This command is also referenced in the `build` command since it is essential that our Tailwind specs are up to date to the Tailwind version we have imported in the app.
```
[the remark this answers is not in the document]
FrancoAguzzi: @lightwalker-eth @notrab since I have removed Tailwind from this codebase, as per an alignment with @notrab, this README will be changed to not include the above info!

apps/storybook/stories/Namekit/DisplayedPrice.stories.tsx
FrancoAguzzi: @lightwalker-eth I manage to improve a lot the props typing but the object props could not be set different

apps/storybook/stories/Namekit/DisplayedPrice.stories.tsx
FrancoAguzzi: Added for both ETH and USD prices!

apps/storybook/README.md
```diff
@@ -1,5 +0,0 @@
-## NameKit React storybook documentation
```
lightwalker-eth: Is it really right to fully delete the ideas here? Should they move to a readme under namekit-react and/or in other package readme files?

apps/storybook/stories/Namekit/DisplayedPrice.stories.tsx
```diff
@@ -297,3 +331,35 @@ export const WithOnClickHandler: Story = {
     onTextClick: () => alert("Price clicked"),
   },
 };
+export const EthMinDisplayPrice: Story = {
```
lightwalker-eth, line 334: Can we please add stories for the min and max of each currency? Thanks

apps/storybook/stories/Namekit/RegistrationPriceInfo.stories.tsx
```diff
@@ -1,33 +1,39 @@
...
     },
     parsedName: {
       namehash:
-        "0xc4feb2ef28ae93ad2847cc7aaa8d614509f29feea35c86555550f071e26cec83",
-      slug: "xtest$1709652875$$$false.eth",
-      displayName: "xtest$1709652875$$$false.eth",
-      normalizedName: "xtest$1709652875$$$false.eth",
-      labelName: "xtest$1709652875$$$false",
+        "0x54f4b08bc10ab9368020f2afdd830ab742058481586c4ae87ba6e37c8912a6ec",
+      slug: "expiredDomain.eth",
```
lightwalker-eth, line 31: A lot of the values in this file are `expiredDomain.eth` with an uppercase D which is unnormalized. Can you please only use normalized name examples in this storybook? Please also update namehash / etc.. values as needed for this change too.

apps/storybook/stories/Namekit/RegistrationPriceInfo.stories.tsx
```diff
@@ -1,33 +1,39 @@
...
+  now,
 } from "@namehash/ens-utils";
 import type { Meta, StoryObj } from "@storybook/react";
 import {
   RegistrationPriceInfo,
   RegistrationPriceInfoDisplayingFormat,
 } from "@namehash/namekit-react";
 
+const ELEVEN_DAYS_FROM_NOW_TIMESTAMP = buildTimestamp(
+  BigInt(new Date().getTime()) / 1005n,
```
lightwalker-eth, line 14: I'm really confused by the logic here. Why 11 days? Eleven days from now would be 11 days in the future, not in the past. Why aren't we using `now()` here? Why are we dividing by a magic number here? Etc..

apps/storybook/stories/Namekit/CurrencySymbol.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+  },
+};
+export const WithCustomSymbolColor: Story = {
+  args: {
+    symbolFillColor: "#007bff",
+    size: PriceDisplaySize.Large,
+    currency: Currency.Eth,
+  },
+};
+export const ShowingTooltipDescription: Story = {
```
lightwalker-eth, line 66: Why does this story cause the symbol to become horizontally aligned center?

apps/storybook/stories/Namekit/CurrencySymbol.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+    size: PriceDisplaySize.Large,
+  },
+};
+export const DAI: Story = {
+  args: {
+    currency: Currency.Dai,
+    size: PriceDisplaySize.Large,
+  },
+};
+export const MicroSize: Story = {
```
lightwalker-eth, line 35: MicroSize and SmallSize appear exactly the same to me on my laptop. This seems confusing.

apps/storybook/stories/Namekit/CurrencySymbol.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+    currency: Currency.Eth,
+  },
+};
+export const MediumSize: Story = {
+  args: {
+    size: PriceDisplaySize.Medium,
+    currency: Currency.Eth,
+  },
+};
+export const LargeSize: Story = {
```
lightwalker-eth, line 53: MediumSize and LargeSize appear exactly the same to me on my laptop.

apps/storybook/stories/Namekit/DisplayedPrice.stories.tsx
```diff
@@ -0,0 +1,365 @@
...
+      value: 198900n,
+    },
+    altPrice: {
+      currency: Currency.Eth,
+      value: 1000000000000000000n,
+    },
+    currencySymbology: CurrencySymbology.Acronym,
+  },
+};
+export const LargeAltPriceDisplaySize: Story = {
```
lightwalker-eth, line 265: The vertical alignment for all the Alt Price stories is broken.

CleanShot 2024-07-01 at 03 24 24

apps/storybook/stories/Namekit/DisplayedPrice.stories.tsx
```diff
@@ -0,0 +1,365 @@
...
+    altPrice: {
+      currency: Currency.Eth,
+      value: 1000000000000000000n,
+    },
+    priceTextDisplaySize: PriceDisplaySize.Large,
+    altPriceDisplayPosition: PriceDisplayPosition.Right,
+    altPriceDisplayFormat: AltPriceDisplayFormat.Text,
+  },
+};
+export const AlternativePriceDisplayedAsTextInBottom: Story = {
```
lightwalker-eth, line 120: Why does this change the horizontal alignment of the UI component?

apps/storybook/stories/Namekit/RegistrationPriceInfo.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+      labelHash:
+        "0x9458d05651ff8443f025747bf1a859bb26f2815f70d68e8cc9286027013961cd",
+      unwrappedTokenId:
+        113026375855800814118284943830039256141780833603018686252863399931967234376006n,
+      wrappedTokenId:
+        89103332435335212332047307598497361224309437240941958630163555473262172236931n,
+    },
+  },
+};
+export const DisplayTextAndIcon: Story = {
```
lightwalker-eth, line 44: In the output of this story the font styling of the price value appears to be heavier weight than all the other output text.

The weight of all the output text should match. The price shouldn't be heavier.

packages/namekit-react/src/components/Tooltip.tsx
```diff
@@ -25,7 +25,7 @@ export function Tooltip({
   const handleClose = () => setOpen(false);
```
lightwalker-eth, line 26: Something has broken with the tooltips on the NameGuard website. Please ensure everything is working.

CleanShot 2024-07-01 at 03 32 17

packages/namekit-react/src/components/icons/DaiSymbol.tsx
```diff
@@ -0,0 +1,28 @@
...
+    <svg
+      {...props}
+      viewBox="0 0 16 16"
+      xmlns="http://www.w3.org/2000/svg"
+      fill={props.fill ? props.fill : "inherit"}
+    >
+      <g clipPath="url(#clip0_12879_99850)">
+        <path
+          d="M8.00078 0.799805C11.9776 0.799805 15.2008 4.02364 15.2008 7.9998C15.2008 11.9766 11.9776 15.1998 8.00078 15.1998C4.02461 15.1998 0.800781 11.9763 0.800781 7.9998C0.800781 4.02364 4.02461 0.799805 8.00078 0.799805Z"
+          fill={props.fill ? props.fill : "#F5AC37"}
```
notrab, line 14: @FrancoAguzzi another topic for discussion.

packages/namekit-react/src/components/DisplayedPrice.tsx
```diff
@@ -0,0 +1,222 @@
...
+      {(currencySymbology === PriceSymbology.Symbol &&
+        altPriceDisplayFormat === AltPriceDisplayFormat.Tooltip &&
+        !altPrice) ||
+      (currencySymbology === PriceSymbology.Symbol &&
+        altPriceDisplayFormat === AltPriceDisplayFormat.Text) ? (
+        <div
+          className={cc([
+            "nk-leading-none",
+            {
+              "nk-mb-[1px]": price.currency !== Currency.Usd,
```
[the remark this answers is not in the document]
FrancoAguzzi: @lightwalker-eth as aligned, documentation about Tailwind prefixes will be done at the end of code development

apps/storybook/stories/Namekit/RegistrationPriceInfo.stories.tsx
```diff
@@ -0,0 +1,82 @@
...
+      expiryTimestamp: { time: 1709652875n },
+      primaryStatus: PrimaryRegistrationStatus.Expired,
+      secondaryStatus: SecondaryRegistrationStatus.RecentlyReleased,
+      registrationTimestamp: { time: 1711380875n },
+      expirationTimestamp: { time: 1709652875n },
+    },
+    parsedName: {
+      namehash:
+        "0xc4feb2ef28ae93ad2847cc7aaa8d614509f29feea35c86555550f071e26cec83",
+      slug: "xtest$1709652875$$$false.eth",
```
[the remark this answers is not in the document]
FrancoAguzzi: @lightwalker-eth as aligned, documentation about domains in premium period will be done at the end of code development

apps/storybook/stories/Namekit/NamePrice.stories.tsx
```diff
@@ -0,0 +1,206 @@
...
+  primaryStatus: PrimaryRegistrationStatus.Expired,
+  secondaryStatus: SecondaryRegistrationStatus.RecentlyReleased,
+  registrationTimestamp: { time: 1711380875n },
+  expirationTimestamp: { time: 1709652875n },
+};
+
+const defaultParsedName = {
+  namehash:
+    "0xc4feb2ef28ae93ad2847cc7aaa8d614509f29feea35c86555550f071e26cec83",
+  slug: "xtest$1709652875$$$false.eth",
```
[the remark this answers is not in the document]
FrancoAguzzi: @lightwalker-eth as aligned, documentation about domains in premium period will be done at the end of code development

packages/namekit-react/src/components/CurrencySymbol.tsx
```diff
@@ -1,7 +1,7 @@
 import { PriceCurrencyFormat, Currency } from "@namehash/ens-utils";
 import { Tooltip } from "./Tooltip";
 
-import { EthereumCircleIcon } from "./icons/EthereumCircleIcon";
+import { EthSymbol } from "./icons/EthSymbol";
 import { UsdcSymbol } from "./icons/UsdcSymbol";
 import { WethSymbol } from "./icons/WethSymbol";
 import { DaiSymbol } from "./icons/DaiSymbol";
```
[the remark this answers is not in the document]
FrancoAguzzi: Moved this and other `Price` related declarations to `DisplayedPrice` file

packages/namekit-react/src/components/CurrencySymbol.tsx
```diff
@@ -29,12 +29,12 @@ export enum PriceDisplayPosition {
   Bottom = "nk-flex nk-flex-col nk-text-right nk-items-end nk-space-y-1",
```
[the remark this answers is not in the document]
FrancoAguzzi: Moved this and other Price related declarations to DisplayedPrice file

packages/namekit-react/src/components/DisplayedPrice.tsx
```diff
@@ -28,26 +28,26 @@ interface DisplayedPriceProps {
...
   // The place to display the alternative price when displayed as text
   altPriceDisplayPosition?: PriceDisplayPosition;
   // Wether or not to display the name of the currency in a tooltip when its symbol is hovered
-  showCurrencyTooltipDescription?: boolean;
+  describeCurrencyInTooltip?: boolean;
 }
 
 export const DisplayedPrice = ({
   price,
   altPrice,
```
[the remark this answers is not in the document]
FrancoAguzzi: 100%

Will do this after main code development part is done!

packages/namekit-react/src/components/DisplayedPrice.tsx
```diff
@@ -28,26 +28,26 @@ interface DisplayedPriceProps {
...
   onTextClick,
```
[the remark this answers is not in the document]
FrancoAguzzi: Semantically, `DisplayedPrice` is a text. The `onTextClick` possibility is an additional feature for integrators that want to give it functionality. Your idea is good but does not match all use cases, e.g. plain text `DisplayedName`

packages/ens-utils/src/price.ts
```diff
@@ -3,23 +3,11 @@ import {
   PriceCurrencyFormat,
   parseStringToCurrency,
 } from "./currency";
-import { DomainCard, DomainName } from "./domain";
+import { DomainName } from "./domain";
 import { MIN_ETH_REGISTRABLE_LABEL_LENGTH } from "./ensname";
```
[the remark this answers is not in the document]
FrancoAguzzi: This is used to determine a name price, I understand this is not generic to price but related to `ensname.ts`, should I move `AvailableNamePriceUSD` from `price.ts` to there?

packages/ens-utils/src/price.ts
```diff
@@ -3,23 +3,11 @@ import {
...
+import { nameCurrentTemporaryPremium } from "./ethregistrar";
```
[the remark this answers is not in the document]
FrancoAguzzi: This is used to determine a name price, I understand this is not generic to price but related to ensname.ts, should I move AvailableNamePriceUSD from price.ts to there?

packages/ens-utils/src/price.ts
```diff
@@ -3,23 +3,11 @@ import {
...
-import {
-  Duration,
-  SECONDS_PER_DAY,
-  Timestamp,
-  addSeconds,
-  buildDuration,
-  formatTimestampAsDistanceToNow,
-  now,
-} from "./time";
+import { Registration } from "./registration";
```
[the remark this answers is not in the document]
FrancoAguzzi: This is used to determine a name price, I understand this is not generic to price but related to `ensname.ts`, should I move `AvailableNamePriceUSD` from `price.ts` to there?

apps/storybook/README.md
FrancoAguzzi: This README.md was added when another approach was used to include Tailwind classes into Storybook, which is now not used anymore. Previously from this version of the README.md, no README existed in this package. 

Would you like me to rewrite this to a descriptive and guided introduction of the Storybook components library after main code development work is done?

FrancoAguzzi: This README.md was added when another approach was used to include Tailwind classes into Storybook, which is now not used anymore. Previously from this version of the README.md, no README existed in this package.

Would you like me to rewrite this to a descriptive and guided introduction of the Storybook components library after main code development work is done?

apps/storybook/stories/Namekit/CurrencySymbol.stories.tsx
FrancoAguzzi: Indeed these have the same width and height and this spec came from artboards: 

Uploading Gravação de Tela 2024-07-10 às 21.25.56.mov…

FrancoAguzzi: https://github.com/namehash/nameguard/assets/49823133/c1cb4fca-15c3-4c7f-a703-1f31655a84d6

apps/storybook/stories/Namekit/CurrencySymbol.stories.tsx
FrancoAguzzi: These in fact are differently sized both in Storybook and Figma:

FrancoAguzzi: https://github.com/namehash/nameguard/assets/49823133/675cff4f-b5c9-4aac-8167-25870ebfe189

packages/namekit-react/src/components/Tooltip.tsx
FrancoAguzzi: @lightwalker-eth this is related to Tailwind prefixes appliance, I will review this with @notrab asap!

apps/storybook/stories/Namekit/DisplayedPrice.stories.tsx
FrancoAguzzi: @lightwalker-eth it seems to me we need to review the use cases we want to support related to price size:

The spacing different is due to the line-height difference of these small & big texts, respectively -- IDK if someone would want to display alt. price such big and main price such small -- maybe it is a good idea to understand which pairs of sizes we want to support and make sure these are correctly displayed instead of creating multiple conditionals, which would be needed, in order to align these frankstein use cases?

e.g. we give support for prices in size X and alt. prices in size X, X - 1 and X - 2, meaning --- we give support for price in Large size! With this, be sure alt. price Large, Medium and Small size are also supported! 

Besides from this use case where the alt. price is displayed bigger than the main price, the previous comments where you identified two sizes with same icon size motivated me to think we should do a little formatting on this component's specification --- What do you think?

apps/storybook/stories/Namekit/RegistrationPriceInfo.stories.tsx
FrancoAguzzi: I've also added a date tooltip to the period related displayed text

apps/storybook/stories/Namekit/DisplayedPrice.stories.tsx
FrancoAguzzi: I centered everything now

packages/namekit-react/src/components/Tooltip.tsx
notrab: We should have to import the styles into the NameGuard website for the package that is now exporting the component for it to continue working as it were

apps/storybook/stories/Namekit/NamePrice.stories.tsx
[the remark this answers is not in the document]
lightwalker-eth: @FrancoAguzzi The feedback shared above doesn't classify as "documentation" in my mind.

packages/namekit-react/src/components/DisplayedPrice.tsx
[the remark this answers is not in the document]
lightwalker-eth: @FrancoAguzzi But if I want to make a DisplayedPrice clickable, why wouldn't I just nest it inside of a <a or <Link tag?

@notrab Could you please share your advice here too?

apps/storybook/README.md
lightwalker-eth: @FrancoAguzzi Ah ok. Then it's ok to just remove the readme here for now. If you think there's an important idea to communicate of course please feel welcome to include it in the readme

lightwalker-eth: @FrancoAguzzi Ah ok. Then it's ok to just remove the readme here for now. If you think there's an important idea to communicate of course please feel welcome to include it in the readme

apps/storybook/stories/Namekit/CurrencySymbol.stories.tsx
lightwalker-eth: @FrancoAguzzi I see that the icon size is the same, but the font size / styling is different.

How do you suggest we manage this?

In my mind, the icon should only have 2 size enum values if it only actually has 2 displayed sizes.

apps/storybook/stories/Namekit/CurrencySymbol.stories.tsx
lightwalker-eth: @FrancoAguzzi Yes because you're looking at a `DisplayedPrice` here, but my feedback above was for a `CurrencySymbol`.

Maybe there actually is a difference but only for the font and only for the currency symbol `$` which needs to match the font of the `DisplayedPrice`?

apps/storybook/stories/Namekit/DisplayedPrice.stories.tsx
lightwalker-eth: @FrancoAguzzi Open to your suggestions. I'm not a CSS guy so I don't have much advice to share.

Without thinking of CSS technical details, from an end-user perspective, I just want all the text to feel like it is sitting on the same "table" such that the vertical alignment of the text appears consistent.

packages/namekit-react/src/components/icons/DaiSymbol.tsx
lightwalker-eth: @notrab Could you please share more details for the code review comments? Thank you :+1:

notrab: I think we should use `fill=currentColor` instead of the hex here and just control the colour by the className on the svg component itself.

apps/storybook/stories/Namekit/CurrencySymbol.stories.tsx
FrancoAguzzi: It is a layout configuration we can decide wether to put the Ui comps :)

Do you like it this way?

apps/storybook/stories/Namekit/CurrencySymbol.stories.tsx
FrancoAguzzi: That's correct !

I suggest we map every PriceDisplaySize to an icon display size, even if these are the same across different display size options. 

The idea behind having only 2 enum options would unfortunately not fit the complete idea of this component since PriceDisplaySize have 4 diff options. Supporting only 2 **requires** changing the type of this prop for full support. We'd then change the mapping of the chosen DisplayedPrice size from PriceDisplaySize --> PriceDisplaySize to PriceDisplaySize --> string. The abstraction of today reuses the same PriceDisplaySize type in the communication between DisplayedPrice and CurrencySymbol. The new option would do the other mapping mentioned 

That's correct. once again!

(respectively answering ;))

apps/storybook/stories/Namekit/CurrencySymbol.stories.tsx
FrancoAguzzi: Ah, true!


Correct! This is the PriceDisplaySize mapping to classname we currently have:

Captura de Tela 2024-07-11 às 22 06 11

Note how micro and small are the same size as medium and large are,

Key references to help rethink sizes: Figma [artboards](https://www.figma.com/design/2e3p7KEFikJjTPXdRyOWfZ/Namekit?node-id=6289-10398&m=dev) and size mapping shared above :)

apps/storybook/stories/Namekit/DisplayedPrice.stories.tsx
FrancoAguzzi: Perfect, @lightwalker-eth 

As you, all users would likely want the same!

This is another business decision to be done! Such as [this one](https://github.com/namehash/nameguard/pull/288#discussion_r1674904531)

I think using alt price bigger than main price does not make sense so I like the suggestion I gave in my previous comment 😅 @notrab could you read it as well and give as your opinion? @lightwalker-eth I also encourage you to opinate on this price + alt price displaying strategies to be open-sourced!

⛓

packages/namekit-react/src/components/icons/DaiSymbol.tsx
FrancoAguzzi: @notrab I agree it is nice idea

Centralizes props. What if some component has multiple fill rules such as WethSymbol, in case the user wants to change the icon color, do you think about this class overwriting all the inline fill HTML props?

packages/namekit-react/src/components/icons/DaiSymbol.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { type SVGProps } from "react";
+
+export const DaiSymbol = (props: SVGProps<SVGSVGElement>) => {
+  return (
+    <svg
+      {...props}
+      viewBox="0 0 16 16"
+      xmlns="http://www.w3.org/2000/svg"
+      fill={props.fill ? props.fill : "inherit"}
```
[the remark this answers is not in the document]
FrancoAguzzi: @notrab share your thoughts with me, ma boy 👂🏻

Are these the same as above: using classes to style instead of fill prop?

packages/namekit-react/src/components/Tooltip.tsx
FrancoAguzzi: Resolved, thank you @notrab !

packages/namekit-react/src/components/icons/DaiSymbol.tsx
notrab: @FrancoAguzzi valid point, I would vote to hardcode some values, or figure out if the SVG can be simplified for transparency, or just leave as is. My only concern in these use cases is that some icons have more customization options than others, which often leads to unwanted side effects.

packages/namekit-react/src/components/DisplayedPrice.tsx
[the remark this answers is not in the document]
lightwalker-eth: @FrancoAguzzi Hey great questions :+1:

1. Possibly. In my thinking, if some situation wanted the DisplayedPrice to be clickable, that should be achieved through an <a or a <Link that wraps the DisplayedPrice. This is outside my expertise though so please push back on this if I'm missing something. Something just doesn't feel right to me when we keep defining all these custom onClick handlers. My intuition is that this is not the right path.
2. Same as above.
3. They can be the same.

I think we'll need to put some boundaries on the customizations we support on these UI components. Without some boundaries the scope and complexity for us grows too much. Additionally, that complexity can make the UI components less easy to use or for people to feel comfortable using the library because it seems too complex.

There's a natural tension between these competing goals.

A good strategy can be to write our code in more composable ways. For example, if we have:

1. A UI component for displaying a primary price
2. A UI component for displaying an alt price
3. A UI component that wraps both of the above to display both a primary and alt price.

Then, in the usual case, people can just use UI component 3. In a more advanced case where someone wants to write different click handlers for primary price vs alt price, they can still make use of 1 and 2.

I appreciate that a refactoring as described here adds scope. If you think this direction makes sense then please create a new and distinct shortcut story for this as we can come back to it later in a separate PR.

Appreciate your advice :+1:

packages/namekit-react/src/components/PriceSymbol.tsx
lightwalker-eth: @FrancoAguzzi Based on the discussion you've been having with Jamie on ideas related to this, please let me know your advice on how we proceed.

apps/storybook/stories/Namekit/CurrencySymbol.stories.tsx
lightwalker-eth: @FrancoAguzzi I think it's better to have everything with a default (left) alignment in all storybook stories.

apps/storybook/stories/Namekit/CurrencySymbol.stories.tsx
lightwalker-eth: @FrancoAguzzi Thanks. Please see my related comment about introducing a `CurrencySymbolSize` enum.

apps/storybook/stories/Namekit/DisplayedPrice.stories.tsx
lightwalker-eth: @FrancoAguzzi Not sure if I understand your question here. I won't be able to help much with the CSS specifics.

If your question is to ask if we can assume a constraint where the alt price is never bigger than the primary price, sure, please proceed with that.

Please work backwards from the goal here: We need to ensure the alignment issue here is fixed. If fixing this means we need to add in a constraint somewhere else, please introduce that constraint and document it.

packages/namekit-react/src/components/PriceSymbol.tsx
FrancoAguzzi: For now, no updates will be done to our handling of SVGs

I will do another effort around this, after the main DomainCard components are ready

What do you think > [Refine use of HTML in SVG icons](https://app.shortcut.com/ps-web3/story/25228/refine-use-of-html-in-svg-icons)?

apps/storybook/stories/Namekit/RegistrationPriceInfo.stories.tsx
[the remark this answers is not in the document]
FrancoAguzzi: OK, fair enough

I will look for a good example and update this story example 🫱🏻‍🫲🏾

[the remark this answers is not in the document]
FrancoAguzzi: So, what I have done is:

1. Fulfilled `RegistrationPriceInfo` Storybook stories with `frankind.eth` `DomainName` objects
2. Made `Registration` props objects that align to the desired use case we want to represent in Storybook

packages/namekit-react/src/components/icons/DaiSymbol.tsx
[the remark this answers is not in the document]
FrancoAguzzi: OK good! Let's do this job inside this story? WDYT? > [Refine use of HTML in SVG icons](https://app.shortcut.com/ps-web3/story/25228/refine-use-of-html-in-svg-icons)

I can take the lead on raising up the discussion with you later on the roadmap

packages/namekit-react/src/components/icons/DaiSymbol.tsx
FrancoAguzzi: Very good point. This indeed can happen: let's take by example, Wrapped Ethereum icon, which represents a multicolor Ethereum icon... How to deal with these multi-colors? Well, nowadays we paint the whole thing as one color if user informed the color, otherwise, we leave it as native. It surely can be better set... I think it is a good opportunity to do these SVG dealing enhancements in this story: [Refine use of HTML in SVG icons](https://app.shortcut.com/ps-web3/story/25228/refine-use-of-html-in-svg-icons)

apps/storybook/stories/Namekit/DisplayedPrice.stories.tsx
FrancoAguzzi: Sounds perfect, @lightwalker-eth

packages/namekit-react/src/components/DisplayedPrice.tsx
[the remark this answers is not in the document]
FrancoAguzzi: I really appreciate your advices shared above, @lightwalker-eth 

Understanding more the time precision, I think it is great idea to wrap these in a Shortcut story and replicate the idea, in code format, soon.

Goal for now is, then, to make DisplayedPrice perfect for merge,
Next goal is to properly create the possibility of destructuring DisplayedPrice into single UI components and maturing these single implementation to the space where these can be used in pairs as well.

This is the [Story](https://app.shortcut.com/ps-web3/story/25520/break-displayedprice-into-two-single-price-components) for next steps 😊

packages/namekit-react/src/components/PriceSymbol.tsx
lightwalker-eth: @FrancoAguzzi I'm ok to handle this in a separate story.

packages/namekit-react/src/components/DisplayedPrice.tsx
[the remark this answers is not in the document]
lightwalker-eth: @FrancoAguzzi Thanks for that. We can consider this goal resolved for this PR then.

apps/storybook/stories/Namekit/CurrencySymbol.stories.tsx
lightwalker-eth: @FrancoAguzzi This is still an open question on this PR.

apps/storybook/stories/Namekit/CurrencySymbol.stories.tsx
lightwalker-eth: @FrancoAguzzi This is still an open question on this PR.

apps/storybook/stories/Namekit/DisplayedPrice.stories.tsx
lightwalker-eth: @FrancoAguzzi Ok. Looking forward to seeing your proposal for fixing this.

packages/namekit-react/src/components/icons/DaiSymbol.tsx
lightwalker-eth: @FrancoAguzzi I'm ok to consider this idea as resolved for this PR and for follow up separately in the story you linked above.

packages/namekit-react/src/components/icons/DaiSymbol.tsx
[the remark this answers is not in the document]
lightwalker-eth: @FrancoAguzzi I'm ok to consider this idea as resolved for this PR and for follow up separately in the story you linked above.

PR #289, opened 2024-06-14
feat(namekit-react): Button

notrab:
This PR adds a `Button` component. Styles don't match our branding yet, just setting the foundation for this PR. Button can be used with the usual `asChild` prop so we can use the same variants for TextLinks.

CleanShot 2024-06-15 at 10 49 29@2x

CleanShot 2024-06-14 at 21 14 59@2x
CleanShot 2024-06-14 at 21 14 57@2x
CleanShot 2024-06-14 at 21 14 55@2x

### DIFF
--- a/apps/storybook/stories/Namekit/Button.stories.tsx
+++ b/apps/storybook/stories/Namekit/Button.stories.tsx
@@ -104,7 +104,7 @@ export const CustomClass: Story = {
     variant: "primary",
     size: "medium",
     children: "Custom Class Button",
-    className: "my-custom-class",
+    className: "custom-class-name",
   },
 };
 
--- a/apps/storybook/stories/Namekit/Button.stories.tsx
+++ b/apps/storybook/stories/Namekit/Button.stories.tsx
@@ -9,7 +9,7 @@ const meta: Meta<typeof Button> = {
     variant: {
       control: {
         type: "select",
-        options: ["primary", "secondary"],
+        options: ["primary", "secondary", "ghost"],
       },
     },
     size: {
@@ -43,6 +43,14 @@ export const SecondaryMedium: Story = {
   },
 };
 
+export const GhostMedium: Story = {
+  args: {
+    variant: "ghost",
+    size: "medium",
+    children: "Ghost Medium Button",
+  },
+};
+
 export const PrimarySmall: Story = {
   args: {
     variant: "primary",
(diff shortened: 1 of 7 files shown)


packages/namekit-react/src/components/Button.tsx
```diff
@@ -0,0 +1,65 @@
...
+  ghost: "nk-text-black nk-border-transparent hover:nk-bg-black/5",
+};
+
+const sizeClasses = {
+  small: "nk-py-1 nk-px-2 nk-text-sm",
+  medium: "nk-py-2 nk-px-4",
+  large: "nk-py-3 nk-px-6 nk-text-lg",
+};
+
+export const Button: React.FC<ButtonProps> = ({
```
lightwalker-eth, line 29: @notrab Great to see the updates here. My first impression is that the default styles here look pretty similar to what we have in Figma. Can you confirm if we've aligned with those 100%?

packages/namekit-react/tailwind.config.ts
```diff
@@ -3,6 +3,14 @@ import type { Config } from "tailwindcss";
 const config: Config = {
   content: ["src/**/*.{ts,tsx,js,jsx,mdx}"],
   prefix: "nk-",
+  theme: {
+    extend: {
+      colors: {
```
lightwalker-eth, line 8: Is this the best practice for us to manage colors?

apps/storybook/stories/Namekit/Button.stories.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+export const GhostLarge: Story = {
+  args: {
+    variant: "ghost",
+    size: "large",
+    children: "Secondary Large Button",
+  },
+};
+
+export const CustomClass: Story = {
```
lightwalker-eth, line 102: Could we customize this one in a more visually obvious way?

lightwalker-eth (changes requested):
@notrab Great to see this. Just a few small questions. Reviewed and shared feedback :+1:

packages/namekit-react/tailwind.config.ts
notrab: @lightwalker-eth indeed. This colour isn't a Tailwind default, so we need to add it here so we can use it. [Name That Color](https://chir.ag/projects/name-that-color) is a great method for naming colors so I used that here.

packages/namekit-react/src/components/Button.tsx
notrab: Yeah, should be inline with Figma.

packages/namekit-react/tailwind.config.ts
lightwalker-eth: @notrab ok awesome, thank you :+1:

lightwalker-eth (approved):
@notrab looks great! Thank you 🚀

PR #290, opened 2024-06-14
feat(namekit-react): input/textarea

notrab:
Similar to Button, this abstracts some of the input/textarea components so we can set once and reuse our styles (pending!)

CleanShot 2024-06-14 at 21 27 49@2x
CleanShot 2024-06-14 at 21 27 46@2x
CleanShot 2024-06-14 at 21 27 44@2x


lightwalker-eth (changes requested):
@notrab Great to see these! Thank you :+1: Reviewed and shared feedback. It will also be good to add a changeset :+1:

packages/namekit-react/src/components/Input.tsx
```diff
@@ -0,0 +1,38 @@
...
+  secondary: "nk-border-gray-200 nk-bg-gray-100",
+};
+
+const sizeClasses = {
+  small: "nk-py-1 nk-px-2 nk-text-sm",
+  medium: "nk-py-2 nk-px-4",
+  large: "nk-py-3 nk-px-6 nk-text-lg",
+};
+
+export const Input: React.FC<InputProps> = ({
```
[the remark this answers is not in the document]
notrab: There is now a `<Input.Slot />` component you can pass anythign to which will expand the input to fit the contents. We can pass icons, text, etc.

packages/namekit-react/src/components/ENSTextarea.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+import { Textarea, TextareaProps } from "./Textarea";
+
+const ensInputClasses = "ens-webfont";
+
+export const ENSTextarea: React.FC<TextareaProps> = (props) => {
+  return (
+    <Textarea
+      {...props}
+      className={`${props.className || ""} ${ensInputClasses}`}
```
lightwalker-eth, line 11: Do we need to use `cc` in a situation like this? Appreciate your advice.

packages/namekit-react/src/components/ENSInput.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+import { Input, InputProps } from "./Input";
+
+const ensInputClasses = "ens-webfont";
+
+export const ENSInput: React.FC<InputProps> = (props) => {
+  return (
+    <Input
+      {...props}
+      className={`${props.className || ""} ${ensInputClasses}`}
```
lightwalker-eth, line 11: Do we need to use `cc` in a situation like this? Appreciate your advice.

lightwalker-eth (commented):
@notrab Two small questions. Main item that needs your help is how the storybook build is not working so not able to fully test this.

packages/namekit-react/src/components/ENSTextarea.tsx
notrab: Yeah I'll switch it out to use `cc`, thanks for pointing out.

PR #291, opened 2024-06-14
feat(namekit-react): link component

notrab:
CleanShot 2024-06-14 at 21 35 59

### DIFF
--- a/apps/storybook/stories/Namekit/Link.stories.tsx
+++ b/apps/storybook/stories/Namekit/Link.stories.tsx
@@ -102,7 +102,6 @@ export const SecondaryExternal: Story = {
 export const CustomClass: Story = {
   args: {
     variant: "primary",
-    size: "medium",
     href: "#",
     children: "Custom Class Link",
     className: "custom-class-name",
@@ -112,7 +111,6 @@ export const CustomClass: Story = {
 export const AsChild: Story = {
   args: {
     variant: "primary",
-    size: "medium",
     href: "#",
     children:
       "Link as a Span (which won't work, but is good for next/link instances)",
@@ -120,11 +118,23 @@ export const AsChild: Story = {
   },
 };
 
+export const CurrentColor: Story = {
+  args: {
+    variant: "primary",
+    children: (
+      <>
+        External Link
+        <Link.ExternalIcon />
+      </>
+    ),
+    asChild: <span style={{ color: "green" }} />,
+  },
+};
+
 export const ExternalLinkIconAfter: Story = {
   args: {
     variant: "primary",
-    size: "medium",
-    href: "#",
+    href: "https://namekit.io",
     children: (
       <>
         External Link
@@ -138,7 +148,7 @@ export const ExternalLinkIconBefore: Story = {
   args: {
     variant: "primary",
     size: "medium",
-    href: "#",
+    href: "https://namekit.io",
     children: (
       <>
         <Link.ExternalIcon />
--- a/apps/storybook/stories/Namekit/Link.stories.tsx
+++ b/apps/storybook/stories/Namekit/Link.stories.tsx
@@ -1,0 +1,105 @@
+import React from "react";
+import type { Meta, StoryObj } from "@storybook/react";
+import { Link } from "@namehash/namekit-react";
+
+const meta: Meta<typeof Link> = {
+  title: "UI/Link",
+  component: Link,
+  argTypes: {
+    variant: {
+      control: {
+        type: "select",
+        options: ["primary", "secondary"],
+      },
+    },
+    size: {
+      control: {
+        type: "select",
+        options: ["small", "medium", "large"],
+      },
+    },
+    className: { control: "text" },
+    href: { control: "text" },
+    target: { control: "text" },
+    rel: { control: "text" },
+  },
+};
+
+export default meta;
+
+type Story = StoryObj<typeof Link>;
+
+export const PrimaryMedium: Story = {
+  args: {
+    variant: "primary",
+    size: "medium",
+    href: "#",
+    children: "Primary Medium Link",
+  },
+};
+
+export const SecondaryMedium: Story = {
+  args: {
+    variant: "secondary",
+    size: "medium",
+    href: "#",
+    children: "Secondary Medium Link",
+  },
+};
+
+export const PrimarySmall: Story = {
+  args: {
+    variant: "primary",
+    size: "small",
+    href: "#",
+    children: "Primary Small Link",
+  },
+};
+
+export const SecondarySmall: Story = {
+  args: {
+    variant: "secondary",
+    size: "small",
+    href: "#",
+    children: "Secondary Small Link",
+  },
+};
+
+export const PrimaryLarge: Story = {
+  args: {
+    variant: "primary",
+    size: "large",
+    href: "#",
+    children: "Primary Large Link",
+  },
+};
+
+export const SecondaryLarge: Story = {
+  args: {
+    variant: "secondary",
+    size: "large",
+    href: "#",
+    children: "Secondary Large Link",
+  },
+};
+
+export const CustomClass: Story = {
+  args: {
+    variant: "primary",
+    size: "medium",
+    href: "#",
+    children: "Custom Class Link",
+    className: "my-custom-class",
+  },
+};
+
+export const AsChild: Story = {
+  args: {
+    variant: "primary",
+    size: "medium",
+    href: "#",
+    children:
+      "Link as a Span (which won't work, but is good for next/link instances)",
+    asChild: <span />,
+  },
+};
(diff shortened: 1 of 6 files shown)


apps/storybook/stories/Namekit/Link.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+export const SecondaryLarge: Story = {
+  args: {
+    variant: "secondary",
+    size: "large",
+    href: "#",
+    children: "Secondary Large Link",
+  },
+};
+
+export const CustomClass: Story = {
```
lightwalker-eth, line 86: Could we make it more visually obvious how the custom class example has been customized?

lightwalker-eth (commented):
@notrab Looks good! Approved! Just 1 small request and also could we please add a changeset?

apps/storybook/stories/Namekit/Link.stories.tsx
notrab: Done

packages/namekit-react/src/components/Link.tsx
```diff
@@ -0,0 +1,72 @@
...
+};
+
+const sizeClasses = {
+  small: "nk-text-sm",
+  medium: "nk-text-base",
+  large: "nk-text-lg",
+};
+
+const isInternalLink = (href: string | undefined) => {
+  if (!href) return false;
```
lightwalker-eth, line 26: Is there any problem if a link is a "javascript link" rather than a "html link"?

For example, we can consider the "on click override" examples here: https://www.nameguard.io/docs/report

CleanShot 2024-06-17 at 12 56 33

The above link is handled in Javascript rather than HTML.

Appreciate your advice on what we should do :+1: Also very open to the possibility that we need to make changes to how UI components such as `ReportBadge` and `ReportIcon` work :+1:

Please note that Franco is very close to merging a PR that makes a number of other changes to the UI components on https://www.nameguard.io/docs/report

* Link to PR: https://github.com/namehash/nameguard/pull/282
* Link to Preview Environment: https://nameguard-pnqz2xot4-namehash.vercel.app/docs/report

packages/namekit-react/src/components/Link.tsx
```diff
@@ -0,0 +1,72 @@
...
+        className,
+        asChild.props.className,
+      ]),
+    };
+
+    return React.cloneElement(asChild, childProps);
+  }
+
+  return (
+    <a className={combinedClassName} {...props}>
```
lightwalker-eth, line 68: Whenever we show the external link icon, I believe it should also be true that:

* We add `target="_blank" rel="noopener noreferrer"`

I shared another comment describing cases where a link is technically an external link but we don't want to show an external link icon. For these cases perhaps it's best not to automatically apply these properties and instead require them to be manually added if desired. This is why I began this comment by saying "Whenever we show an external link icon" rather than saying "Whenever this is technically an external link".

Appreciate your advice :+1:

lightwalker-eth (changes requested):
@notrab Looks awesome! Reviewed and shared some feedback 😄

packages/namekit-react/src/components/Link.tsx
```diff
@@ -0,0 +1,72 @@
...
+export interface LinkProps extends React.ComponentPropsWithoutRef<"a"> {
+  className?: string;
+  variant?: "primary" | "secondary";
+  size?: "small" | "medium" | "large";
+  asChild?: React.ReactElement;
+}
+
+const linkBaseClasses = "nk-transition";
+
+const variantClasses = {
```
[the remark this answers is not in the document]
notrab: @lightwalker-eth I think for cases like this, `className` override is perfect because the Link is being customised for the context its within.

packages/namekit-react/src/components/Link.tsx
notrab: @lightwalker-eth yeah I was in two minds on this one. I'll push some updates.

packages/namekit-react/src/components/Link.tsx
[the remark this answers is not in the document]
notrab: @lightwalker-eth I think what might be a better solution for `primary` variant is to make this `text-current` so it inherits the current colour.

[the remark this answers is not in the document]
lightwalker-eth: @notrab Those sound like nice ideas :+1:

packages/namekit-react/src/components/Link.tsx
```diff
@@ -0,0 +1,72 @@
...
+}) => {
+  const combinedClassName = cc([
+    linkBaseClasses,
+    variantClasses[variant],
+    sizeClasses[size],
+    className,
+  ]);
+
+  const enhancedChildren = !isInternalLink(props.href) ? (
+    <>{children} ↗</>
```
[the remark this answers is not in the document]
notrab: @lightwalker-eth there are some places where the link styles are different.

Footer is gray with a regular `underline`:

CleanShot 2024-06-17 at 13 05 46@2x

Navbar of nameguard is black text with regular underline, no effect on hover:

CleanShot 2024-06-17 at 13 07 51@2x

Are these cases to be covered as well, or are they legacy?

packages/namekit-react/src/components/Link.tsx
```diff
@@ -27,30 +31,41 @@ const isInternalLink = (href: string | undefined) => {
   return href.startsWith("/") || href.startsWith("#") || !href.includes("://");
 };
 
-export const Link: React.FC<LinkProps> = ({
+export const ExternalIcon: React.FC = () => <span>↗</span>;
```
lightwalker-eth, line 34: Perhaps we should add some horizontal spacing before / after this depending on it's placement? Right now it feels too close to the anchor text.

apps/storybook/stories/Namekit/Link.stories.tsx
```diff
@@ -119,3 +119,31 @@ export const AsChild: Story = {
     asChild: <span />,
   },
 };
+
+export const ExternalLinkIconAfter: Story = {
+  args: {
+    variant: "primary",
+    size: "medium",
+    href: "#",
```
lightwalker-eth, line 127: Suggest we make the href for these truly external.

apps/storybook/stories/Namekit/Link.stories.tsx
```diff
@@ -119,3 +119,31 @@ export const AsChild: Story = {
...
+      </>
+    ),
+  },
+};
+
+export const ExternalLinkIconBefore: Story = {
+  args: {
+    variant: "primary",
+    size: "medium",
+    href: "#",
```
lightwalker-eth, line 141: Suggest we make the href for these truly external.

packages/namekit-react/src/components/Link.tsx
```diff
@@ -8,11 +8,15 @@ export interface LinkProps extends React.ComponentPropsWithoutRef<"a"> {
...
+interface LinkComponent extends React.FC<LinkProps> {
+  ExternalIcon: typeof ExternalIcon;
+}
+
 const linkBaseClasses = "nk-transition";
 
 const variantClasses = {
   primary:
-    "nk-text-black nk-underline sm:nk-underline-offset-[4px] sm:nk-transition-all sm:nk-duration-200 sm:hover:nk-underline-offset-[2px]",
+    "nk-text-current nk-underline sm:nk-underline-offset-[4px] sm:nk-transition-all sm:nk-duration-200 sm:hover:nk-underline-offset-[2px]",
```
lightwalker-eth, line 19: Could we add a story to the storybook showing an example where the background is dark and the text is light?

lightwalker-eth (commented):
@notrab Hey cool updates! I like the direction. Reviewed and shared some fresh feedback

packages/namekit-react/src/components/Link.tsx
notrab: @lightwalker-eth yeah, was going to address if this was the right direction 😄

packages/namekit-react/src/components/Link.tsx
notrab: The custom class name is the example for this 😅

notrab: I added an explicit example though.

apps/storybook/stories/Namekit/Link.stories.tsx
```diff
@@ -0,0 +1,121 @@
...
+  },
+};
+
+export const AsChild: Story = {
+  args: {
+    variant: "primary",
+    size: "medium",
+    href: "#",
+    children:
+      "Link as a Span (which won't work, but is good for next/link instances)",
```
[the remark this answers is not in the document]
notrab: It's fixed.

packages/namekit-react/src/components/Link.tsx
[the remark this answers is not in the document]
notrab: @lightwalker-eth I've added an `underline` variant that is always dark with dark underline. For cases of the footer and other cases of different colour underlines/text, we can pass the custom className option 👍🏻

packages/namekit-react/src/components/Link.tsx
notrab: I think this should be handled as a button or something else. No need to use the wrong semantic element.

PR #292, opened 2024-06-14
feat: add tailwind to storybook app

FrancoAguzzi:


notrab:
Hey @FrancoAguzzi 

Can you better help me understand the reason for adding Tailwind to storybook?

The components we render here should have their own styles from nameguard-react and namekit-react. I worry adding this right now will mask any style issues we have since they'll look normal even, because styles will fallback to Tailwind.

If there's another reason, please let me know. Thanks!

PR #294, opened 2024-06-15
feat: [sc-23853] Contact Page for NameGuard website

eduramme:
Story details: https://app.shortcut.com/ps-web3/story/23853


apps/nameguard.io/src/app/components/atoms/index.ts
```diff
@@ -1,3 +1,11 @@
...
 export * from "./CalButton";
 export * from "./ListSectionBadge";
+export * from "./PreSectionText";
+export * from "./SectionText";
+export * from "./ColorfulBg";
+export * from "./SectionTitle";
+export * from "./AirplaneIllustration";
+export * from "./";
+export * from "./";
+export * from "./";
```
notrab, line 11: Duplicate exports

lightwalker-eth (changes requested):
@eduramme Thanks for these updates :+1: Reviewed and shared feedback.

apps/nameguard.io/public/assets/completion_score.svg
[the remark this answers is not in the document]
eduramme: We manage SVGs in React components and HTML differently. For example, HTML SVGs use stroke-dasharray and React uses strokeDasharray in camelCase. I attempted to fix an SVG in a React component and ended up breaking the regular ones. My bad. Thanks for pointing it out :)

lightwalker-eth (changes requested):
@eduramme Thanks for your updates here :+1: Found a few more items in a fresh QA review. Shared feedback for you here: https://app.shortcut.com/ps-web3/story/23853/contact-page-for-nameguard-website#activity-25354

package.json
```diff
@@ -16,7 +16,7 @@
     "eslint": "^8.56.0",
     "prettier": "^3.2.5",
     "rimraf": "^5.0.7",
-    "turbo": "1.13.3"
+    "turbo": "2.0.5"
```
lightwalker-eth, line 19: This change applies to the whole monorepo and don't suggest we include it in this PR. Can we revert this change?

It would be great if you could create a Shortcut story for the goal of upgrading to turbo 2.x. Please note there's a whole guide of steps that we should follow when we implement this change: https://turbo.build/repo/docs/crafting-your-repository/upgrading

apps/nameguard.io/src/app/components/organisms/Header.tsx
```diff
@@ -61,9 +61,9 @@ export const Header = () => {
             >
               Contact
             </Link>
-            <div className="px-5 py-2 rounded-md bg-black text-white hover:bg-gray-800 transition-all duration-300 whitespace-nowrap">
-              <CalButton>Discuss an integration</CalButton>
-            </div>
+            <CalButton className="px-4 py-2 text-sm rounded-md bg-black text-white hover:bg-gray-800 transition-all duration-300 whitespace-nowrap">
+              Discuss an integration
```
lightwalker-eth, line 65: Can we make this text appear with the same weight as the text of the other buttons / links right next to it? The other  buttons / links right next to it appear to have bold font, while the text in this button doesn't. All these buttons / links should appear bold.

lightwalker-eth (changes requested):
@eduramme Thanks for updates. Shared feedback :+1:

package.json
eduramme: Seems like we're already using turbo 2.0.5 on main 🤔

lightwalker-eth: @eduramme Ah if this change is already on main, then I didn't mean to create confusion with my feedback

lightwalker-eth: @eduramme ... but I still understand we need to follow that upgrade process properly that I linked above and I'm not sure we've followed that yet. Therefore please let me know how you suggest we proceed. You're welcome to create a new shortcut story for the turbo upgrade as one option for how we proceed.

lightwalker-eth (approved):
@eduramme Thanks for the updates. Reviewed and approved :+1: Please take the next steps here to get this merged into main.

PR #299, opened 2024-07-02
fix(input): use correct styling and stories

notrab:
This doesn't include a changeset since the original changeset hasn't yet been merged.


lightwalker-eth (approved):
@notrab Updates look great. Thank you :+1:

PR #300, opened 2024-07-02
fix(textarea): use correct styling and stories

notrab:
No changeset, previous one wasn't merged yet.


packages/namekit-react/src/components/ENSTextArea.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import cc from "classcat";
+
+import { TextArea, TextAreaProps } from "./TextArea";
+
+export const ENSTextarea: React.FC<TextAreaProps> = (props) => {
```
lightwalker-eth, line 6: @notrab suggest we also rename this to `ENSTextArea` with an uppercase `A` and also update related items in Storybook, ex:

CleanShot 2024-07-03 at 10 13 37

lightwalker-eth (changes requested):
@notrab Thanks for updates. All looks good except for a capitalization suggestion.

packages/namekit-react/src/components/ENSTextArea.tsx
notrab: Good catch, fixed

lightwalker-eth (approved):
@notrab Thanks, looks good :+1:

PR #301, opened 2024-07-08
Feat/nft avatar migration

FrancoAguzzi:


apps/storybook/stories/Namekit/RegistrationPriceInfo.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+  buildTimestamp,
+  now,
+} from "@namehash/ens-utils";
+import type { Meta, StoryObj } from "@storybook/react";
+import {
+  RegistrationPriceInfo,
+  RegistrationPriceInfoDisplayingFormat,
+} from "@namehash/namekit-react";
+
+const ELEVEN_DAYS_FROM_NOW_TIMESTAMP = buildTimestamp(
```
lightwalker-eth, line 13: I remember leaving a comment about this in another PR. Can you please look for where those comments were and follow up on all of those again in this PR?

I'm worried that some of the detailed comments and feedback I've shared somewhere else is getting lost.

lightwalker-eth (changes requested):
@FrancoAguzzi Wow this is a really big PR with lots of progress. Excited to see this.

There's a few key follow ups that need your help:
1. I wasn't able to test any of this because preview deployments are failing. Please investigate and resolve.
2. There's several places where I saw an issue that I remember previously sharing a code review feedback on, but now that feedback is gone from this PR 301. It seems there's some other earlier PR where I had open feedback for you but that isn't reflected in PR 301 yet? Can you please investigate? I'm worried that I saw some important issue in a different PR and added comments for it there but those ideas are not being reflected in PR 301.
3. There's a few files inside nameguard-react that I think should be removed as they now  should exclusively live inside namekit-react. Right?
4. Added other comments and feedback for you :+1:

packages/nameguard-react/src/components/TruncatedText/TruncatedText.tsx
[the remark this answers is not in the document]
FrancoAguzzi: I've added documentation to come back to this removal once things are stable, after the new comp. is merged into main, as agreed :)

packages/namekit-react/src/components/EnsAvatar.tsx
```diff
@@ -60,7 +60,9 @@ export const EnsAvatar = ({ name }: AvatarProps) => {
 
   const EnsAvatarLayout =
     name && name.normalization !== Normalization.normalized ? (
-      <NotNormalizedNameEnsAvatar normalization={name.normalization} />
+      <NotNormalizedNameEnsAvatar
+        normalization={name.normalization as Normalization}
```
lightwalker-eth, line 64: Hmm. Why wouldn't the type system already know that `name.normalization` is type `Normalization`?

lightwalker-eth (commented):

PR #303, opened 2024-07-11
rename storybook

notrab:


lightwalker-eth (commented):
@notrab  renaming is good. Appreciate if we can fix the builds :+1:

lightwalker-eth (approved):
@notrab Thanks for update :+1: Looks good! Approved!

PR #305, opened 2024-07-15
rename repo path

notrab:
@BanaSeba @djstrong we want to rename the `nameguard/nameguard` repo to `nameguard/namekit`.


lightwalker-eth (approved):
@notrab Let's go for it :+1:

PR #306, opened 2024-07-18
feat: use cross-fetch

notrab:


lightwalker-eth (changes requested):
@notrab Looks good, thank you! Could you please add a changeset for this PR that includes about the changes in this PR? Ex: fetch library and tsconfig updates

PR #311, opened 2024-07-19
feat: [sc-25343] Nameguard - Update Metadata URL to Use Environment Variable

eduramme:
Story details: https://app.shortcut.com/ps-web3/story/25343


apps/nameguard.io/src/app/layout.tsx
```diff
@@ -18,9 +18,13 @@ const title = "Protect your community with NameGuard for ENS";
 const description =
   "Guard your users from heartbreak and encourage best practice usage of ENS.";
 
+const productionUrl =
```
lightwalker-eth, line 21: This doesn't work the way you've assumed it would. If this environment variable isn't defined it will result in the following value for `productionUrl`: "https://undefined".

Please make sure to test logic like this. Thanks

lightwalker-eth (dismissed):
@eduramme Found an issue. Looking forward to your fix. Thanks.

notrab (approved):

notrab:
@eduramme we can add back the hardcoded values if that's what we need to make this work. I really don't understand Next.js these days and its "config over configuration" when it comes to `metadataBase`. Some examples I use the production URL provided by Vercel, and in staging PRs, nobody ever has this value set correctly.

PR #312, opened 2024-07-20
feat: Domain Card Ownership Badge

FrancoAguzzi:


packages/namekit-react/src/components/DomainOwnershipBadge.tsx
```diff
@@ -0,0 +1,62 @@
...
+  getUserOwnership,
+  UserOwnershipOfDomain,
+} from "@namehash/ens-utils";
+import React from "react";
+import cc from "classcat";
+
+interface DomainOwnershipBadgeProps {
+  domain: DomainCard | null; // when null we show the skeleton
+  onClickHandler?: () => void;
+  addressToCompareOwnership: Address | null;
```
lightwalker-eth, line 14: Can you please suggest some more clear way to name and/or document this field? It's not clear for me how that field works just by reading the name.

apps/storybook.namekit.io/stories/Namekit/DomainOwnershipBadge.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+const meta: Meta<typeof DomainOwnershipBadge> = {
+  component: DomainOwnershipBadge,
+  title: "Namekit/DomainOwnershipBadge",
+  parameters: {
+    layout: "centered",
+  },
+  args: {
+    onClickHandler: undefined,
+    domain: getMockedDomainCard({
+      domainStatus: DomainStatus.Normal,
```
lightwalker-eth, line 20: What is "Normal"? What does this mean in the context of ENS?

lightwalker-eth (changes requested):
@FrancoAguzzi Thanks for this. Reviewed and shared feedback :+1:

packages/namekit-react/src/components/DomainOwnershipBadge.tsx
```diff
@@ -11,6 +11,12 @@ import cc from "classcat";
 interface DomainOwnershipBadgeProps {
   domain: DomainCard | null; // when null we show the skeleton
   onClickHandler?: () => void;
+
+  /*
+    The address to compare ownership represents the address
```
lightwalker-eth, line 16: @FrancoAguzzi so then just rename this field to something intuitive such as `currentUserAddress`.

lightwalker-eth (changes requested):

PR #313, opened 2024-07-20
feat: Domain Card Menu

FrancoAguzzi:


packages/namekit-react/src/components/DomainMenu.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  const textSize = isMobile ? "nk-text-md" : "nk-text-sm";
+  const menuItems = (
+    <>
+      {!!domainCard.ownerAddress && (
+        <MenuItem>
+          <Link
+            target="_blank"
+            variant="secondary"
+            href={`https://app.ens.domains/${domainCard.ownerAddress}`}
```
lightwalker-eth, line 71: We're open sourcing these components. It's not proper to hardcode things like this. It needs to be something that can be configured dynamically.

lightwalker-eth (changes requested):
@FrancoAguzzi Thanks. Reviewed and shared feedback.

PR #318, opened 2024-07-26
feat: delete Normalization enum from nameguard-sdk

FrancoAguzzi:


lightwalker-eth (approved):

lightwalker-eth (changes requested):
@FrancoAguzzi Also, we're missing the changeset for this PR

packages/nameguard-sdk/package.json
```diff
@@ -39,6 +39,7 @@
   },
   "devDependencies": {
     "@types/node": "^20.6.2",
+    "@namehash/ens-utils": "workspace:*",
```
lightwalker-eth, line 42: Hmm.. I don't think it's right that this is listed only under `devDependencies`?

lightwalker-eth: Shouldn't it be under `dependencies` instead? Please confirm everything is configured correctly for us here and in the lockfile.

lightwalker-eth (approved):

PR #319, opened 2024-07-26
Skip NameGuard API CICD unless it changes

lightwalker-eth:


.github/workflows/nameguard_api_cicd.yml
```diff
@@ -3,8 +3,10 @@ name: Deploy NameGuard to AWS as a serverless lambda
 on:
   push:
     branches:
-      - ci-cd-lambda-nameguard-v2 #temporary for testing
       - main
+    paths:
+      - 'api/**'
+      - '.github/workflows/nameguard_api_cicd.yml'
```
notrab, line 9: Will this action not be evaluated regardless? So this line specifically can be removed?

notrab (approved):

.github/workflows/nameguard_api_cicd.yml
lightwalker-eth: @notrab Wasn't sure how that would be handled, so added this line just in case. Think we should remove it?

lightwalker-eth: @notrab I think it makes sense to include this line because we might have a PR where we don't change any of the code but update the cicd instead. That should run the new cicd.

notrab: It will run regardless iirc.

PR #321, opened 2024-07-28
Cleanup workflows

notrab:
This PR updates the workflows to be:

- `api-python-release.yml` which is used to publish the python package on changes to `api`
- `api-tests.yml` is used to run tests when the api changes
- `changesets.yml` is used to publish the NPM packages on changes to `packages`
- `deploy-api.yml` is used to deploy the API package when `api` is merged
- `lint.yml` is used to lint the project. I'm not sure exactly on this tool, and didn't originally add it, so need to investigate more. 
- `package-tests.yml` is used to run `pnpm test` on the package in the matrix.


.github/workflows/api-tests.yml
```diff
@@ -1,32 +1,29 @@
-name: Tests
-
+name: API Tests
```
notrab, line 1:
```suggestion
name: NameGuard API Tests
```

I was going to name this `NameGuard API Tests`, but since this will only run when the NameGuard API folder is changed, I didn't really see the point. 

Feel free to commit this suggestion if we want it to be more explicit.

djstrong: For me "API" is confusing. It is now referring to `api` directory which is the Python package.

notrab: @djstrong please suggest any changes. I've never looked at the files in those directories too close to know how things are structured.

djstrong: Maybe just use the same string in filenames. Now we have `api` and `api-python`.
Also under namekit's `api` directory is NameGuard (Python). Maybe it should be under `api/nameguard`? If we plan to move more Python packages under `api` directory then it is probably a must. I guess Python packages can not be tested in one workflow then the names could be: 
* `api-python-release.yml` -> 'api-nameguard-release.yml'
* `api-tests.yml` -> `api-nameguard-tests.yml`
* `deploy-api.yml` -> `deploy-api-nameguard.yml`

djstrong:
@notrab ci_sdk can't be removed

.github/workflows/nameguard-end-to-end-tests.yml
```diff
@@ -0,0 +1,91 @@
...
+
+      - name: Local api healthcheck
+        if: needs.check_changes.outputs.api_changed > 0
+        run: curl http://0.0.0.0:8000/v1-beta/inspect-name/mainnet/test.eth
+
+      - name: Run tests
+        env:
+          PROVIDER_URI_MAINNET: ${{ secrets.PROVIDER_URI_MAINNET }}
+          PROVIDER_URI_SEPOLIA: ${{ secrets.PROVIDER_URI_SEPOLIA }}
+          NAMEGUARD_TEST_ENDPOINT: ${{ needs.check_changes.outputs.api_changed > 0 && 'http://0.0.0.0:8000' || '' }}
```
djstrong, line 90: @lightwalker-eth @notrab Here, if API doesn't change then we test SDK against production endpoint - is it ok or should test against local endpoint?
Not running local endpoint might be faster.
Production and local endpoints might give different results because production endpoint is not automatically deployed - so I think it should be tested against local version.

djstrong: However, nameguard-sdk is automatically released (right?) so there might be a mismatch.

notrab: `nameguard-sdk` is released on merge to main, so tests would have to have been approved prior to that, am I right?

djstrong: So if we change SDK and API then we are testing SDK against local API.
So test against local API pass, and test against prod API fail. If we skip test against prod API, then released SDK will not work (against prod API).

notrab: Ok, we can revert

djstrong: There is no solution for this.

notrab: imo, we can always test local running api + sdk when api and/or sdk changes.

We have to assume what ends up in prod is good, and passing tests.

We could alternatively, extend the deploy api workflow to run the SDK tests there again to be safe, but it sounds like overkill.

BanaSeba (approved):

lightwalker-eth (approved):
Looks good 🚀

PR #338, opened 2024-07-30
feat: [sc-25261] Next Steps on DisplayedName and TruncatedText

FrancoAguzzi:
Story details: https://app.shortcut.com/ps-web3/story/25261


lightwalker-eth (changes requested):
@FrancoAguzzi It's nice to see that this is a smaller PR, but there's some issues with it:

1. Am I correct to understand that this would result in these UI components being duplicated across both packages? That's fine, but our mindset should be that we should always have really quality and well documented code on our main branch. The way this PR is currently it leaves the main branch in a confusing state. The fix I would propose to this is to include a comment at the top of each of the duplicated files (in both packages) explaining what is going on and instructions for anyone looking at that code to understand what's happening.
2. I don't consider this PR as being properly focused. Focus is not only about reducing the breadth of changes, it is also about increasing the depth of changes. This means that for any change we introduce in a PR, we should really see through that change deeply. If a PR is adding a UI component to NameKit, why isn't that PR also adding that UI component to Storybook? Etc.. At all times there should be a beautiful alignment between the UI components in our packages and the storybook that documents them and helps the world understand them.

PR #340, opened 2024-07-30
feat: upgrade ens-utils and namekit-react tsconfig.json

FrancoAguzzi:
These `tsconfig.json` updates are necessary for the upcoming events on these libraries and its dependents.

For reference: 
- https://www.typescriptlang.org/tsconfig/#rootDir
- https://www.typescriptlang.org/tsconfig/#outDir
- https://www.typescriptlang.org/tsconfig/#moduleResolution
- https://www.typescriptlang.org/tsconfig/#module
- https://www.typescriptlang.org/tsconfig/#target

### DIFF
--- a/.changeset/two-rockets-sneeze.md
+++ b/.changeset/two-rockets-sneeze.md
@@ -1,0 +1,6 @@
+---
+"@namehash/namekit-react": minor
+"@namehash/ens-utils": minor
+---
+
+Update tsconfig.json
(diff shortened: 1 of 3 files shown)


lightwalker-eth (changes requested):
@FrancoAguzzi Thank you for the updates. We'll need changesets on this PR though before we can merge it.

lightwalker-eth (approved):
@FrancoAguzzi Thank you :+1:

PR #343, opened 2024-08-05
Refine domain logic

lightwalker-eth:


packages/ens-utils/src/ethregistrar.ts
```diff
@@ -3,323 +3,351 @@ import {
...
       }
     }
 
-    // TODO: refine this. For example, there's a maximum length limit, etc.
-    return true;
-  }
-}
+    if (duration.seconds < 1n) {
+      // TODO: enforce that `duration` is the minimum duration allowed for a registration
+      // TODO: need to put the right constant here.
```
FrancoAguzzi, line 110: Suggest to use the old `DEFAULT_REGISTRATION_YEARS` multiplied by `SECONDS_PER_DAY`

packages/ens-utils/src/ethregistrar.ts
```diff
@@ -3,323 +3,351 @@ import {
...
-// known registrars
-export const WRAPPED_MAINNET_ETH_REGISTRAR_CONTRACT = buildContractRef(
-  MAINNET,
-  buildAddress("0xD4416b13d2b3a9aBae7AcD5D6C2BbDBE25686401"),
-);
-export const UNWRAPPED_MAINNET_ETH_REGISTRAR_CONTRACT = buildContractRef(
-  MAINNET,
-  buildAddress("0x57f1887a8BF19b14fC0dF6Fd9B2acc9Af147eA85"),
-);
+  public getManagedSubname = (name: ENSName): ENSName | null => {
```
FrancoAguzzi, line 54: Should this function be called `getDirectSubname`?

packages/ens-utils/src/registrar.ts
```diff
@@ -0,0 +1,217 @@
...
+  price: Price;
+  reason?: undefined;
+  validity?: undefined;
+}
+
+export interface RegistrarSpecialNameFee extends AbstractRegistrarCharge {
+  type: typeof RegistrarChargeType.SpecialNameFee;
+  price: Price;
+  reason: string;
+  validity?: undefined;
```
FrancoAguzzi, line 110:
```suggestion
  validity: undefined;
```

packages/ens-utils/src/registrar.ts
```diff
@@ -0,0 +1,217 @@
...
+  price: Price;
+  reason: string;
+  validity: TimePeriod;
+}
+
+export interface RegistrarServiceFee extends AbstractRegistrarCharge {
+  type: typeof RegistrarChargeType.ServiceFee;
+  price: Price;
+  reason?: string;
+  validity?: undefined;
```
FrancoAguzzi, line 124:
```suggestion
  validity: undefined;
```

packages/ens-utils/src/registrar.ts
```diff
@@ -0,0 +1,217 @@
...
+   * If `null`, the `RegistrarCharge` is not guaranteed to be time-limited.
+   */
+  validity?: TimePeriod;
+}
+
+export interface RegistrarBaseFee extends AbstractRegistrarCharge {
+  type: typeof RegistrarChargeType.BaseFee;
+  price: Price;
+  reason?: undefined;
+  validity?: undefined;
```
FrancoAguzzi, line 103:
```suggestion
  validity: undefined;
```

packages/ens-utils/src/ethregistrar.ts
```diff
@@ -3,323 +3,351 @@ import {
...
-): Registrar {
-  const registrars = getPotentialKnownRegistrars(name, chain, isWrapped);
-  if (registrars.length > 1) {
-    throw new Error(
-      `Multiple potential registrars found for name: "${name.name}" on chainId: ${chain.chainId} when isWrapped: ${isWrapped}`,
+    if (!existingRegistration.expirationTimestamp) {
+      throw new Error(`Invariant violation`); // TODO: refine message... just making the type system happy.
+    }
+
+    // TODO: review for possible off-by-1 errors
```
FrancoAguzzi, line 214: @lightwalker-eth can you extend on off-by-1 spec?

packages/ens-utils/src/ethregistrar.ts
```diff
@@ -3,323 +3,351 @@ import {
...
 
-    // last label must be "eth"
-    if (name.labels[1] !== ETH_TLD) return false;
+      const releaseTimestamp = getDomainReleaseTimestamp(existingRegistration);
 
-    // first label must be of sufficient length
-    return charCount(name.labels[0]) >= MIN_ETH_REGISTRABLE_LABEL_LENGTH;
+      if (releaseTimestamp && releaseTimestamp.time > atTimestamp.time) {
+        // if the name is released, we can't renew it anymore
+        // TODO: check for possible off-by-1 errors in the logic above
```
FrancoAguzzi, line 185: @lightwalker-eth can you extend on off-by-1 spec?

packages/ens-utils/src/ethregistrar.ts
```diff
@@ -3,323 +3,351 @@ import {
...
 
-    // must have exactly 2 labels to be a direct subname of ".eth"
-    if (name.labels.length !== 2) return false;
+    if (existingRegistration) {
+      if (
+        existingRegistration.registrationTimestamp &&
+        atTimestamp.time < existingRegistration.registrationTimestamp.time
+      ) {
+        // if the renewal is requested before the registration, we can't renew it
+        // TODO: check for possible off-by-1 errors in the logic above
```
FrancoAguzzi, line 177: @lightwalker-eth can you extend on off-by-1 spec?

packages/ens-utils/src/registrar.ts
```diff
@@ -0,0 +1,217 @@
...
+export const scaleAnnualPrice = (annualPrice: Price, duration: Duration) => {
+  // Small performance optimization if no scaling is needed
+  if (duration.seconds === SECONDS_PER_YEAR.seconds)
+    return annualPrice;
+
+  // TODO: verify we're doing this division of bigints correctly
+  const scaledRate = Number(duration.seconds) / Number(SECONDS_PER_YEAR.seconds);
+
+  return approxScalePrice(annualPrice, scaledRate);
+}
```
FrancoAguzzi, line 217: I'd rather do it this way:

```suggestion
export const scaleAnnualPrice = (annualPrice: Price, duration: Duration) => {
  // Small performance optimization if no scaling is needed
  if (duration.seconds === SECONDS_PER_YEAR.seconds)
    return annualPrice;

  const scaledRate = (duration.seconds) / (SECONDS_PER_YEAR.seconds);

  return approxScalePrice(annualPrice, Number(scaledRate));
}
```

Since a rate shall be smaller than a seconds value, making the seconds representation as bigint and the rate as number.

BUT

As divisions by bigint will never result in floating point numbers, for precision sake and with the understandment that, hardly some seconds value will be greater than `Number.MAX_SAFE_INTEGER` (9007199254740991 which represents date that is later than 259000)

FrancoAguzzi: Considering above information, you are doing the division correctly and safely

packages/ens-utils/src/registrar.ts
```diff
@@ -0,0 +1,217 @@
...
+export interface OnchainRegistrar extends Registrar {
+  getOnchainRegistrar(): ContractRef;
+}
+
+export interface OffchainRegistrar extends Registrar {
+  getOnchainRegistrar(): null;
+}
+
+/**
+ * Identifies that a name was passed to a `Registrar` function that was not relevant for that `Registrar`.
```
FrancoAguzzi, line 194: @lightwalker-eth I am not sure "relevant" is the best word to be used in here.

Relevant shall mean something related to quality, to some rating level, while it seams to me that this error and its applications are related to Registrar support, name validity, registration process effectiveness understanding.

packages/ens-utils/src/registrar.ts
```diff
@@ -0,0 +1,217 @@
...
+}
+
+export interface RenewalPriceQuote extends AbstractRegistrarPriceQuote {
+  action: typeof RegistrarAction.Renew;
+}
+
+export interface Registrar {
+  getManagedSubname(name: ENSName): ENSName | null;
+
+  // NOTE: Throws RegistrarUnsupportedNameError if name is not supported by this registrar
```
FrancoAguzzi, line 173: I am not sure I would leave this note here since this is already defined inside `getValidatedSubname` function

packages/ens-utils/src/registrar.ts
```diff
@@ -0,0 +1,217 @@
...
+  /**
+   * The `TimePeriod` that this `RegistrarPriceQuote` is for.
+   */
+  rentalPeriod: TimePeriod;
+
+  /**
+   * The set of distinct `RegistrarCharge` values that must be paid to the
+   * `Registrar` to rent the domain for the period of `rentalPeriod`.
+   * 
+   * May be empty if the domain is free to rent for `rentalPeriod`.
```
FrancoAguzzi, line 149: Is this case possible?

FrancoAguzzi: In reality

packages/ens-utils/src/registrar.ts
```diff
@@ -0,0 +1,217 @@
...
+ * Identifies that a name was passed to a `Registrar` function that was not relevant for that `Registrar`.
+ * 
+ * @param message a reason why `name` was NOT relevant for `Registrar`.
+ * @param label the `ENSName` value that was NOT relevant.
+ */
+export class RegistrarUnsupportedNameError extends Error {
+
+  public constructor(message: string, name: ENSName) {
+    super(`RegistrarUnsupportedNameError for name "${name.name}": ${message}`);
+    this.name = "RegistrarUnsupportedNameError";
```
FrancoAguzzi, line 203: Might be good to exchange `name` convention so we don't have a errorName as string and a ensName as ENSName?

packages/ens-utils/src/registrar.ts
```diff
@@ -0,0 +1,217 @@
...
+  
+  canRenew(name: ENSName, atTimestamp: Timestamp, duration: Duration, existingRegistration: Registration): boolean;
+  getRenewalPriceQuote(name: ENSName, atTimestamp: Timestamp, duration: Duration, existingRegistration: Registration): RenewalPriceQuote;
+}
+
+export interface OnchainRegistrar extends Registrar {
+  getOnchainRegistrar(): ContractRef;
+}
+
+export interface OffchainRegistrar extends Registrar {
```
FrancoAguzzi, line 189: Might be good to document this is a future-support-want feature that is not ready to be used?

FrancoAguzzi (changes requested):

packages/ens-utils/src/ethregistrar.ts
```diff
@@ -3,323 +3,351 @@ import {
...
+      const releaseTimestamp = getDomainReleaseTimestamp(existingRegistration);
 
-    if (name.labels.length >= 2) {
-      if (name.labels[1] === ETH_TLD) {
-        // first label must be of sufficient length
-        if (charCount(name.labels[0]) < MIN_ETH_REGISTRABLE_LABEL_LENGTH)
-          return false;
+      if (releaseTimestamp && releaseTimestamp.time > atTimestamp.time) {
+        // if the name is not yet released, we can't register it
+        // TODO: check for possible off-by-1 errors in the logic above
```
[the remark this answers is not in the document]
lightwalker-eth: We need to see how ENS smart contracts actually work. Ex: Actual expiration time? 1 second earlier? 1 second later?

packages/ens-utils/src/ethregistrar.ts
lightwalker-eth: No it's not a good interface to use years here. It should be defined in seconds as thats how ENS smart contracts work.

The goal described above is totally different though. Please see ENS contracts there's a minimum registration time. You can't register for 1 second.

packages/ens-utils/src/ethregistrar.ts
lightwalker-eth: No. Sometimes the managed subname isn't direct.

packages/ens-utils/src/registrar.ts
lightwalker-eth: No, that removes the benefit of the approach I wrote has with the type system.

packages/ens-utils/src/registrar.ts
lightwalker-eth: No, that removes the benefit of the approach I wrote has with the type system.

packages/ens-utils/src/registrar.ts
lightwalker-eth: No, that removes the benefit of the approach I wrote has with the type system.

packages/ens-utils/src/ethregistrar.ts
lightwalker-eth: We need to see how ENS smart contracts actually work. Ex: Actual expiration time? 1 second earlier? 1 second later?

packages/ens-utils/src/ethregistrar.ts
lightwalker-eth: We need to see how ENS smart contracts actually work. Ex: Actual expiration time? 1 second earlier? 1 second later?

packages/ens-utils/src/ethregistrar.ts
lightwalker-eth: We need to see how ENS smart contracts actually work. Ex: Actual expiration time? 1 second earlier? 1 second later?

packages/ens-utils/src/registrar.ts
lightwalker-eth: Thanks do you still have a suggestion that we change anything here?

packages/ens-utils/src/registrar.ts
lightwalker-eth: You're welcome to rename "relevant" to something more specific, such as "managed".

The purpose of this is to communicate the case that an `ENSName` is passed to a `Registrar` that the `Registrar` doesn't manage.

packages/ens-utils/src/registrar.ts
lightwalker-eth: This idea needs to be documented as part of the interface.

packages/ens-utils/src/registrar.ts
lightwalker-eth: Sure. For example: cb.id or uni.eth subnames.

We want to be building towards support for these now and not needing to fundamentally change everything in our code once we expand support officially.

packages/ens-utils/src/registrar.ts
lightwalker-eth: I don't understand?

packages/ens-utils/src/registrar.ts
lightwalker-eth: Sure you're welcome to do that.

packages/ens-utils/src/registrar.ts
lightwalker-eth: For example, if "frankind.cb.id" was passed to the .eth registrar asking for the registration price, the .eth registrar needs to have a well defined way of saying I can't answer that question because I don't manage that name.

packages/ens-utils/src/registrar.ts
lightwalker-eth: Appreciate if you could please create unit tests for this function. That's the best way to be certain.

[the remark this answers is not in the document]
lightwalker-eth: You're welcome to add these :+1:

packages/ens-utils/src/ensname.ts
```diff
@@ -358,3 +343,68 @@ export function getRegistrationPotential(name: ENSName): RegistrationPotential {
...
+
+  const split = inputName.split(".");
+  const labels = [split[0].slice(1, -1), keccak(split[1])];
+
+  for (let i = labels.length - 1; i >= 0; i--) {
+    const labelSha = labels[i];
+    node = keccak(Buffer.from(node + labelSha, "hex"));
+  }
+  return "0x" + node;
+};
```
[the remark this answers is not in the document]
lightwalker-eth: You're welcome to add these :+1:

packages/ens-utils/src/domain.ts
```diff
@@ -25,127 +24,71 @@ export interface DomainCard {
...
-  const split = inputName.split(".");
-  const labels = [split[0].slice(1, -1), keccak(split[1])];
-
-  for (let i = labels.length - 1; i >= 0; i--) {
-    const labelSha = labels[i];
-    node = keccak(Buffer.from(node + labelSha, "hex"));
-  }
-  return "0x" + node;
-};
+};
```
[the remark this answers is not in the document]
lightwalker-eth: You're welcome to add these :+1:

packages/ens-utils/src/registrar.ts
FrancoAguzzi: Both `constructor` `name` param *and* `RegistrarUnsupportedNameError` class `name` prop are defined as `name`... in the first case, a class constructor param is defined as `ENSName`, while, in the second one, the class public variable is defined as `string`.

My suggestion is to rename the class constructor `name` to `ensName` to avoid these duplicates

packages/ens-utils/src/registrar.ts
FrancoAguzzi: Nicely thought!

packages/ens-utils/src/ethregistrar.ts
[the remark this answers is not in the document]
FrancoAguzzi: Captura de Tela 2024-08-06 às 20 16 40

This function was extracted from BaseRegistrar and reveals that above logic could look different: added a commit suggestion above!

packages/ens-utils/src/namekitregistrar.ts
```diff
@@ -0,0 +1,51 @@
...
+// You're not supposed to use this yourself directly. You should deploy your own instance of
+// this contract.
+export const MAINNET_EXAMPLE_NAMEKIT_REGISTRAR_CONTRACT = buildContractRef(
+  MAINNET,
+  buildAddress("0x232332263e6e4bd8a134b238975e2200c8b7dac1"),
+);
+
+export const MAINNET_EXAMPLE_NAMEKIT_REGISTRAR = new NameKitRegistrar(
+    MAINNET_EXAMPLE_NAMEKIT_REGISTRAR_CONTRACT
+);
```
FrancoAguzzi, line 51:
```suggestion
);

```

packages/ens-utils/src/ethregistrar.ts
```diff
@@ -3,323 +3,351 @@ import {
...
+        return false;
+      }
 
-    // last label must be "eth"
-    if (name.labels[1] !== ETH_TLD) return false;
+      const releaseTimestamp = getDomainReleaseTimestamp(existingRegistration);
 
-    // first label must be of sufficient length
-    return charCount(name.labels[0]) >= MIN_ETH_REGISTRABLE_LABEL_LENGTH;
+      if (releaseTimestamp && releaseTimestamp.time > atTimestamp.time) {
```
FrancoAguzzi, line 183:
```suggestion
      if (releaseTimestamp && releaseTimestamp.time >= atTimestamp.time) {
```

packages/ens-utils/src/ethregistrar.ts
FrancoAguzzi: Captura de Tela 2024-08-06 às 20 16 40

This function was extracted from BaseRegistrar and reveals that above logic could look different: added a commit suggestion above!

packages/ens-utils/src/ethregistrar.ts
FrancoAguzzi: OK...

I'd say there is no problem with current logic.

If `atTimestamp` is the same than `registrationTimestamp`, it is already possible to renew it for longer periods, meaning `canRenew` should go on in the function scope, but, if it is smaller, it really is not possible to renew this domain: as per my review this has achieved its goal.

packages/ens-utils/src/ethregistrar.ts
FrancoAguzzi: Our logic seems to implement the same as below:

Captura de Tela 2024-08-06 às 20 29 09

When summing `expiries[id] + duration` to determine the new timestamp.

FrancoAguzzi (approved):

packages/ens-utils/src/ethregistrar.ts
FrancoAguzzi: I see, they validate duration is bigger than 0 with below requirement:

Captura de Tela 2024-08-06 às 20 38 09

FrancoAguzzi: Added one suggestion above

FrancoAguzzi, line 110:
```suggestion
    if (addSeconds(atTimestamp, duration) === atTimestamp) {
      // duration is at least 1 second
```

lightwalker-eth: @FrancoAguzzi I don't see any benefit to this suggestion? And this is not the correct minimum value.

packages/ens-utils/src/registrar.ts
```diff
@@ -175,29 +191,50 @@ export interface Registrar {
...
 export interface OffchainRegistrar extends Registrar {
   getOnchainRegistrar(): null;
 }
 
 /**
- * Identifies that a name was passed to a `Registrar` function that was not relevant for that `Registrar`.
- * 
- * @param message a reason why `name` was NOT relevant for `Registrar`.
- * @param label the `ENSName` value that was NOT relevant.
+ * Identifies that a name was passed to a `Registrar` function that is not managed by that `Registrar`.
```
lightwalker-eth, line 232: @FrancoAguzzi I've updated the terminology "managed by" to "issued by". In the context of a `Registrar` we should generally use the terminology "issued by".

I then further refined this from "issued by" to "issuable" because "issued by" suggests that the subname has already been issued, but what matters is that the `Registrar` has the capability to issue.

packages/ens-utils/src/registrar.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+describe("scaleAnnualPrice() function", (t) => {
+  it("should scale a one year price to a half an year price", () => {
+    const price = buildPrice(100n, Currency.Usd);
+    const years = 0.5;
+    const secondsPerYearAsNumber = Number(SECONDS_PER_YEAR.seconds);
+    const duration = BigInt(secondsPerYearAsNumber * years);
+    const secondsPerYears = buildDuration(duration);
+
+    const expectedResult = {
```
lightwalker-eth, line 15: @FrancoAguzzi Please avoid manually building prices. use `buildPrice`.

Additionally, in a context like this I think it's best to explicitly define the expected result rather than to dynamically compute it.

packages/ens-utils/src/registrar.test.ts
```diff
@@ -0,0 +1,77 @@
...
+import { scaleAnnualPrice } from "./registrar";
+import { buildPrice } from "./price";
+import { Currency } from "./currency";
+import { buildDuration, SECONDS_PER_YEAR } from "./time";
+
+describe("scaleAnnualPrice() function", (t) => {
+  it("should scale a one year price to a half an year price", () => {
+    const price = buildPrice(100n, Currency.Usd);
+    const years = 0.5;
+    const secondsPerYearAsNumber = Number(SECONDS_PER_YEAR.seconds);
```
lightwalker-eth, line 11: @FrancoAguzzi There's a much easier way to achieve this goal. See `scaleDuration`.

packages/ens-utils/src/registrar.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    expect(result).toEqual(expectedResult);
+  });
+  it("should not scale the price if the duration is the same as current price duration", () => {
+    const price = buildPrice(100n, Currency.Usd);
+    const years = 1n;
+    const secondsPerYears = buildDuration(SECONDS_PER_YEAR.seconds * years);
+
+    const result = scaleAnnualPrice(price, secondsPerYears);
+
+    expect(result).toEqual(result);
```
lightwalker-eth, line 75: @FrancoAguzzi You're not actually testing what you think you are here.

lightwalker-eth:
```
expect(result).toEqual(result);
```

Doesn't test anything.

packages/ens-utils/src/ethregistrar.ts
FrancoAguzzi: Your review is perfect, we can keep this as it is!

One loose end of this function, I see, though, is that if user provides no `Registration` + has a valid duration + provides an `atTimestamp` at the past. Can he register? Actually not, but function would return true.

Do you see it as well?

packages/ens-utils/src/registrar.ts
```diff
@@ -167,6 +171,18 @@ export interface RenewalPriceQuote extends AbstractRegistrarPriceQuote {
   action: typeof RegistrarAction.Renew;
 }
 
+/*
+  Top-Level Domains (TLDs), like .eth and .test, are owned by smart contracts 
+  called registrars, which specify rules governing the allocation of their names. 
+  Enabling seamless interoperability with the DNS (Domain Name System).
+
+  The source of truth for a name and its subdomains does not always have to be on-chain
```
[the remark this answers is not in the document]
FrancoAguzzi: @lightwalker-eth why do `OffchainRegistrar`s do not have an `OnchainRegistry`? 

I thought the registry was the root from where all `Registrar`s connected to 🧐

packages/ens-utils/src/registrar.ts
FrancoAguzzi: Exact 🎯

packages/ens-utils/src/ethregistrar.ts
```diff
@@ -3,323 +3,370 @@ import {
...
+import { buildAddress } from "./address";
 
-export interface Registrar {
-  contract: ContractRef;
-
-  getTokenId(name: ENSName, isWrapped: boolean): TokenId;
-  isClaimable(name: ENSName, isWrapped: boolean): boolean;
-}
+/**
+ * The `EthRegistrar` models the policy implmentations shared by both of
```
FrancoAguzzi, line 49:
```suggestion
 * The `EthRegistrar` models the policy implementations shared by both of
```

FrancoAguzzi: @lightwalker-eth If I can create my own `EthRegistrar` that registers `.eth`, what are the conditions to do so in a valid way?
 
Make this Registrar aligned to `BaseRegistrar`?

For learning purpose, not a high priority to answer!

packages/ens-utils/src/ethregistrar.ts
```diff
@@ -3,323 +3,370 @@ import {
...
+    // last label must be "eth"
+    if (name.labels[1] !== ETH_TLD) return null;
+
+    // NOTE: now we know we have a direct subname of ".eth"
+
+    // first label must be of sufficient length
+    const subnameLength = charCount(name.labels[0]);
+    if (subnameLength < MIN_ETH_REGISTRABLE_LABEL_LENGTH) return null;
+
+    // TODO: also add a check for a maximum length limit as enforced by max block size, etc?
```
FrancoAguzzi, line 82: Sounds interesting

packages/ens-utils/src/ethregistrar.ts
```diff
@@ -3,323 +3,370 @@ import {
...
-  getTokenId(name: ENSName, isWrapped: boolean): TokenId {
-    if (!this.isClaimable(name, isWrapped)) {
+  public getRegistrationPriceQuote(
+    name: ENSName,
+    atTimestamp: Timestamp,
+    duration: Duration,
+    existingRegistration?: Registration,
+  ): RegistrationPriceQuote {
+    if (!this.canRegister(name, atTimestamp, duration, existingRegistration)) {
+      // TODO: refine the way we handle these exception cases.
```
FrancoAguzzi, line 143: Do you mean creating specific error handling for each use case?

e.g.

Error 1: `atTimestamp` cannot be in the past
Error 2: `duration` cannot be 0 seconds

And so on, creating better described errors... OR do you suggest something diff?

packages/ens-utils/src/ethregistrar.ts
```diff
@@ -3,323 +3,370 @@ import {
...
-  name: ENSName,
-  chain: ChainId,
-  isWrapped: boolean,
-): Registrar {
-  const registrars = getPotentialKnownRegistrars(name, chain, isWrapped);
-  if (registrars.length > 1) {
-    throw new Error(
-      `Multiple potential registrars found for name: "${name.name}" on chainId: ${chain.chainId} when isWrapped: ${isWrapped}`,
+    if (!existingRegistration.expirationTimestamp) {
+      throw new Error(`Invariant violation`); // TODO: refine message... just making the type system happy.
```
FrancoAguzzi, line 230:
```suggestion
      throw new Error(`Cannot get a renewal price quote for a domain that has no expirationTimestamp. Please confirm the domain's existingRegistration has an expirationTimestamp.`); 
```

packages/ens-utils/src/namekitregistrar.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (subnameLength === 3 || subnameLength == 4) {
+      annualServiceFee = buildPrice(999n, Currency.Usd); // $9.99 USD
+    } else {
+      annualServiceFee = buildPrice(99n, Currency.Usd); // $0.99 USD
+    }
+
+    const serviceFee: RegistrarServiceFee = {
+      type: RegistrarChargeType.ServiceFee,
+      price: annualServiceFee,
+      reason: "Example NameKit convenience fee.", // TODO: update this
```
FrancoAguzzi, line 34: @lightwalker-eth need any help with ideation?

packages/ens-utils/src/registrar.ts
```diff
@@ -0,0 +1,356 @@
...
+  GracePeriod = "GracePeriod",
+  RecentlyReleased = "RecentlyReleased",
+}
+
+// TODO: make more generic to support non .eth 2nd level domains
+export type Registration = {
+  // Below timestamps are counted in seconds
+  registrationTimestamp: Timestamp | null;
+  expirationTimestamp: Timestamp | null;
+  expiryTimestamp: Timestamp | null; // TODO: Franco, could you please remove this for us?
```
FrancoAguzzi, line 31: Yes, I will handle it [here](https://app.shortcut.com/ps-web3/story/25399/remove-duplicate-expirytimestamp-field)

```suggestion
  expiryTimestamp: Timestamp | null;
```

packages/ens-utils/src/registrar.ts
```diff
@@ -0,0 +1,356 @@
...
+    this.name = "RegistrarUnsupportedNameError";
+    Error.captureStackTrace(this, this.constructor);
+  }
+}
+
+export const scaleAnnualPrice = (annualPrice: Price, duration: Duration) => {
+  // Small performance optimization if no scaling is needed
+  if (duration.seconds === SECONDS_PER_YEAR.seconds) return annualPrice;
+
+  // TODO: verify we're doing this division of bigints correctly
```
FrancoAguzzi, line 327:
```suggestion
```

FrancoAguzzi (changes requested):

packages/ens-utils/src/ethregistrar.ts
lightwalker-eth: @FrancoAguzzi Yes that's right and a purposeful design decision here.

Please note the importance of what it means for this class to "model the policy implementation" it does not literally enforce the policy.

In other words, if all the inputs to this function were actually true, then this function would model the correct result. If the inputs to the function are not actually true, then it doesn't matter that's not what this class is for.

packages/ens-utils/src/ethregistrar.ts
lightwalker-eth: @FrancoAguzzi I see you have a few questions about the possibility of `atTimestamp` being in the past. It's important to adjust the way you're thinking about the goals this class works to achieve.

It's totally outside the scope of concerns for this class if the params you pass into it are aligned with reality or not. Instead, this class models the behaviour of the .eth registrar contracts assuming that the inputs you provide to it were actually reality.

packages/ens-utils/src/ethregistrar.ts
lightwalker-eth: @FrancoAguzzi I've added the correct constants now for `MIN_REGISTRATION_PERIOD`

packages/ens-utils/src/namekitregistrar.ts
lightwalker-eth: @FrancoAguzzi Right now my mindset is to optimize for merging this PR sooner rather than later. Thanks.

PR #345, opened 2024-08-06
Refine time logic

lightwalker-eth:


packages/ens-utils/src/time.test.ts
```diff
@@ -332,7 +332,44 @@ describe("formatTimestamp() function", () => {
     expect(result).toStrictEqual("Feb 1, 2024");
   });
 
-  it("showDateOnly - Greenwich time zone", () => {
+  // // TODO: build a more generic version of this unit test that works in any system default time zone
```
lightwalker-eth, line 335: @FrancoAguzzi We shouldn't remove or change these unit tests. Note what each of these unit tests are actually testing. Ex: No options or no explicit time zone. We can't implement these tests by providing options or an explicit time zone.

FrancoAguzzi: @lightwalker-eth  indeed, but we also won't be able to get this right due to system preferences... Each system: mine, yours, @notrab's, GH's virtual machine, etc. is in one timezone so has a own default. We either integrate one other data getter to gather the information relative to current timezone to fulfill `expectedResult` or we will never have our `expectedResult` string matching all systems...

lightwalker-eth: @FrancoAguzzi That's true, but it's also true that the current approach here is the best available option for now. There is still value in keeping these unit tests in commented out form as they can be verified locally when needed, but they need to remain commented out to ensure they don't break for someone.

There's nothing for us to change here right now.

PR #348, opened 2024-08-08
feat: [SC-25563] Create Tooltip component and Stories

FrancoAguzzi:


apps/storybook.namekit.io/stories/Namekit/Tooltip.stories.tsx
lightwalker-eth, line 1: @FrancoAguzzi I understand that we are following a strategy of temporarily leaving this UI component in NameGuard while duplicating it in its new home in NameKit.

Whenever we do this, please always take special actions to help us manage this better and not get super confused. For example:

1. Please add a comment at the top of the OLD UI component file identifying that it should not be used and making a reference to the NEW UI component that should be used instead.
2. Please create a shortcut story and link to it from that comment that helps us keep track of the goal and need to fully remove any use of the OLD UI component.

Thanks

apps/storybook.namekit.io/stories/Namekit/Tooltip.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+  },
+};
+
+export const DefaultMaxTooltipWidth: Story = {
+  args: {
+    placement: "bottom",
+    trigger: <>Tooltip Trigger</>,
+    children: (
+      <>
+        The default max tooltip width is the largest positive number possible so
```
lightwalker-eth, line 56:
```suggestion
        The default max tooltip width is positive infinity so
```

apps/storybook.namekit.io/stories/Namekit/Tooltip.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+        it is really up to you to decider when to limit the text content,
```
lightwalker-eth, line 57:
```suggestion
        it is up to you to set an explicit max width limit,
```

apps/storybook.namekit.io/stories/Namekit/Tooltip.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+        horizontally! Note: see example below to understand how to do it 😉
```
lightwalker-eth, line 58:
```suggestion
        See the "Custom Max Tooltip Width" story for an example of setting a max width.
```

apps/storybook.namekit.io/stories/Namekit/Tooltip.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export const CustomMaxTooltipWidth: Story = {
+  args: {
+    placement: "bottom",
+    maxTooltipWidth: 400,
+    trigger: <>Tooltip Trigger for tooltip with 400px max width</>,
+    children: (
+      <>
+        See it! The tooltip content is limited to 400px width. You can set the
+        maxTooltipWidth to any number you want to limit the text content display
```
lightwalker-eth, line 72: I don't think it's right to say "any number you want". That can accept silly values such as NaN, negative numbers, etc.. Better to just say "You can set maxTooltipWidth to the desired value to limit ..."

lightwalker-eth (changes requested):
@FrancoAguzzi Reviewed and shared feedback. When you submit a PR for review please always check first if a changeset is needed for that PR and if so, add the needed changsets before requesting review. Thanks

lightwalker-eth (approved):
@FrancoAguzzi thanks for actions here

PR #361, opened 2024-08-21
feat: migrate DisplayedPrice and create DisplayedPriceWithAltPrice

FrancoAguzzi:


lightwalker-eth (changes requested):
@FrancoAguzzi Thanks for your update here. I can see some good care went into this update :+1: Reviewed and shared feedback

packages/namekit-react/src/components/DisplayedPrice.tsx
```diff
@@ -0,0 +1,144 @@
...
+    case PriceDisplaySize.Large:
+      return CurrencySymbolSize.Large;
+  }
+};
+
+interface DisplayedPriceProps {
+  // The price to be displayed
+  price: Price;
+  // The onClick event handler for the price text
+  onTextClick?: () => void;
```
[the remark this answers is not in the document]
FrancoAguzzi: Hi @lightwalker-eth 

This can certainly be defined as `Button` code, removing the `onTextClick` param need

packages/namekit-react/src/components/DisplayedPrice.tsx
```diff
@@ -0,0 +1,144 @@
...
+  // Wether to display a currency symbology as its acronym or symbol
+  currencySymbology?: PriceSymbology;
```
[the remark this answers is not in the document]
FrancoAguzzi: It is the currency symbology (e.g. USDC symbology, DAI simbology, etc.) that is displayed for a price (which is defined in a currency, which is display either as a Symbol or an Acronym based on the definition of the flagged prop)

packages/namekit-react/src/components/DisplayedPrice.tsx
```diff
@@ -0,0 +1,144 @@
...
+            "nk-order-1 nk-mr-1":
+              symbolPosition === CurrencySymbolPosition.Left,
+            "nk-mr-1 nk-order-3":
+              symbolPosition === CurrencySymbolPosition.Right,
+            "nk-hidden":
+              !withSymbol || currencySymbology === PriceSymbology.Acronym,
+          },
+        ])}
+      >
+        <CurrencySymbol
```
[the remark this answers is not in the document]
FrancoAguzzi: That approach is good, can be useful for className and different size definitions, yes.

packages/namekit-react/src/components/DisplayedPrice.tsx
```diff
@@ -0,0 +1,144 @@
...
+                trigger={<p>{PriceCurrencyFormat[price.currency].Acronym}</p>}
+              >
+                {
+                  PriceCurrencyFormat[price.currency]
+                    .ExtendedCurrencyNameSingular
+                }
+              </Tooltip>
+            </div>
+          ) : (
+            <p className="nk-ml-1 nk-order-2">
```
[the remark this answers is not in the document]
FrancoAguzzi: We can store this elm in a `const` and reuse it in both places, indeed.

packages/namekit-react/src/components/DisplayedPrice.tsx
```diff
@@ -0,0 +1,144 @@
...
+      <div
+        className={cc([
+          "nk-leading-none",
+          {
+            "nk-mr-1": price.currency === Currency.Usd,
+            "nk-order-1 nk-mr-1":
+              symbolPosition === CurrencySymbolPosition.Left,
+            "nk-mr-1 nk-order-3":
+              symbolPosition === CurrencySymbolPosition.Right,
+            "nk-hidden":
```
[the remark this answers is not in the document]
FrancoAguzzi: No special reason, we can wrap the element in a condition like mentioned.

packages/namekit-react/src/components/DisplayedPrice.tsx
```diff
@@ -0,0 +1,144 @@
...
+      >
+        <CurrencySymbol
+          additionalClasses={onTextClick ? "nk-cursor-pointer" : ""}
+          describeCurrencyInTooltip={describeCurrencyInTooltip}
+          size={priceDisplaySizeAsCurrencyDisplaySize}
+          currency={price.currency}
+        />
+      </div>
+
+      {currencySymbology === PriceSymbology.Acronym && (
```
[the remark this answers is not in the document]
FrancoAguzzi: Here, `DisplayedPrice` is handling the currency symbol rendering for the given price:

It is not possible to include this kind of logic in a children component since the father component, the wrapper, is the one responsible for positioning the currency symbol relatively to the price.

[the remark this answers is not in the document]
FrancoAguzzi: It is pretty much saying:

Component A receives user input about **how** and **where** to place currency symbol relatively to price.
Component A renders either component B or C based on the **how**.
Component A always renders component D.

Component A then, is the one responsible for saying "Hey, based on user's preferences let's render either":
1. B + D
2. D + B
3. C + D
4. D + C

It is a wrapper that based on user preferences, displays the price with a currency symbology :)

I hope this clarifies what `DisplayedPrice` does, let me know if anything!

packages/namekit-react/src/components/DisplayedPriceWithAltPrice.tsx
```diff
@@ -0,0 +1,225 @@
...
+  (typeof AltPriceDisplayFormat)[keyof typeof AltPriceDisplayFormat];
+
+export const PriceDisplayPosition = {
+  Right: "nk-flex nk-inline-flex nk-items-end nk-space-x-2",
+  Bottom: "nk-flex nk-flex-col nk-text-right nk-items-end nk-space-y-1",
+} as const;
+export type PriceDisplayPosition =
+  (typeof PriceDisplayPosition)[keyof typeof PriceDisplayPosition];
+
+interface DisplayedPriceWithAltPriceProps {
```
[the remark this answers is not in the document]
FrancoAguzzi: Hey @lightwalker-eth , I understand your concerns here and will structure a proposal

packages/namekit-react/src/components/DisplayedPrice.tsx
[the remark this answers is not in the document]
FrancoAguzzi: I just reviewed this code and noticed that there are no repetitions here. 

We render twice `{PriceCurrencyFormat[price.currency].Acronym}`, but, with different classes

In one scenario we render it without classes, as the trigger of a tooltip

In another scenario we render it with classes that set spacing and order to it, as pure text.

packages/namekit-react/src/components/DisplayedPrice.tsx
```diff
@@ -35,62 +12,39 @@ export const PriceDisplaySize = {
...
 }
 
 export const DisplayedPrice = ({
   price,
-  describeCurrencyInTooltip = true,
+  symbol,
   displaySize = PriceDisplaySize.Small,
   tooltipTriggerToDisplayPriceInTooltip,
-  currencySymbology = PriceSymbology.Symbol,
-  symbolPosition = CurrencySymbolPosition.Left,
```
lightwalker-eth: @FrancoAguzzi Suggest we still keep this idea in `DisplayedPrice`. Please see how `IconButton` includes a prop for this idea and how they implemented it.

packages/namekit-react/src/components/DisplayedPriceWithSymbol.tsx
```diff
@@ -0,0 +1,168 @@
...
+    case PriceDisplaySize.Micro:
+    case PriceDisplaySize.Small:
+      return CurrencySymbolSize.Small;
+    case PriceDisplaySize.Medium:
+    case PriceDisplaySize.Large:
+      return CurrencySymbolSize.Large;
+  }
+};
+
+export const DisplayedPriceWithSymbol = ({
```
lightwalker-eth, line 72: I don't understand why this UI component should exist. In my mind it shouldn't -- or if it does it should be an extremely thin wrapper around other components.

packages/namekit-react/src/components/DisplayedPrice.tsx
```diff
@@ -152,15 +65,13 @@ export const DisplayedPrice = ({
           <>
```
lightwalker-eth, line 65: I don't understand why there's any `Tooltip` included in `DisplayedPrice`. I couldn't find any story where this is visible. In my mind there's no need for this. Am I missing something?

packages/namekit-react/src/components/DisplayedPrice.tsx
```diff
@@ -35,62 +12,39 @@ export const PriceDisplaySize = {
...
   displaySize?: PriceDisplaySize;
-  // Wether or not to display the name of the currency in a tooltip when its symbol is hovered
-  describeCurrencyInTooltip?: boolean;
-  // Whenever below prop is defined, the price is displayed inside a Tooltip which is triggered by the below content
+  /**
+   * Whenever below prop is defined, the price will be
+   * displayed inside a Tooltip and the below content
+   * will be the trigger of this Tooltip displaying.
+   */
   tooltipTriggerToDisplayPriceInTooltip?: React.ReactNode;
```
lightwalker-eth, line 39: I don't understand why there's any `Tooltip` included in `DisplayedPrice`. I couldn't find any story where this is visible. In my mind there's no need for this. Am I missing something?

lightwalker-eth (changes requested):
@FrancoAguzzi Thanks for updates. Reviewed and shared feedback

packages/namekit-react/src/components/DisplayedPriceWithSymbol.tsx
lightwalker-eth: @FrancoAguzzi In other words, there's way too much logic in this UI component. That's a strong sign that we aren't writing our code in a properly composable way. If you get the composition correct than there should be very little logic in this file.

lightwalker-eth: @FrancoAguzzi Very important: you need to think more about the context where these UI components are used in an overall app. You want to be giving power to that context and stop trying to control everything inside these UI components you're defining directly.

FrancoAguzzi: Hi @lightwalker-eth !

This component is a wrapper around DisplayedPrice that gives us the ability to render a price with:

1. Left aligned symbology
2. Right aligned symbology
3. Acronym symbology for inline DisplayedPrice 
4. Icon (Symbol) symbology for inline DisplayedPrice
5. Acronym symbology for _Tooltiped_ DisplayedPrice 
6. Icon (Symbol) symbology for _Tooltiped_ DisplayedPrice

The implementation is pretty concise, maybe what you are aiming here is not to achieve a smaller comp. definition but a comp. definition with fewer variants?

packages/namekit-react/src/components/DisplayedPrice.tsx
FrancoAguzzi: I assumed we handling displaying the price in a Tooltip would be a good feature, let me know if you find it best to remove this variant from DisplayedPrice internal logics. 

[Here](http://localhost:6006/?path=/story/namekit-displayedprice--eth-price-in-a-tooltip) is one example of **any** trigger being able to display the formatted price in a tooltip

packages/namekit-react/src/components/DisplayedPrice.tsx
FrancoAguzzi: I assumed we handling displaying the price in a Tooltip would be a good feature, let me know if you find it best to remove this variant from DisplayedPrice internal logics.

[Here](http://localhost:6006/?path=/story/namekit-displayedprice--eth-price-in-a-tooltip) is one example of any trigger being able to display the formatted price in a tooltip

packages/namekit-react/src/components/DisplayedPrice.tsx
lightwalker-eth: @FrancoAguzzi I can't see any reason why we would want this. To me this looks like completely the opposite from the goal of composability that I keep asking you to put more attention to.

If someone wants to put the string for a price into a tooltip message, why in the world would they want to do it this way? I'm struggling to put into words just how wrong this whole direction is. Please change it!

packages/namekit-react/src/components/DisplayedPrice.tsx
```diff
@@ -0,0 +1,75 @@
...
+export interface DisplayedPriceProps {
+  /**
+   * The price to be displayed
+   * @example { currency: Currency.Eth, value: 1000000000000000000n }
+   */
+  price: Price;
+  /**
+   * The currency symbol to display alongside the price.
+   */
+  symbol: React.ReactNode;
```
lightwalker-eth, line 36: I see from storybook that this param is optional.

Therefore it should be defined with `?` in this interface and we also need to document how it is optional and what happens if it is not defined.

One idea is to make it so that `symbol` is defined as `symbol?: React.ReactNode | null`.

Then, if `symbol` is defined and non-null, we use the defined symbol. Great! This gives full flexibility to people who use `DisplayedPrice`.

If `symbol` is defined and `null` then we can render no symbol at all. This should be a special exception case and we can create a single story for this in storybook. There's no need for so many examples in Storybook to show the idea of hiding the symbol.

If `symbol` is undefined then I'm happy to then have some default definition of the symbol to use. Ex: A `CurrencySymbol` of the appropriate `Currency`, size, and with the tooltip turned on.

It might sound like we're going in circles because I was just making a big push for composability, but please note how we have achieved our composability goals: A fully custom `symbol` can now be provided when desired. Therefore we can have a very simple implementation of a default `symbol` if it is undefined. It's nice to provide defaults that are generally ok. It's important that we don't add any props to `DisplayedPrice` that are for the default `CurrencySymbol` though. If someone wants to customize the `CurrencySymbol` they should pass in a custom value into the `symbol` prop!

Open to feedback on everything above.

apps/storybook.namekit.io/stories/Namekit/DisplayedPrice.stories.tsx
```diff
@@ -0,0 +1,694 @@
...
+};
+export const EthPriceWithAcronym: Story = {
+  args: {
+    price: {
+      currency: Currency.Eth,
+      value: 1000000000000000000n,
+    },
+    symbol: (
+      <span>
+        {getCurrencySymbology(Currency.Eth, CurrencySymbology.Acronym)}
```
lightwalker-eth, line 306: Suggest updating the `CurrencySymbol` UI component so that it has a new prop: `symbology`. This can take in a value of type `CurrencySymbology`.

You should then move these acronym and other custom symbology use cases to Storybook stories for `CurrencySymbol`.

In other words, "ETH" is also a `CurrencySymbol`! The `CurrencySymbol` UI component should have the logic inside of itself to handle this case.

When we write the JSX for an app UI, we want to make the markup nicely self documenting. Ex: All these cases are just `CurrencySymbol` with different symbology.

packages/ens-utils/src/price.ts
```diff
@@ -123,14 +128,44 @@ export const multiplyPriceByNumber = (price1: Price, price2: number): Price => {
   };
 };
 
+export const CurrencySymbology = {
```
lightwalker-eth, line 131: This idea doesn't belong inside `ens-utils`. It should move into `CurrencySymbol` in `namekit-react`.

packages/ens-utils/src/price.ts
```diff
@@ -123,14 +128,44 @@ export const multiplyPriceByNumber = (price1: Price, price2: number): Price => {
...
+export type CurrencySymbology =
+  (typeof CurrencySymbology)[keyof typeof CurrencySymbology];
+
+/**
+ * Returns a string containing the currency's symbol or acronym.
+ * @param currency: Currency. The currency to get the symbology for (e.g. Currency.Eth)
+ * @param symbology: CurrencySymbology. The symbology to use (e.g. CurrencySymbology.Symbol)
+ * @returns: string. The currency's symbol or acronym (e.g. "$" or "USD")
+ */
+export const getCurrencySymbology = (
```
lightwalker-eth, line 150: This idea doesn't belong inside `ens-utils`. It should move into `CurrencySymbol` in `namekit-react`.

packages/namekit-react/src/components/DomainCardPrice.tsx
```diff
@@ -0,0 +1,41 @@
...
+          displaySize={props.displaySize}
+          symbolPosition={props.symbolPosition}
+        />
+      }
+    >
+      <DisplayedPrice
+        price={altPrice}
+        symbol={
+          <span>
+            {getCurrencySymbology(altPrice.currency, altPriceSymbology)}
```
lightwalker-eth, line 35: Suggest we update this to also use `CurrencySymbol` based on other suggestions I've shared. We can just pass in a custom `symbology` to that component if it is needed.

packages/namekit-react/src/components/DomainCardPrice.tsx
```diff
@@ -0,0 +1,41 @@
...
+} from "@namehash/ens-utils";
+import { Tooltip } from "./Tooltip";
+import React from "react";
+
+interface DomainCardPriceProps extends DisplayedPriceProps {
+  altPrice: Price;
+  altPriceSymbology: CurrencySymbology;
+}
+
+export const DomainCardPrice = ({
```
lightwalker-eth, line 15: I like what you've done with this UI component :+1:

We shouldn't call it a `DomainCardPrice` though because this idea isn't specific to `DomainCard`.

Maybe `DisplayedPriceConversion`? Assuming so we should also rename `altPrice` and related fields to something such as `convertedPrice`, etc..

apps/storybook.namekit.io/stories/Namekit/DomainCardPrice.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+};
+
+export default meta;
+
+type Story = StoryObj<typeof DomainCardPrice>;
+
+export const EthPriceWithCurrencySymbolAndAnUsdAltPriceWithSymbolSymbology: Story =
+  {
+    args: {
+      price: { currency: Currency.Eth, value: 1000000000000000000n },
```
lightwalker-eth, line 42: Unless there's a special reason not to for a particular story, let's please use the `numberToPrice` function for all of the price values in all storybook stories. Then you can just just do `numberToPrice(1)` to get 1 ETH.

We should always use these utility functions to build our special types and never explicitly build them the way you are here. For example, we might change the utility functions to add special invariants in the future, but if you build them explicitly as you are here then it would create bugs and we might not notice for some time.

apps/storybook.namekit.io/stories/Namekit/DisplayedPrice.stories.tsx
```diff
@@ -0,0 +1,694 @@
...
+  getCurrencySymbology,
+} from "@namehash/ens-utils";
+import React from "react";
+
+const meta: Meta<typeof DisplayedPrice> = {
+  component: DisplayedPrice,
+  title: "Namekit/DisplayedPrice",
+  argTypes: {
+    price: {
+      control: "object",
```
lightwalker-eth, line 20: Let's please better configure our Storybook stories so that we never have controls of type "object".

CleanShot 2024-08-25 at 18 48 00

These "object" controls aren't usable and provide a bad UX in storybook.

One of the goals of Storybook is to ensure its easy for people to use the controls to see the UI component in various states.

This means you need to separate the idea of the schema for the controls in Storybook from the schema of the props that are actually passed into the constructor of the UI component. Please see the docs for Storybook they have examples for how to do this.

We should see a control in storybook that has a nice UX for picking the currency and a different control that has a nice UX for picking the price value.

You might want to consider also making it so that the control that picks the price value is just used as input to the `numberAsPrice` function to get the `Price` object that we actually pass into `DisplayedPrice`. Otherwise people can get confused when they first look at the storybook without understanding all the technical background.

lightwalker-eth (changes requested):
@FrancoAguzzi Thanks for updates. It's nice to see that this is becoming more composable. There's still some key points to improve. Reviewed and shared feedback.

packages/namekit-react/src/components/DomainCardPrice.tsx
FrancoAguzzi: Sounds good, @lightwalker-eth, 🔝 I will do so

packages/namekit-react/src/components/CurrencySymbol/CurrencySymbol.tsx
```diff
@@ -1,81 +1,107 @@
...
 import React from "react";
 
-export enum CurrencySymbolSize {
-  Small = "nk-w-4",
-  Large = "nk-w-5",
-}
+/**
+ * The size of the `CurrencySymbol`.
+ * This is defined as a mapping from a `CurrencySymbolSize` to a number.
+ * This number represents the size a `CurrencySymbol` should be displayed at, in pixels.
```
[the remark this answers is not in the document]
FrancoAguzzi: @lightwalker-eth maybe what we need is to rename it?

Because the same symbology rules are used in `formattedPrice`, an `ens-utils` feature

packages/namekit-react/src/components/CurrencySymbol/CurrencySymbol.tsx
```diff
@@ -1,81 +1,107 @@
...
+}: CurrencySymbolProps): JSX.Element => {
   let symbol: JSX.Element;
 
   switch (currency) {
     case Currency.Usd:
       symbol = (
-        <p className="nk--mr-1" style={{ color: symbolFillColor }}>
-          $
-        </p>
+        <span
```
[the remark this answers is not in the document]
FrancoAguzzi: Optimized :) Thx

packages/namekit-react/src/components/DisplayedPrice.tsx
```diff
@@ -0,0 +1,75 @@
...
+   */
+  symbolPosition?: CurrencySymbolPosition;
+  /**
+   * The position of the currency symbol relative to the price.
+   * Defaults to `CurrencySymbolPosition.Left`.
+   *
+   * If `CurrencySymbolPosition.Left`, the currency symbol will be displayed to the left of the price.
+   * If `CurrencySymbolPosition.Right`, the currency symbol will be displayed to the right of the price.
+   */
+  displaySize?: PriceDisplaySize;
```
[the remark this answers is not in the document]
FrancoAguzzi: My bad

packages/ens-utils/src/price.ts
FrancoAguzzi: @lightwalker-eth maybe what we need is to rename it?

Because the same symbology rules are used in formattedPrice, an ens-utils feature

packages/namekit-react/src/components/DisplayedPrice.tsx
FrancoAguzzi: @lightwalker-eth ok sounds good! i'd only like to hear back a feedback from the getCurrencySymbology placement first, I plan to reuse it here for default use case :)

apps/storybook.namekit.io/stories/Namekit/DisplayedPrice.stories.tsx
FrancoAguzzi: Sounfs good idea!

packages/ens-utils/src/price.ts
FrancoAguzzi: @lightwalker-eth maybe what we need is to rename it?

Because the same symbology rules are used in formattedPrice, an ens-utils feature

packages/namekit-react/src/components/DomainCardPrice.tsx
FrancoAguzzi: perfect!

apps/storybook.namekit.io/stories/Namekit/DisplayedPrice.stories.tsx
FrancoAguzzi: Not sure how to do it but will manage doing so :)

lightwalker-eth: @FrancoAguzzi Check out the docs for storybook. That should give some ideas.

packages/namekit-react/src/components/DisplayedPrice.tsx
lightwalker-eth: @FrancoAguzzi I continue to believe that `CurrencySymbology` and `getCurrencySymbology` do not belong in `ens-utils`.

Please have a look at the changes that this PR is proposing to `formattedPrice` in `ens-utils`. `formattedPrice` in `ens-utils` worked just fine without `CurrencySymbology` before in the past. I don't see any reason why it should be added now?

In my thinking all changes to `packages/ens-utils/src/price.ts` in this PR should be reverted.

Please let me know if I'm missing something important?

lightwalker-eth (changes requested):
@FrancoAguzzi Reviewed and shared feedback :+1:

packages/namekit-react/src/components/CurrencySymbol/CurrencySymbol.tsx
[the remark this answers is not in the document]
FrancoAguzzi: My bad on this latest feedback

packages/namekit-react/src/components/CurrencySymbol/CurrencySymbol.tsx
```diff
@@ -1,81 +1,107 @@
...
 
   /**
    * If `true`, hovering over the `CurrencySymbol` will display the
    * name of `currency` in a `Tooltip`. If `false` then the `CurrencySymbol`
    * won't have any `Tooltip`.
-   * 
+   *
    * Defaults to `true`.
    */
   describeCurrencyInTooltip: boolean;
```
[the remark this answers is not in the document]
FrancoAguzzi: Now you are able to see it :)

apps/storybook.namekit.io/stories/Namekit/DisplayedPrice.stories.tsx
FrancoAguzzi: @lightwalker-eth the best way to provide users the ability to set such arguments (React.ReactNode ones or typed Objects) is to provide multiple options for them. I have created multiple symbol and price options so people can test alternative examples of our Stories. Please let me know if you would like to see different options in our documentation!

packages/namekit-react/src/components/DisplayedPrice.tsx
FrancoAguzzi: @lightwalker-eth it should be good to at least specify `describeCurrencyInTooltip` as in `DisplayedPriceConversion` we want to avoid this tooltip in `DisplayedPrice` rendering:
Captura de Tela 2024-08-29 às 19 41 38

apps/storybook.namekit.io/stories/Namekit/DisplayedPriceConversion.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+  CurrencySymbol,
+  CurrencySymbology,
+  CurrencySymbolPosition,
+  DisplayedPriceConversion,
+  PriceDisplaySize,
+} from "@namehash/namekit-react";
+import type { Meta, StoryObj } from "@storybook/react";
+import React from "react";
+
+const meta: Meta<typeof DisplayedPriceConversion> = {
```
lightwalker-eth, line 12: @FrancoAguzzi I don't understand what is causing all the issues and mistakes that are happening. Can you please help me understand what is going wrong and causing so many issues?

For example, all of the current stories for `DisplayedPriceConversion` totally fail to communicate the big idea for this UI component. Maybe that's because you don't understand what the big idea is yourself? If you don't know you need to be asking more questions. The big idea for this UI component is to handle the UX for displaying a price in multiple currencies. For example, a NFT might be priced at 1 ETH, but the user might want to also see what that price represents in USD.

If you understand the big idea for this UI component, then why would the stories that are defined here all be so useless and confusing? For example, it's useless to show a conversion from ETH to ETH or from ETH to WETH. That's silly. I shouldn't need to catch these kinds of mistakes myself. You're making our UI components look pointless and confusing by failing to communicate the big idea to the audience who will look at these stories.

The next big issue is I continue to see a clear lack of you doing your own testing and quality control. For example, check the screenshot below:

CleanShot 2024-08-30 at 13 58 32

Before you submit a PR for review, you should always first take on the responsibility of reviewing everything that PR is changing and verifying that all outputs of that changed code is correct. How do you think my process works where I review a PR and find so many issues to fix? I'm following a very simple process to find all of these problems. You should be following this process yourself and finding + fixing these issues before you ask me to review your work.

packages/namekit-react/src/components/CurrencySymbol/CurrencySymbol.tsx
lightwalker-eth, line 1: I note how `CurrencySymbol` is inside a subdirectory, unlike other UI components in this package.

Would it make sense to move `CurrencySymbol` directly into the `components` subdirectory like all other UI components in this package?

Then maybe we create a new directory for `packages/namekit-react/src/icons` and move all the icon components there?

Appreciate your advice.

apps/storybook.namekit.io/stories/Namekit/DisplayedPrice.stories.tsx
```diff
@@ -40,654 +86,201 @@ export default meta;
...
       <CurrencySymbol
         currency={Currency.Usd}
-        describeCurrencyInTooltip={true}
+        symbology={CurrencySymbology.Icon}
       />
     ),
   },
 };
-export const UsdMaxDisplayPriceWithSymbol: Story = {
+export const UsdMaxDisplayPriceWithCustomCurrencyIcon: Story = {
```
lightwalker-eth, line 211: We should rename these "max" or "min" price stories.

Instead of "max" we should use the word "overflow".

Instead of "min" we should use the word "underflow".

It's true that each `Currency` has a min or max display price. But those are single fixed values. There can be an infinite number of `Price` values that are higher than the max display price for a given `Currency. It doesn't matter which of those values we choose in the story, whenever there is an "overflow" we display the max display price.

Does this make sense? Please feel welcome to ask me questions.

lightwalker-eth (changes requested):

apps/storybook.namekit.io/stories/Namekit/DisplayedPriceConversion.stories.tsx
FrancoAguzzi: @lightwalker-eth https://github.com/namehash/namekit/pull/361#discussion_r1737312089

packages/namekit-react/src/components/CurrencySymbol/CurrencySymbol.tsx
```diff
@@ -40,74 +57,104 @@ interface CurrencySymbolProps extends React.ComponentProps<React.ElementType> {
...
+/**
+ * Returns a string containing the currency's symbol or acronym.
+ * @param currency: Currency. The currency to get the symbology for (e.g. Currency.Eth)
+ * @param symbology: CurrencySymbolSize. The size to use for Icon Symbology.
+ *                   For other symbologies it is ignored. We suggest you use props to define
+ *                   other symbologies sizes, as these are not SVGs but texts, instead. (e.g. className="myCustomClassForTextSize")
+ * @param symbology: CurrencySymbology. The symbology to use (e.g. CurrencySymbology.Symbol)
+ * @returns: string. The currency's symbol or acronym (e.g. "$" or "USD")
+ */
+export const getCurrencySymbology = ({
```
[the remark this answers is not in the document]
FrancoAguzzi: No

packages/namekit-react/src/components/DisplayedPrice.tsx
```diff
@@ -32,42 +37,71 @@ export interface DisplayedPriceProps {
...
+};
+
 export const DisplayedPrice = ({
   price,
   symbol,
-  displaySize = PriceDisplaySize.Small,
   symbolPosition = CurrencySymbolPosition.Left,
+  displaySize = PriceDisplaySize.Small,
 }: DisplayedPriceProps) => {
+  const displayDefaultSymbol = typeof symbol === "undefined";
```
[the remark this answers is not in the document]
FrancoAguzzi: @lightwalker indeed it looks _cowboy coding mode_!

Please check below assertions:

Captura de Tela 2024-08-31 às 17 27 00

As symbol is defined as `symbol?: React.ReactNode | null`, which means `undefined | React.ReactNode | null`

And, above truths,

We cannot handle knowing if symbol is defined by using Boolean or !! syntaxes. We can know by knowing the type of the prop, as it is being checked above. Best practice reference told me that, yet, it is better to do so by `symbol === undefined` than `typeof symbol === undefined`, so I have updated the code 🙂

apps/storybook.namekit.io/stories/Namekit/DisplayedPrice.stories.tsx
FrancoAguzzi: It does! Do you think I should rename this story, thus?

packages/namekit-react/src/components/CurrencySymbol/CurrencySymbol.tsx
FrancoAguzzi: Sounds good, @lightwalker-eth ! The structure looks good standard

apps/storybook.namekit.io/stories/Namekit/DisplayedPrice.stories.tsx
```diff
@@ -40,654 +86,201 @@ export default meta;
...
       <CurrencySymbol
         currency={Currency.Eth}
-        describeCurrencyInTooltip={true}
+        symbology={CurrencySymbology.Icon}
       />
     ),
   },
 };
-export const UsdMinDisplayPriceWithSymbol: Story = {
+export const UsdMinDisplayPriceWithCustomCurrencyIcon: Story = {
```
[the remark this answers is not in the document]
FrancoAguzzi: Hey, no big idea behind this. Goal was that, since we have a lot of Currency possibilities, to display how to use our components for building each Currency displaying price. I find it good to make it summarized for one or two examples, people might get the idea when they notice that CurrencySymbol has support to more currencies

apps/storybook.namekit.io/stories/Namekit/DisplayedPriceConversion.stories.tsx
FrancoAguzzi: Hi @lightwalker-eth !

About: 
1. Examples used: I like how WETH is displayed visually but you are correct :: WETH <> ETH conversions are not very intelligent!! I understand how this is well known in the ecosystem, the goal was to display this component beautifully and I am happy to exchange WETH to DAI to continue doing so, in a more intelligent way;
2. Duplicate of Tooltip: [I've mentioned](https://github.com/namehash/namekit/pull/361#discussion_r1737312089) how it would be nice to add a custom `describeCurrencyInATooltip` prop in `DisplayedPrice`-- we have previously decided to not add `CurrencySymbol` logics to `DisplayedPrice` for separation of ideas but now we see how this is needed in order to remove the default `symbol` custom prop (which aims to remove the `Tooltip` of `CurrencySymbol` whenever the `DisplayedPrice` is rendered inside `DisplayedPriceConversion`)

apps/storybook.namekit.io/stories/Namekit/DisplayedPrice.stories.tsx
FrancoAguzzi: It does!

lightwalker-eth (changes requested):

.changeset/quick-ads-pretend.md
```diff
@@ -0,0 +1,6 @@
+---
+"@namehash/namekit-react": minor
+"@namehash/ens-utils": minor
+---
+
+Create DisplayedPrice and DisplayedPriceWithSymbol components + new price utilities
```
lightwalker-eth, line 6:
```suggestion
Create DisplayedPrice and DisplayedPriceConversion components + new price utilities
```

Also, shouldn't there be something in this changeset description about `CurrencySymbol` too?

packages/namekit-react/src/components/CurrencySymbol.tsx
```diff
@@ -0,0 +1,161 @@
...
+  currency,
+  iconSize = CurrencySymbolSize.Small,
+  ...props
+}: {
+  iconSize: CurrencySymbolSize;
+  currency: Currency;
+}): JSX.Element => {
+  switch (currency) {
+    case Currency.Usd:
+      return <UsdSymbol {...props} />;
```
lightwalker-eth, line 88: Suggest making a comment here explaining how for USD we always display it as a text symbol and never as an icon.

packages/namekit-react/src/components/icons/DaiSymbol.tsx
lightwalker-eth, line 1: Suggest we rename each of the "(currency)Symbol.tsx" icons to "(currency)Icon.tsx" to be more precise and clear that they are the specific form of symbol known as an icon.

This renaming should influence both the name of the UI component and of the related .tsx file.

We also have too much opportunity for a misunderstanding inside `CurrencySymbology` with `CurrencySymbology.Symbol`. A lot of people can read that and assume it would be a graphical symbol but in reality it is a text-based symbol. Therefore suggest that we rename `CurrencySymbology.Symbol` to `CurrencySymbology.TextSymbol`.

This also gives us a nice solution for how we handle `UsdSymbol` which we don't actually render as an icon. For USD I suggest we  rename it to something like "UsdTextSymbol" to make it explicitly clear that it is a text symbol and not an icon.

packages/namekit-react/src/components/CurrencySymbol.tsx
```diff
@@ -0,0 +1,161 @@
...
+import { EthSymbol } from "./icons/EthSymbol";
+import { DaiSymbol } from "./icons/DaiSymbol";
+import React from "react";
+
+/**
+ * The size of the `CurrencySymbol`.
+ * This is defined as a mapping from a `CurrencySymbolSize` to a number.
+ * This number represents the size (width and height) a `CurrencySymbol` should be displayed at, in pixels.
+ */
+export const CurrencySymbolSize = {
```
lightwalker-eth, line 16: Shouldn't we rename this to `CurrencyIconSize` to be more clear it is only relevant for icons and not other symbologies?

packages/namekit-react/src/components/CurrencySymbol.tsx
```diff
@@ -0,0 +1,161 @@
...
+    case Currency.Usdc:
+      return <UsdcSymbol {...props} width={iconSize} height={iconSize} />;
+    case Currency.Dai:
+      return <DaiSymbol {...props} width={iconSize} height={iconSize} />;
+    case Currency.Weth:
+      return <WethSymbol {...props} width={iconSize} height={iconSize} />;
+    case Currency.Eth:
+      return <EthSymbol {...props} width={iconSize} height={iconSize} />;
+    default:
+      // TODO: We haven't created symbols for `Currency.Gas` yet.
```
lightwalker-eth, line 98: I'm going to ask Serge to make an icon for `Currency.Gas` for us 😄

packages/namekit-react/src/components/CurrencySymbol.tsx
```diff
@@ -0,0 +1,161 @@
...
+    case CurrencySymbology.Icon:
+      return getCurrencyIcon({ currency, iconSize, ...props });
+  }
+};
+
+export const CurrencySymbol = ({
+  currency,
+  iconSize = CurrencySymbolSize.Small,
+  describeCurrencyInTooltip = true,
+  symbology = CurrencySymbology.Symbol,
```
lightwalker-eth, line 138: The `symbology` param is more logical to put after `currency` and not as the last param. That's because `iconSize` comes "downstream" of the `symbology` param.

packages/namekit-react/src/components/CurrencySymbol.tsx
```diff
@@ -0,0 +1,161 @@
...
+ */
+const getCurrencySymbology = ({
+  currency,
+  iconSize = CurrencySymbolSize.Small,
+  symbology = CurrencySymbology.Symbol,
+  ...props
+}: {
+  currency: Currency;
+  iconSize?: CurrencySymbolSize;
+  symbology: CurrencySymbology;
```
lightwalker-eth, line 122: The `symbology` param is more logical to put after `currency` and not as the last param. That's because `iconSize` comes "downstream" of the `symbology` param.

packages/namekit-react/src/components/CurrencySymbol.tsx
```diff
@@ -0,0 +1,161 @@
...
+      throw new Error(
+        `Error creating CurrencySymbol: unsupported Currency: "${currency}".`,
+      );
+  }
+};
+
+/**
+ * Returns a JSX.Element containing the currency's symbol, acronym or icon.
+ * @param currency: Currency. The currency to get the symbology for (e.g. Currency.Eth)
+ * @param symbology: CurrencySymbolSize. The size to use for Icon Symbology.
```
lightwalker-eth, line 108:
```suggestion
 * @param iconSize: CurrencySymbolSize. The size to use for Icon Symbology.
```

packages/namekit-react/src/components/CurrencySymbol.tsx
```diff
@@ -0,0 +1,161 @@
...
+import { UsdcSymbol } from "./icons/UsdcSymbol";
+import { WethSymbol } from "./icons/WethSymbol";
+import { EthSymbol } from "./icons/EthSymbol";
+import { DaiSymbol } from "./icons/DaiSymbol";
+import React from "react";
+
+/**
+ * The size of the `CurrencySymbol`.
+ * This is defined as a mapping from a `CurrencySymbolSize` to a number.
+ * This number represents the size (width and height) a `CurrencySymbol` should be displayed at, in pixels.
```
lightwalker-eth, line 14:
```suggestion
 * This number represents the size (width and height) that the icon for a `CurrencySymbol` should be displayed at, in pixels.
```

packages/namekit-react/src/components/CurrencySymbol.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+import { UsdSymbol } from "./icons/UsdSymbol";
+import { UsdcSymbol } from "./icons/UsdcSymbol";
+import { WethSymbol } from "./icons/WethSymbol";
+import { EthSymbol } from "./icons/EthSymbol";
+import { DaiSymbol } from "./icons/DaiSymbol";
+import React from "react";
+
+/**
+ * The size of the `CurrencySymbol`.
```
lightwalker-eth, line 12:
```suggestion
 * The size of the `CurrencySymbol` icon.
```

lightwalker-eth (changes requested):
@FrancoAguzzi Here's some additional feedback in a second PR review. Please see my additional review above for more feedback.

packages/namekit-react/src/components/CurrencySymbol.tsx
FrancoAguzzi: True!

packages/namekit-react/src/components/CurrencySymbol.tsx
```diff
@@ -133,9 +135,9 @@ const getCurrencySymbology = ({
...
   currency,
-  iconSize = CurrencySymbolSize.Small,
-  describeCurrencyInTooltip = true,
-  symbology = CurrencySymbology.Symbol,
+  iconSize = CurrencyIconSize.Small,
+  describeCurrencyInTooltip = false,
+  symbology = CurrencySymbology.TextSymbol,
   ...props
 }: CurrencySymbolProps): JSX.Element => {
   const symbologyToDisplay = (
```
lightwalker-eth, line 143: This comment is for the `aria-label` a few lines below this. I notice that we incorrectly hardcoded that to always be `Currency.Eth` and not to use the specific `currency` that was provided.

apps/storybook.namekit.io/stories/Namekit/CurrencySymbol.stories.tsx
```diff
@@ -1,96 +1,90 @@
...
+      control: { type: "select" },
+    },
   },
-};
-export const DAI: Story = {
   args: {
-    currency: Currency.Dai,
-    size: CurrencySymbolSize.Large,
+    iconSize: CurrencyIconSize.Small,
+    describeCurrencyInTooltip: true,
```
lightwalker-eth, line 45: This is not in alignment with the new default args.

apps/storybook.namekit.io/stories/Namekit/CurrencySymbol.stories.tsx
```diff
@@ -1,96 +1,90 @@
...
+
+export const AsAnAcronym: Story = {
   args: {
-    size: CurrencySymbolSize.Small,
     currency: Currency.Eth,
+    symbology: CurrencySymbology.Acronym,
   },
 };
-export const LargeSize: Story = {
+export const AsAnIcon: Story = {
```
lightwalker-eth, line 60: We're missing a story for as a text symbol. Please add it.

apps/storybook.namekit.io/stories/Namekit/CurrencySymbol.stories.tsx
```diff
@@ -1,96 +1,90 @@
...
+    className: { control: { type: "text" } },
+  },
   args: {
-    describeCurrencyInTooltip: true,
-    size: CurrencySymbolSize.Large,
     currency: Currency.Eth,
+    className: "nk-text-3xl colorful-text",
   },
 };
 export const NotShowingTooltipDescription: Story = {
```
lightwalker-eth, line 85: This story should be inverted now that the default is `false`. Therefore we now need a story showing the `true` case. Please be sure to update the name of the story too.

apps/storybook.namekit.io/stories/Namekit/CurrencySymbol.stories.tsx
```diff
@@ -1,96 +1,90 @@
...
+type Story = StoryObj<typeof CurrencySymbol>;
+
+export const AsAnAcronym: Story = {
   args: {
-    size: CurrencySymbolSize.Small,
     currency: Currency.Eth,
+    symbology: CurrencySymbology.Acronym,
   },
 };
-export const LargeSize: Story = {
```
lightwalker-eth, line 42: We should have a story for a large size icon.

lightwalker-eth (changes requested):

packages/namekit-react/src/components/CurrencySymbol.tsx
```diff
@@ -62,69 +64,69 @@ interface CurrencySymbolProps extends React.ComponentProps<React.ElementType> {
...
   currency: Currency;
+  iconSize: CurrencyIconSize;
 }): JSX.Element => {
   switch (currency) {
     case Currency.Usd:
-      return <UsdSymbol {...props} />;
+      /**
+       * USD is always displayed as a text symbol, never being represented as a SVG icon.
+       */
+      return <UsdTextSymbol {...props} />;
```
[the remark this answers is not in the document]
FrancoAguzzi: I could achieve this by understanding this goal:

Captura de Tela 2024-09-11 às 15 08 02

And arriving at this solution:

```
/**
 *
 * @param currency: Currency. The currency to get the Icon for (e.g. Currency.Eth)
 * @param iconSize: CurrencyIconSize. The Icon size (width and height), in pixels.
 * @returns
 */
const getCurrencyIcon = ({
  currency,
  iconSize = CurrencyIconSize.Small,
  ...props
}: {
  currency: Currency;
  iconSize: CurrencyIconSize;
}): JSX.Element => {
  let symbology = <></>;

  switch (currency) {
    case Currency.Usd:
      /**
       * USD is always displayed as a text symbol, never being represented as a SVG icon.
       */
      symbology = (
        <CurrencySymbol
          {...props}
          currency={currency}
          symbology={CurrencySymbology.TextSymbol}
        />
      );
      break;
    case Currency.Usdc:
      symbology = <UsdcIcon {...props} width={iconSize} height={iconSize} />;
      break;
    case Currency.Dai:
      symbology = <DaiIcon {...props} width={iconSize} height={iconSize} />;
      break;
    case Currency.Weth:
      symbology = <WethIcon {...props} width={iconSize} height={iconSize} />;
      break;
    case Currency.Eth:
      symbology = <EthIcon {...props} width={iconSize} height={iconSize} />;
      break;
    case Currency.Gas:
      symbology = <GasIcon {...props} width={iconSize} height={iconSize} />;
      break;
  }

  return (
    <span
      aria-label={`${PriceCurrencyFormat[currency].ExtendedCurrencyNameSingular} symbol`}
    >
      {symbology}
    </span>
  );
};
```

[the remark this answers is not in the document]
FrancoAguzzi: It fixes the previous problem and it matches our goal 👍🏼

BUT: is it good?
IMO it is. All icons returned by this function will now have a custom icon "alt text" set.

packages/namekit-react/src/components/CurrencySymbol.tsx
```diff
@@ -46,11 +48,11 @@ interface CurrencySymbolProps extends React.ComponentProps<React.ElementType> {
   /**
    * The size of the `CurrencySymbol` icon.
    *
-   * Defaults to `CurrencySymbolSize.Small`.
+   * Defaults to `CurrencyIconSize.Small`.
    * Only applicable when `symbology` is `CurrencySymbology.Icon`.
-   * If you want to set a custom size for other symbologies, use props to define it (e.g. className="myCustomClassForTextSize")
+   * If you want to set a custom appearance for other text-based symbologies, use the `className` prop to define it (e.g. className="myCustomClassForTextSize")
```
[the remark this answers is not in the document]
FrancoAguzzi: I like the explicit setting 👍🏼

PR #364, opened 2024-08-22
Fix CI and Build outputs

notrab:


packages/nameguard-sdk/src/index.ts
```diff
@@ -494,16 +494,15 @@ export interface SecurePrimaryNameOptions {
...
+interface FakeEthNameOptions {}
 
 interface FieldsWithRequiredTitle extends Record<string, string> {
   title: string;
 }
 
 export class NameGuard {
   private endpoint: URL;
-  protected network: Network;
+  public network: Network;
```
lightwalker-eth, line 505: We need to make this readonly

packages/ens-utils/tsup.config.ts
```diff
@@ -2,9 +2,10 @@ import { defineConfig } from "tsup";
...
 export default defineConfig({
   entry: ["src/index.ts"],
-  splitting: true,
-  sourcemap: false,
+  splitting: false,
+  sourcemap: true,
   clean: true,
+  bundle: true,
   dts: true,
   treeshake: true,
```
lightwalker-eth, line 10: I notice how we have `treeshake` configured for `ens-utils` but not other packages. Is there any special reason all packages don't use the same config?

packages/nameguard-js/tsup.config.js
```diff
@@ -8,4 +8,5 @@ export default defineConfig({
   clean: true,
   bundle: true,
   dts: true,
+  external: ["@namehash/nameguard"],
```
lightwalker-eth, line 11: Interesting. I assumed we wouldn't need to specify anything explicit as `external`? As I understand any dependencies in package.json are automatically excluded? https://tsup.egoist.dev/#excluding-packages

packages/namekit-react/tsup.config.js
```diff
@@ -3,7 +3,10 @@ import { defineConfig } from "tsup";
 export default defineConfig({
   entry: ["src/index.ts"],
   format: ["esm"],
+  external: ["react", "react-dom"],
```
lightwalker-eth, line 6: Interesting. I assumed we wouldn't need to specify anything explicit as `external`? As I understand any dependencies in package.json are automatically excluded? https://tsup.egoist.dev/#excluding-packages

packages/ens-utils/tsup.config.ts
```diff
@@ -2,9 +2,10 @@ import { defineConfig } from "tsup";
...
   target: [
```
lightwalker-eth, line 11: I notice a few distinct configurations here for `ens-utils` but not other packages. Is there any special reason all packages don't use the same config?

1. `target` (all our packages are supposed to run in node and in browsers, not only ens-utils)
2. `skipNodeModulesBundle`

Additionally, I see we have `external` set to `["node_modules"]` here. I assumed we wouldn't need to specify anything explicit as `external`? As I understand any dependencies in package.json are automatically excluded? https://tsup.egoist.dev/#excluding-packages

packages/ens-utils/tsup.config.ts
```diff
@@ -2,9 +2,10 @@ import { defineConfig } from "tsup";
 
 export default defineConfig({
```
lightwalker-eth, line 3: Would it make sense to create another internal package that effectively provides a common config that all packages and apps can then make use of?

internal/seo/tsconfig.json
```diff
@@ -1,4 +1,6 @@
 {
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "extends": "@namekit/tsconfig/base.json",
   "compilerOptions": {
```
lightwalker-eth, line 4: I see some of the `compilerOptions` we are setting explicitly here are already defined in the tsconfig that we are extending. Suggest that we remove any redundant definitions so that it's more clear exactly what we're trying to override from the default that we're extending.

internal/tsconfig/nextjs.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "display": "Next.js",
+  "extends": "./base.json",
+  "compilerOptions": {
```
lightwalker-eth, line 5: I see some of the `compilerOptions` we are setting explicitly here are already defined in the tsconfig that we are extending. Suggest that we remove any redundant definitions so that it's more clear exactly what we're trying to override from the default that we're extending.

internal/tsconfig/base.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "display": "Default",
```
lightwalker-eth, line 3: Should it be like this? Appreciate your advice. Cautious about calling this "Default" as it serves as a base rather than a default.

```suggestion
  "display": "Base",
```

internal/tsconfig/nextjs.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "display": "Next.js",
```
lightwalker-eth, line 3: Could you update all our Next.js apps to use this tsconfig?

lightwalker-eth (changes requested):
@notrab Hey super happy to see this 🙌 Reviewed and shared feedback

lightwalker-eth (approved):
@notrab Thanks for your updates here. Approving this PR and created the following story to track next steps: https://app.shortcut.com/ps-web3/story/25660

PR #381, opened 2024-08-27
feat: [SC-25261] create namekit-react TruncatedText

FrancoAguzzi:


packages/namekit-react/src/components/TruncatedText.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { Tooltip } from "./Tooltip";
+import cc from "classcat";
+
+interface TruncatedTextProps {
+  text: string;
+  maxTooltipWidth?: number;
+  maxDisplayWidth?: number;
+  textStylingClasses?: string;
+  tooltipTextStylingClasses?: string;
+  displayTooltipWhenTextOverflows?: boolean;
```
notrab, line 28: I would consider adding two additional props:

```ts
renderTooltip,
tooltipProps = {},
```

This makes the component more composable that allows us to pass a custom tooltip structure if necessary. Think instances of Popover card vs Tooltip.

packages/namekit-react/src/components/TruncatedText.tsx
```diff
@@ -0,0 +1,127 @@
...
+const DEFAULT_MAX_DISPLAY_WIDTH = 300;
+const DEFAULT_MAX_TOOLTIP_WIDTH = 400;
+
+export const TruncatedText = ({
+  text,
+  textStylingClasses = "",
+  tooltipTextStylingClasses = "",
+  displayTooltipWhenTextOverflows = true,
+  maxTooltipWidth = DEFAULT_MAX_TOOLTIP_WIDTH,
+  maxDisplayWidth = DEFAULT_MAX_DISPLAY_WIDTH,
```
notrab, line 40: I would consider adding two additional props:

```ts
renderTooltip,
tooltipProps = {},
```

This makes the component more composable that allows us to pass a custom tooltip structure if necessary. Think instances of Popover card vs Tooltip.

packages/namekit-react/src/components/TruncatedText.tsx
```diff
@@ -0,0 +1,127 @@
...
+}: TruncatedTextProps) => {
+  const invisibleTextWidthTester = useRef<null | HTMLParagraphElement>(null);
+
+  /* 
+    Below state is only true if the text displayed 
+    is longer than text maxDisplayWidth
+  */
+  const [textOverflows, setTextOverflows] = useState<boolean>(false);
+
+  useEffect(() => {
```
notrab, line 50: I'm pretty sure we could modify this `useEffect` so that it only runs when necessary, and uses lists the dependencies. Something like this?

```ts
useEffect(() => {
    if (displayTooltipWhenTextOverflows && invisibleTextWidthTester.current) {
      const textOverflows =
        Math.ceil(invisibleTextWidthTester.current.getBoundingClientRect().width) > maxDisplayWidth;
      setTextOverflows(textOverflows);
    }
  }, [displayTooltipWhenTextOverflows, maxDisplayWidth, text]);
```

packages/namekit-react/src/components/TruncatedText.tsx
```diff
@@ -0,0 +1,127 @@
...
+      const textOverflows =
+        Math.ceil(
+          invisibleTextWidthTester.current.getBoundingClientRect().width,
+        ) > maxDisplayWidth;
+
+      setTextOverflows(textOverflows);
+    }
+  }, [invisibleTextWidthTester]);
+
+  const getTextElm = (
```
notrab, line 65: I would consider replacing this with a simpler `renderText` method:

```ts
const renderText = (classes: string, maxWidth = maxDisplayWidth) => (
    <p style={{ maxWidth }} className={classes}>
      {text}
    </p>
  );
```

packages/namekit-react/src/components/TruncatedText.tsx
```diff
@@ -0,0 +1,127 @@
...
+  ): JSX.Element => {
+    return (
+      <p style={{ maxWidth: maxWidth }} className={classes}>
+        {text}
+      </p>
+    );
+  };
+
+  const textDefaultClasses = "nk-text-black nk-text-sm nk-truncate";
+
```
notrab, line 77: I would add something like this for managing the tooltip:

```ts
const tooltipContent = renderTooltip ? renderTooltip(text) : renderText(
    cc(["nk-text-white nk-text-sm nk-leading-5 nk-break-all", tooltipTextStylingClasses]),
    maxTooltipWidth
  );
```

packages/namekit-react/src/components/TruncatedText.tsx
```diff
@@ -0,0 +1,127 @@
...
+  return (
```
notrab, line 78: Taking into consideration the other suggestions (based on composability), I would potentially replace the render with:

```ts
 <div>
      {displayTooltipWhenTextOverflows && (
        <div className="nk-invisible nk-absolute nk-left-0 nk-top-0 nk-pointer-events-none">
          <div ref={invisibleTextWidthTester}>{text}</div>
        </div>
      )}
      {textOverflows && displayTooltipWhenTextOverflows ? (
        <Tooltip
          trigger={renderText(cc([textStylingClasses, textDefaultClasses]))}
          {...tooltipProps}
        >
          {tooltipContent}
        </Tooltip>
      ) : (
        renderText(cc([textStylingClasses, textDefaultClasses]))
      )}
    </div>
```

notrab (changes requested):
Nice work on this @FrancoAguzzi!

I left a few suggestions that @lightwalker-eth might want to chip in on as well.

The idea around my suggestions were based on composability goals, and future use of `TruncatedText`.

apps/storybook.namekit.io/stories/Namekit/TruncatedText.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+const meta: Meta<typeof TruncatedText> = {
+  component: TruncatedText,
+  title: "Namekit/TruncatedText",
+  args: {
+    text: "",
+    textStylingClasses: "",
+    tooltipTextStylingClasses: "",
+    displayTooltipWhenTextOverflows: true,
+    maxDisplayWidth: undefined,
```
lightwalker-eth, line 26: @FrancoAguzzi Let's please ensure we never create storybook stories where one of the controls is undefined.

CleanShot 2024-08-28 at 11 10 20

packages/namekit-react/src/components/TruncatedText.tsx
lightwalker-eth: @notrab Interesting.

Am I correct to assume that if `renderTooltip` prop is undefined (the default) then the `TruncatedText` UI component would then just use the default `Tooltip` UI component?

If so, could you please create a new PR for Franco's PR 381 that would implement your suggestions? Please also include:
1. Comments on the newly added props describing how they work / rules for how they work.
2. Storybook stories with examples.

Alternatively, if the scope for completing the above isn't sufficiently small to action right now, then can you please create a new Shortcut story for us to track these ideas and then we can come back to them another time?

packages/namekit-react/src/components/TruncatedText.tsx
lightwalker-eth: @notrab Thanks for feedback. Shared a related reply in another thread here: https://github.com/namehash/namekit/pull/381/files#r1733631164

packages/namekit-react/src/components/TruncatedText.tsx
lightwalker-eth: @notrab Thanks for suggestion :+1:

@FrancoAguzzi What do you think? Please proceed with updating this as you think is best.

packages/namekit-react/src/components/TruncatedText.tsx
lightwalker-eth: @notrab Thanks for suggestion. It's not clear for me how the above is more simple? Can you please share more details?

The above suggestion seems the same to me with exceptions that it:
1. Doesn't explicitly define the return type (I like to always make this explicit).
2. Doesn't explicitly include a return statement.
3. Renames the function.

I'm not sure what else has changed?

packages/namekit-react/src/components/TruncatedText.tsx
lightwalker-eth: @notrab Thanks for suggestion :+1:

@FrancoAguzzi What do you think? Please share your suggested code updates with us.

packages/namekit-react/src/components/TruncatedText.tsx
lightwalker-eth: @notrab Thanks for suggestions :+1:

@FrancoAguzzi Can you please let us know your suggested code updates based on Jamie's feedback above?

apps/storybook.namekit.io/stories/Namekit/TruncatedText.stories.tsx
lightwalker-eth: @FrancoAguzzi I shared a suggestion somewhere else previously (not sure where) that the `Tooltip` UI component should not have a default max width of infinity but instead something that was a reasonable default for lots of cases. Ex: Maybe somewhere between 200 or 400px.

We want to have reasonable defaults so that people only need to override when its a special situation.

There's almost no situation where someone would actually want an infinite width tooltip.

packages/namekit-react/src/components/TruncatedText.tsx
```diff
@@ -0,0 +1,127 @@
...
+ */
+
+import React, { useEffect, useState, useRef } from "react";
+import { Tooltip } from "./Tooltip";
+import cc from "classcat";
+
+interface TruncatedTextProps {
+  text: string;
+  maxTooltipWidth?: number;
+  maxDisplayWidth?: number;
```
lightwalker-eth, line 25: @FrancoAguzzi I don't believe the `maxDisplayWidth` prop should be optional like this. If someone wants a `TruncatedText` they should be forced to always provide both the `text` and the `maxDisplayWidth` explicitly.

packages/namekit-react/src/components/TruncatedText.tsx
lightwalker-eth: @notrab @FrancoAguzzi If we made `TruncatedText` more composable so that a custom `Tooltip` (or other component could be passed in) then would it also make sense to remove the following props from `TruncatedText`?

1. `maxTooltipWidth`
2. `tooltipTextStylingClasses`

If someone wanted to customize these then it seems they should just make use of the new `renderTooltip` prop to achieve their goals instead?

packages/namekit-react/src/components/TruncatedText.tsx
```diff
@@ -0,0 +1,127 @@
...
+  textStylingClasses?: string;
+  tooltipTextStylingClasses?: string;
+  displayTooltipWhenTextOverflows?: boolean;
+}
+
+const DEFAULT_MAX_DISPLAY_WIDTH = 300;
+const DEFAULT_MAX_TOOLTIP_WIDTH = 400;
```
lightwalker-eth, line 32: @FrancoAguzzi The idea of a default max tooltip width should move into `Tooltip` and not be here in `TruncatedText`.

packages/namekit-react/src/components/TruncatedText.tsx
```diff
@@ -0,0 +1,127 @@
...
+interface TruncatedTextProps {
+  text: string;
+  maxTooltipWidth?: number;
+  maxDisplayWidth?: number;
+  textStylingClasses?: string;
+  tooltipTextStylingClasses?: string;
+  displayTooltipWhenTextOverflows?: boolean;
+}
+
+const DEFAULT_MAX_DISPLAY_WIDTH = 300;
```
lightwalker-eth, line 31: @FrancoAguzzi I believe we should remove this default max display width and instead require a max display width to always be explicitly provided to a `TruncatedText`.

lightwalker-eth (changes requested):
@FrancoAguzzi @notrab Thanks for updates and comments :+1: Reviewed and shared feedback :+1:

packages/namekit-react/src/components/TruncatedText.tsx
FrancoAguzzi: Concisely written, @lightwalker-eth 

And great ideation for lib's value increasing, @notrab

packages/namekit-react/src/components/TruncatedText.tsx
FrancoAguzzi: @notrab I tested the approach suggest with both normal and Slow 3G (which results in different DOM / JS bundle loading times) and it works just fine as the other approach.  

I suggest the mentioned props are added to `useEffect`'s deps array without the removal of `invisibleTextWidthTester` present in `useEffect`'s dependency array:

All these props affect the component mounting: if text, maxDisplayWidth, displayTooltipWhenTextOverflows or invisibleTextWidthTester are reset / remounted at any time, these affect the DOM element of our `TruncatedText`

`invisibleTextWidthTester` is yet important because depending on the custom classes added to the `TruncatedText`, this element **will or not** reach the `maxDisplayWidth`, needing the maths discussed, then.

Your comment was yet constructive for another piece of this Ui component: we were not styling `invisibleTextWidthTester` DOM element with user preferences for `textStylingClasses`! 

_Quick example on why this is important to be done and code was updated:_

> A user defines `textStylingClasses` as `text-[450px]` and `maxDisplayWidth` as 50px for `text` "easy".
> At first, for human eyes, we shall think that "easy" is pretty small word and `50px` are enough for displaying the full text. At a second level thinking layer, the machine might calculate that 450px fontSize is pretty big and 50px are enough for "ea", only! If `displayTooltipWhenTextOverflows` is true, this `TruncatedText` should, then, display a tooltip with the full "easy" text.

With the update, `invisibleTextWidthTester` is now styled as the real `text` will be so we have real security that the tooltip will be displayed correctly if `displayTooltipWhenTextOverflows` is true.

Thank you for your suggestion, @notrab, it was merged with current approach and provided an error case antecipated handling ✅

packages/namekit-react/src/components/TruncatedText.tsx
FrancoAguzzi: I liked the idea, @notrab !

@lightwalker-eth , I have updated the code with similar abstraction 🔑

take a look and let me know what you think

packages/namekit-react/src/components/TruncatedText.tsx
FrancoAguzzi: Hi @notrab ! I have created one `renderText` function that avoids the needed param as all places it is used need the exactly same array of classes! Please review and let me know if you liked it or not!

packages/namekit-react/src/components/TruncatedText.tsx
FrancoAguzzi: Noted and updated!

packages/namekit-react/src/components/TruncatedText.tsx
FrancoAguzzi: I am awaiting for idea alignment before implementing this 😉

apps/storybook.namekit.io/stories/Namekit/TruncatedText.stories.tsx
```diff
@@ -0,0 +1,130 @@
...
+    textStylingClasses: "ens-webfont custom-class-name huge-font-size",
+    text: getExampleTruncatedText(TruncatedTextExample.LONG_TEXT),
+  },
+};
+export const DefaultTooltipTextStyles: Story = {
+  args: {
+    text: getExampleTruncatedText(TruncatedTextExample.LONG_TEXT),
+  },
+};
+export const SmallMaxTooltipWidth: Story = {
```
lightwalker-eth, line 100: Stories like `SmallMaxTooltipWidth` and `BigMaxTooltipWidth` do not belong in `TruncatedText`. These ideas should be demonstrated in storybook stories for `Tooltip`.

If we have to repeat these stories again in `TruncatedText` then it shows we haven't properly decomposed our UI components. I hope it is clear why this is true. Please ask me if any questions.

packages/namekit-react/src/components/TruncatedText.tsx
lightwalker-eth: @FrancoAguzzi I'd like to see this action taken. It's very important to make these UI components composable. For me its a sign of bad design to see props like `maxTooltipWidth` or `tooltipTextStylingClasses` on `TruncatedText`. It's a sign that we failed to make our components composable.

apps/storybook.namekit.io/global.css
```diff
@@ -2,3 +2,12 @@
   color: red;
   background: white;
 }
+
+.huge-font-size {
+  font-size: 80px;
+}
+
+.beautiful-text {
```
lightwalker-eth, line 10:
```suggestion
.colorful-text {
```

Better to make this more specific.

packages/namekit-react/src/components/TruncatedText.tsx
```diff
@@ -21,23 +21,20 @@ import cc from "classcat";
 
 interface TruncatedTextProps {
   text: string;
-  maxTooltipWidth?: number;
-  maxDisplayWidth?: number;
   textStylingClasses?: string;
   tooltipTextStylingClasses?: string;
   displayTooltipWhenTextOverflows?: boolean;
+  maxTooltipWidth?: number;
+  maxDisplayWidth: number;
```
lightwalker-eth, line 28: Required props should always be defined first before optional props

lightwalker-eth (changes requested):

packages/namekit-react/src/components/TruncatedText.tsx
FrancoAguzzi: @lightwalker-eth happy to bootstrap this with few specs. If any adjustments are needed, let's discuss this together.

**TLDR;**
I suggest a very thin `TruncatedText` with `text`, `textStylingClasses`, `maxDisplayWidth` and `onTextOverflow` props.
I suggest a new component called `TextWithTooltipForOverflows` that integrates `TruncatedText` and `Tooltip`.
I suggest we export `TruncatedText` with the new `onTextOverflow` prop so integrators get to know when to display their custom tooltips and do this within their code.
Why is that? Triggering an _external_ tooltip and managing its props in relation to our `TruncatedText` props can be way more complex! Tooltip positioning will probably be the main challenge we might face. 

I started this component refinement and already found 2 challenges:
1. Tooltip can receive props which will be passed to the Tooltip itself. Text inside tooltip is by default set as white. Tooltip popover and tooltip arrow down are by default black. Giving users the ability to render their own tooltip is basically saying that TruncatedText does not support a tooltip internally, you should create your own tooltip and pass it to us... But, this pretty much breaks the idea of building a component that if necessary puts a tooltip in the surroundings of text. Why? Well, this is very simple. How can we determine the Tooltip trigger? How can we determine the tooltip trigger text? How can we determine the transition effect of the tooltip and **most important:** how can we determine the classes that style the tooltip and truncate it? Giving users full ability of determining their tooltips is good for _composability_? Might be. What I am saying is that it is bad for our component work. Let's take this as example: TruncatedText can receive custom tooltip JSX.Element. Whenever textOverflows is true, we display the `renderTooltip` passed to this component. This would mean:

```
{textOverflows ? (
   renderTooltip()
) : (
   renderText()
)}
``` 

So far so good. Let's now, compare the code share with the code present in the file tagged with this thread.

- The current `TruncatedText` uses renderText() to render the `Tooltip` trigger, including, then, `[textStylingClasses, textDefaultClasses]` in the classes of the informed `text` (as well as `maxDisplayWidth`)
- The shared code would leave up to people to render and customize the specs of the `Tooltip` of theirs. How can we guarantee they will truncate the trigger element? How can we force `maxDisplayWidth` of their trigger element? How can we guarantee all the specs that makes our TruncatedText, more than a simple paragraph element? --- We can surely guarantee we will render their JSX.Element based on their `renderTooltip` function. We can surely guarantee whatever they pass to us as `thisProps` or `thereProps` will be applied `this` and `there`. We cannot, though, guarantee the trigger element of their tooltip will be truncated as we want since whenever we say "Tooltip" we say "attached to", which means: how can we ensure integrators tooltip will be attached to our trigger element? The only option I see is by creating A LOT of complexity around how to specify `TruncatedText`. The idea of creating one component is to make it easy for integrators to reuse it and, after doing local tests, I share my opinion that we might be wanting to go to far with composability in this component. IMO, what would make an ideal TruncatedText with the thinest layer possible would be to remove our Tooltip component from it and add a `onTextOverflows` callback which is triggered whenever the informed text, with the informed CSS classes, with the informed maxDisplayWidth: overflows. This way we can create a component that reuses TruncatedText and our Tooltip (setting the tooltip whenever `onTextOverflows` is called) and leave up to the users the ability of building the Tooltip as they want to.

2. The problem I see is that we do not want our tooltip declared internally in the component! The goal I see is that we want integrators to have full control over the Tooltip to be used! The resolution I see is shared above.

apps/storybook.namekit.io/stories/Namekit/TruncatedText.stories.tsx
FrancoAguzzi: As aligned, these select options where replaced by a text field 🧑🏼‍💻👍🏼

packages/namekit-react/src/components/TruncatedText.tsx
FrancoAguzzi: Hi!

Happy to share I have created this story to keep the progress on this thread's subject:

[TruncatedText with custom integrators Tooltip?](https://app.shortcut.com/ps-web3/story/25710/truncatedtext-with-custom-integrators-tooltip)

packages/namekit-react/src/components/TruncatedText.tsx
FrancoAguzzi: Hi!

Happy to share I have created this story to keep the progress on this thread's subject:

[TruncatedText with custom integrators Tooltip?](https://app.shortcut.com/ps-web3/story/25710/truncatedtext-with-custom-integrators-tooltip)

lightwalker-eth (changes requested):

lightwalker-eth (approved):
@FrancoAguzzi Looks good 🚀 Thank you!

PR #387, opened 2024-09-02
Add a `computeNameGuardReport` parameter to the NameGuard API

djstrong:


packages/nameguard-sdk/src/index.ts
```diff
@@ -682,6 +682,8 @@ export class NameGuard {
    *
    * Returns `display_name` to be shown to users and estimates `impersonation_status`
    *
+   * If `options.computeNameGuardReport` is `False`, the `nameguard_result` field will be `null`.
```
lightwalker-eth, line 685: @djstrong I think it's better to invert the way this idea is documented. This way we also (implicitly) define what happens if this param is undefined. For example:

```suggestion
   * If `address` has a primary name and `options.computeNameGuardReport` is `True`, then NameGuard will attempt to inspect that primary name. Else, NameGuard will not perform any primary name inspection and the returned  `nameguard_result` field  will be `null`.
```

packages/nameguard-sdk/src/index.ts
```diff
@@ -702,13 +704,9 @@ export class NameGuard {
 
     // TODO: We need to add a `computeNameGuardReport` parameter to the API.
```
lightwalker-eth, line 705: @djstrong Seems ok for us to remove this comment now :+1:

api/nameguard/web_api.py
```diff
@@ -497,14 +499,16 @@ async def secure_primary_name_get(address: str, network_name: NetworkName) -> Se
     2. For ENS names using CCIP-Read: requests to externally defined gateway servers.
 
     Returns `display_name` to be shown to users and estimates `impersonation_status`.
+
+    If `return_nameguard_report` is `False`, the `nameguard_result` field will be `None`.
```
lightwalker-eth, line 503: Please see my other comment about how we document this field in the TS library and apply similar ideas here. Thanks.

api/nameguard/nameguard.py
```diff
@@ -440,7 +440,9 @@ def _inspect_confusable(self, grapheme: InspectorConfusableGraphemeResult) -> Co
             is_canonical=False,
         )
 
-    async def secure_primary_name(self, address: str, network_name: str) -> SecurePrimaryNameResult:
+    async def secure_primary_name(
+        self, address: str, network_name: str, return_nameguard_report: bool = True
```
lightwalker-eth, line 444: Ah interesting, here we default to `True` but in the SDK we default to false? How do you suggest we align this?

api/nameguard/models/result.py
```diff
@@ -386,7 +386,7 @@ class SecurePrimaryNameResult(BaseModel):
 
     nameguard_result: Optional[NameGuardReport] = Field(
         description='NameGuard report for the `primary_name`.\n'
-        '* `null` if `primary_name_status` is `no_primary_name` (primary name is not found)'
+        '* `null` if `primary_name_status` is `no_primary_name` (primary name is not found) or `computeNameGuardReport` is `false`',
```
lightwalker-eth, line 389: Please see my related feedback about how we document this idea when I made a comment on the TS library.

packages/nameguard-sdk/src/index.test.ts
```diff
@@ -64,6 +64,15 @@ describe("NameGuard", () => {
     expect(data.display_name).toBe("nick.eth");
   });
 
+  it("should analyze a primary name and not include the nameguard_result", async () => {
+    const data = await nameguard.getSecurePrimaryName(
+      "0xb8c2C29ee19D8307cb7255e1Cd9CbDE883A267d5",
+      { computeNameGuardReport: false },
+    );
+
+    expect(data.display_name).toBe("nick.eth");
```
lightwalker-eth, line 73: We should add an additional `expect` here to verify no nameguard_result provided.

lightwalker-eth (changes requested):
@djstrong Hey thanks for these updates. Reviewed and shared feedback :+1:

api/nameguard/models/result.py
djstrong: Other fields are defined this way so it will be less consistent.

djstrong:
The test is not passing because it runs against production version of API.

djstrong:
@lightwalker-eth @notrab Here we have some issue with CICD. This PR is not passing tests but GitHub claims it is (because not all tests have run on last commit).

packages/nameguard-sdk/src/index.ts
```diff
@@ -689,6 +689,8 @@ export class NameGuard {
    *
    * Returns `display_name` to be shown to users and estimates `impersonation_status`
    *
+   * If `address` has a primary name and `options.returnNameGuardReport` is `True`, then NameGuard will attempt to inspect that primary name. Else, NameGuard will not perform any primary name inspection and the returned `nameguard_result` field will be `null`.
```
lightwalker-eth, line 692:
```suggestion
   * If `address` has a primary name and `options.returnNameGuardReport` is `true`, then NameGuard will attempt to inspect that primary name. Else, NameGuard will not perform any primary name inspection and the returned `nameguard_result` field will be `null`.
```

In JS / TS we write boolean constants in all lowercase.

packages/nameguard-sdk/src/index.ts
```diff
@@ -457,7 +457,7 @@ export interface SecurePrimaryNameResult {
    * NameGuard report for the `primary_name`.
    *
    * * `null` if `primary_name_status` is `no_primary_name` (primary name is not found)
-   * * `null` if `SecurePrimaryNameOptions.computeNameGuardReport` is `false` or not provided
+   * * `null` if `SecurePrimaryNameOptions.returnNameGuardReport` is `false` or not provided
    */
   nameguard_result: NameGuardReport | null;
```
lightwalker-eth, line 462: It seems we should rename something here.

In some cases we have the param `returnNameGuardReport` that uses the phrasing "...report".
In other cases (such as here) we have the param `nameguard_result` that uses the phrasing "...result".

It seems to me that we should either use "...report" or "...result" everywhere and not use both.

My preference is to use "...report" everywhere as that's a bit more specific. What do you think?

lightwalker-eth (changes requested):
@djstrong Thanks for updates. Reviewed and shared feedback :+1:

packages/nameguard-sdk/src/index.ts
Carbon225: @lightwalker-eth @djstrong  I tried refactoring all `result`s to `report`s. There are so many places this needs to be changed that I think it requires a new PR and a careful review. I could just ctrl+f replace everything but that seems kind of wrong.

djstrong: OK!

PR #389, opened 2024-09-08
feat(nameguard-react): add unicode version to grapheme modal

notrab:
CleanShot 2024-09-08 at 18 44 03@2x


lightwalker-eth (approved):
@notrab looks good :+1:

PR #391, opened 2024-09-09
fix infinite loading on nameguard homepage

notrab:
I noticed the following error on the homepage causing some infinite loading and blocking IO:

```
Error: async/await is not yet supported in Client Components, only Server Components. This error is often caused by accidentally adding `'use client'` to a module that was originally written for the server.
```


lightwalker-eth (approved):
@notrab looks good :+1:

PR #393, opened 2024-09-10
Eduardoramme/sc 25626/input component

eduramme:


notrab (approved):
Looks good!

apps/nameguard.io/src/components/organisms/ContactUsForm.tsx
```diff
@@ -274,30 +253,19 @@ export const ContactUsForm = ({ title }: ContactUsFormProps) => {
...
-                    },
-                  ])}
-                />
-                {validationErrors[FormFields.Telegram] && (
-                  <span className="mt-2 text-sm font-normal text-red-600">
-                    {validationErrors[FormFields.Telegram]}
-                  </span>
-                )}
-              </div>
+              <Input
```
lightwalker-eth, line 256: @eduramme Hey can you please investigate the alignments here? We might need your help to make some adjustments to the `Input` component to fix this.

Here's a screenshot from this form field as it currently exists on nameguard.io:

CleanShot 2024-09-11 at 14 59 35

As you can see, the vertical alignment is centered and consistent for both the slot value "@" and the value being typed into the field "abc". Everything here looks great.

However, here's a screenshot of the current preview environment:

CleanShot 2024-09-11 at 15 03 01

It seems that the "@" symbol is not vertically centered perfectly.

The vertical alignment is also inconsistent between the "@" symbol and "abc".

apps/nameguard.io/src/components/organisms/ContactUsForm.tsx
```diff
@@ -274,30 +253,19 @@ export const ContactUsForm = ({ title }: ContactUsFormProps) => {
...
+                autoComplete="off"
+                disabled={isLoading}
+                onChange={handleInputChange}
+                name={FormFields.Telegram}
+                error={validationErrors[FormFields.Telegram]}
+              >
+                <Input.Slot>@</Input.Slot>
+              </Input>
             </div>
             <div className="gap-1">
```
lightwalker-eth, line 270: @eduramme It seems the input styling for the form fields is not consistent. For example:

CleanShot 2024-09-11 at 15 06 27

The text in blue has a different vertical alignment than the top 2 fields.

The text in red is styled differently than the other fields.

Please ensure the text styling on all these fields is always exactly the same.

I suppose part of this is that we need to update the `<textarea ...` below this comment to become a `<TextArea` as defined in our UI library. Please include the goal of updating all text areas across our monorepo as part of this PR. Otherwise we will have inconsistent UIs as we can see from the screenshot above.

Thanks

apps/nameguard.io/src/components/organisms/ContactUsForm.tsx
```diff
@@ -274,30 +253,19 @@ export const ContactUsForm = ({ title }: ContactUsFormProps) => {
...
-                  <span className="mt-2 text-sm font-normal text-red-600">
-                    {validationErrors[FormFields.Telegram]}
-                  </span>
-                )}
-              </div>
+              <Input
+                type="text"
+                id="telegram"
+                slotPosition="left"
+                slot="@"
```
lightwalker-eth, line 260: Is it right that we pass this `slot` param here? I thought we were supposed to only use the nested child `<Input.Slot>...` for that?

Appreciate your advice :+1:

apps/nameguard.io/src/components/organisms/ContactUsForm.tsx
```diff
@@ -274,30 +253,19 @@ export const ContactUsForm = ({ title }: ContactUsFormProps) => {
...
+                autoComplete="off"
+                disabled={isLoading}
```
lightwalker-eth, line 262: The disabled state when loading isn't working for me in the preview environment.

It might be that we need to add support for this case and define the styling for it inside the <Input> component.

Appreciate your advice and suggestions. Thank you :+1:

apps/nameguard.io/tailwind.config.ts
```diff
@@ -56,7 +56,7 @@ const config: Config = {
       },
     },
   },
-  plugins: [require("@tailwindcss/forms"), require("tailwind-scrollbar-hide")],
```
lightwalker-eth, line 59: I'm not sure if we should be removing this yet.

I see that the styling for these checkboxes is broken now:

CleanShot 2024-09-11 at 15 18 53

I want us to create standardized checkbox UI components soon, just as we've done for `Button` and now `Input` but we better do that later in a separate PR.

apps/namehashlabs.org/components/2 - molecules/contact-us-form.tsx
lightwalker-eth, line 1: It's a bummer that we have all this duplicated code in both namehashlabs.org and nameguard.io.

What do you think about moving the contact form and related logic just into https://github.com/namehash/namekit/tree/main/internal similar to what we've done with https://github.com/namehash/namekit/tree/main/internal/seo ?

For example, what if you defined a new package for `/internal/contact-form`? If so, suggest copying the existing configs, build scripts, etc.. from the existing SEO package.

If we defined that package in just one place then in theory we could import it into both projects?

Appreciate your advice :+1:

packages/namekit-react/src/components/Input.tsx
```diff
@@ -60,26 +62,34 @@ export const Input: React.FC<InputProps> & { Slot: React.FC<SlotProps> } = ({
...
+          </label>
+        )}
+        <input id={inputId} className={inputClasses} {...props} />
+        {hasSlot && slotPosition === "right" && (
+          <label htmlFor={inputId} className={slotClassName}>
+            {children}
+          </label>
+        )}
+      </div>
+      <span className="nk-mt-2 nk-text-sm nk-font-normal nk-text-red-600">
```
lightwalker-eth, line 90: Maybe not relevant, since an empty span is fine as I understand it? But perhaps we first check if `error` is defined before we emit this span and its related nested logic?

apps/storybook.namekit.io/stories/Namekit/Input.stories.tsx
```diff
@@ -112,6 +112,17 @@ export const WithSlotRight: Story = {
...
 };
 
+export const WithError: Story = {
+  args: {
+    variant: "primary",
+    inputSize: "medium",
+    placeholder: "Say something",
+    slotPosition: "right",
+    error: "This is an error",
+    children: <Input.Slot>@</Input.Slot>,
```
lightwalker-eth, line 122: Let's remove any use of slots from this story. It's best to make our examples more "pure" to show just 1 key idea at a time.

apps/storybook.namekit.io/stories/Namekit/Input.stories.tsx
```diff
@@ -112,6 +112,17 @@ export const WithSlotRight: Story = {
   },
 };
 
+export const WithError: Story = {
+  args: {
+    variant: "primary",
+    inputSize: "medium",
+    placeholder: "Say something",
+    slotPosition: "right",
```
lightwalker-eth, line 120: Let's remove any use of slots from this story. It's best to make our examples more "pure" to show just 1 key idea at a time.

apps/storybook.namekit.io/stories/Namekit/Input.stories.tsx
```diff
@@ -112,6 +112,17 @@ export const WithSlotRight: Story = {
   },
 };
 
+export const WithError: Story = {
+  args: {
+    variant: "primary",
+    inputSize: "medium",
+    placeholder: "Say something",
```
lightwalker-eth, line 119:
```suggestion
    placeholder: "Email address",
```

One of our goals for these stories is to define them to be less academic and theoretical and instead to make them more practical real life situations.

apps/storybook.namekit.io/stories/Namekit/Input.stories.tsx
```diff
@@ -112,6 +112,17 @@ export const WithSlotRight: Story = {
...
+    slotPosition: "right",
+    error: "This is an error",
```
lightwalker-eth, line 121:
```suggestion
    error: "Email is required",
```

One of our goals for these stories is to define them to be less academic and theoretical and instead to make them more practical real life situations.

apps/nameguard.io/src/components/organisms/ContactUsForm.tsx
eduramme: @lightwalker-eth You're right, slot param here is not needed

packages/namekit-react/src/components/Input.tsx
eduramme: @lightwalker-eth Sounds good

apps/namehashlabs.org/components/2 - molecules/contact-us-form.tsx
eduramme: @lightwalker-eth Thanks for the suggestion. I agree that consolidating the contact form into one place would simplify things.

Before proceeding, though, what do you think about creating a single package for all internal components (like contact forms, SEO, etc.)? This could reduce the overhead of managing multiple small packages while still keeping everything organized and reusable across projects.

Let me know what you think before we move forward.

apps/nameguard.io/src/components/organisms/ContactUsForm.tsx
eduramme: @lightwalker-eth I created a disabled version, both for `Input` and `TextArea`. Let me know what you think :)

apps/nameguard.io/src/components/organisms/ContactUsForm.tsx
eduramme: @lightwalker-eth I've replaced `textarea` by `TeaxtArea` everywhere

apps/namehashlabs.org/components/2 - molecules/contact-us-form.tsx
lightwalker-eth: @eduramme Sounds good. I created a story to track this goal here: https://app.shortcut.com/ps-web3/story/25807/create-consolidated-internal-package

packages/namekit-react/src/components/Input.tsx
```diff
@@ -57,6 +52,17 @@ export const Input: React.FC<InputProps> & { Slot: React.FC<SlotProps> } = ({
...
   );
 
+  const variantClasses = {
+    primary: cc([
+      "nk-text-black nk-border-gray-300 nk-shadow-[0_1px_2px_0_rgba(0,0,0,0.05)] hover:nk-border-gray-400 has-[input:focus]:nk-border-gray-600 has-[input:focus:hover]:nk-border-gray-600",
+      disabled ? "nk-bg-gray-50" : "nk-bg-white",
+    ]),
+    secondary: [
+      "nk-border-gray-200 hover:nk-border-gray-300 has-[input:focus]:nk-border-gray-400 has-[input:focus:hover]:nk-border-gray-400",
+      disabled ? "nk-bg-gray-200" : "nk-bg-gray-100",
```
lightwalker-eth, line 62: @eduramme Suggest we make a few styling updates to the disabled variation.

For reference, here's a screenshot of the `secondary` variation:

CleanShot 2024-09-20 at 23 25 55

And here's a screenshot of the `disabled` variation:

CleanShot 2024-09-20 at 23 26 37

See the issues here?

1. The background color of the disabled case should be darker than the secondary case.
2. The font color of the disabled case should also be made darker. It shouldn't feel so "alive".
3. Perhaps the border styles should also be adjusted? Appreciate your suggestions.

Also, am I correct to understand that there's distinct disabled styling for both primary and secondary variations? If so, please create distinct stories for "Primary Disabled" and "Secondary Disabled". Thanks :+1: We always want to make it easy to visualize the distinct states of these UI components.

apps/storybook.namekit.io/stories/Namekit/Input.stories.tsx
```diff
@@ -112,22 +112,28 @@ export const WithSlotRight: Story = {
   },
 };
 
+export const WithSlotLeftSmall: Story = {
```
lightwalker-eth, line 115: Can you help me understand the goal of this story?

It seems we didn't add a slot here as expected?

CleanShot 2024-09-20 at 23 31 44

apps/storybook.namekit.io/stories/Namekit/Input.stories.tsx
lightwalker-eth, line 1: CleanShot 2024-09-20 at 23 34 27

Can you please investigate and fix this "custom class" story?

The goal for this story is to show an example of how you can customize the styling of an `Input`. For example: Making the colors blue.

Here's a few suggestions:
1. Rename this story to "CustomStyling"
2. Stop using the class name "custom-class-name". No developer looking at our stories will understand what this means when they see this value in the "Controls" UI.
3. Instead of passing a custom class name, perhaps pass raw CSS styles that don't involve tailwind at all? The benefit of this approach in showing customization is that it is "universal". Not everyone uses tailwind or has tailwind class names memorized. But everyone looking at these stories can hopefully understand pure CSS styles.

The end goal to resolve this issue is:
1. This story should show blue text in the text field.
2. Ideally the value in the field on the "Controls" UI that is making the text blue on this story can be understood by other devs.

apps/storybook.namekit.io/stories/Namekit/TextArea.stories.tsx
lightwalker-eth, line 1: Please see the issues I raised about `Input` and apply all the same ideas to `TextArea` too. Thanks.

packages/namekit-react/src/components/TextArea.tsx
lightwalker-eth, line 1: Please see the feedback I shared on the `Input` component and apply similar ideas here :+1:

apps/namehashlabs.org/components/2 - molecules/contact-us-form.tsx
```diff
@@ -272,27 +272,18 @@ export const ContactUsForm = ({ title }: ContactUsFormProps) => {
...
+                error={validationErrors[FormFields.Message]}
               />
-
-              {validationErrors[FormFields.Message] && (
-                <span className="mt-2 text-sm font-normal text-red-600">
-                  {validationErrors[FormFields.Message]}
-                </span>
-              )}
             </div>
+            {/* Hidden input to capture the source URL of the form submission.
```
lightwalker-eth, line 285: Let's please keep the code in sync between both variations of this contact form for now. I've created a separate story for us to merge this duplicated logic into a new package in a separate PR :+1:

lightwalker-eth (changes requested):
@eduramme Thanks. Reviewed and shared feedback :+1:

packages/namekit-react/src/components/Input.tsx
eduramme: @lightwalker-eth Nice! great catch.  Yes, we have two distinct disabled styles: one for primary and one for secondary. I hadn’t created a story for the secondary disabled version yet, but I’ve just added it now :)

apps/storybook.namekit.io/stories/Namekit/Input.stories.tsx
eduramme: @lightwalker-eth Thanks, fixed

apps/storybook.namekit.io/stories/Namekit/Input.stories.tsx
eduramme: @lightwalker-eth Nice! done

apps/namehashlabs.org/components/2 - molecules/contact-us-form.tsx
notrab: My general approach to this kind of stuff is wrap as much of the reusable logic into a package, and then repeat the glue in the projects.

The glue in our case could be creating a `route.ts` file that exports a handler we export from a `@internal/X` package.

apps/storybook.namekit.io/stories/Namekit/Input.stories.tsx
lightwalker-eth: @eduramme Something is going wrong here. Can you please investigate:

For some reason, when I type in 1. the value in the text field does not update as expected. It only updates when I type in 2. This does not demonstrate the proper behaviour of this UI component in Storybook.

CleanShot 2024-09-30 at 13 12 44

Please verify we can properly type in all these `Input` stories.

apps/storybook.namekit.io/stories/Namekit/Input.stories.tsx
lightwalker-eth: @eduramme I see this comment marked as resolved, but I don't see any change made to the placeholder value in any of the 3 "slot" related stories?

We really need you to take a step back and think about the big picture goals we need to achieve. If you don't do this then we get the problem where there's round after round after round of PR reviews and feedback to finally get the correct result. This is super time consuming for everyone.

What do the goals for the StoryBook include? Is the StoryBook supposed to just be academic and theoretical examples, or is it supposed to show useful cases that people actually want? If this wasn't already clear, 3 weeks ago I made the comment above with that specific instruction.

But now look at the actual output of all 3 of the "slot" related stories we have in Storybook. They are crap! No one wants to see this kind of output. It doesn't look sufficiently useful for the use cases of someone building an ENS app! It NEEDS to do this!!

You should be asking yourself:

1. What's a realistic example of a use case in an ENS app where someone needs to use "slot left" on an Input?
2. What's a realistic example of a use case in an ENS app where someone needs to use "slot right" on an Input?

For case 1: You've already implemented one yourself: A telegram handle input. Therefore you should update the placeholder text on slot left stories to make it look like a placeholder for a telegram username input.

For case 2: The first idea I think of is putting ".eth" or something like ".uni.eth" (given the way that the Uniswap app has built their integration with ENS). Then the placeholder text could be something like "Uniswap handle". This would then demonstrate a potentially realistic use case where some app that wants to take as input ENS names in a context where they only want names that are subnames of X (such as ".uni.eth") and integrates our `Input` component to help achieve that goal.

apps/storybook.namekit.io/stories/Namekit/Input.stories.tsx
```diff
@@ -129,11 +134,20 @@ export const WithError: Story = {
...
+    value: "Primary disabled input",
+    disabled: true,
+  },
+};
+
+export const SecondaryDisabled: Story = {
+  args: {
+    variant: "secondary",
+    inputSize: "medium",
+    value: "Secondary disabled input",
```
lightwalker-eth, line 150: Please see my other comments about the problems of passing a `value` like this. This is an issue because in Storybook I can use the controls to make this UI component enabled again, but then it doesn't actually take input in this story. For a quick fix, suggest switching to a placeholder value.

apps/storybook.namekit.io/stories/Namekit/TextArea.stories.tsx
```diff
@@ -92,11 +91,21 @@ export const CustomClass: Story = {
   },
 };
 
-export const Disabled: Story = {
+export const PrimaryDisabled: Story = {
   args: {
     variant: "primary",
     size: "medium",
-    value: "Disabled Textarea",
+    value: "Primary Disabled Textarea",
```
lightwalker-eth, line 98: Please see the issues I reported about using `value` like this.

apps/storybook.namekit.io/stories/Namekit/TextArea.stories.tsx
```diff
@@ -92,11 +91,21 @@ export const CustomClass: Story = {
...
+    rows: 4,
+    disabled: true,
+  },
+};
+
+export const SecondaryDisabled: Story = {
+  args: {
+    variant: "secondary",
+    size: "medium",
+    value: "Secondary Disabled Textarea",
```
lightwalker-eth, line 108: Please see the issues I reported about using `value` like this.

packages/namekit-react/src/components/TextArea.tsx
```diff
@@ -30,11 +30,11 @@ export const TextArea: React.FC<TextAreaProps> = ({
   const variantClasses = {
```
lightwalker-eth, line 30: When a `TextArea` is disabled, the borders shouldn't animate on hover.

packages/namekit-react/src/components/Input.tsx
```diff
@@ -55,11 +55,11 @@ export const Input: React.FC<InputProps> & { Slot: React.FC<SlotProps> } = ({
   const variantClasses = {
```
lightwalker-eth, line 55: When an `Input` is disabled, the borders shouldn't animate on hover.

apps/storybook.namekit.io/stories/Namekit/TextArea.stories.tsx
eduramme: @lightwalker-eth Done, I've changed that to use defaultValue instead of value

apps/storybook.namekit.io/stories/Namekit/TextArea.stories.tsx
eduramme: Fixed

packages/namekit-react/src/components/Input.tsx
eduramme: Done

apps/storybook.namekit.io/stories/Namekit/Input.stories.tsx
```diff
@@ -84,12 +84,15 @@ export const SecondaryLarge: Story = {
...
 };
 
-export const CustomClass: Story = {
+export const CustomStyling: Story = {
   args: {
     variant: "primary",
     inputSize: "medium",
-    placeholder: "Custom Class Input",
-    className: "custom-class-name",
+    value: "Custom Styled Input",
```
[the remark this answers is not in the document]
eduramme: @lightwalker-eth done, I've changed that to use "defaultValue" instead of "value"

lightwalker-eth (approved):
@eduramme Looks good! Approved! 🚀

PR #394, opened 2024-09-11
New IconButton and Button API

notrab:
## Changes

1. **`asChild` is now a `boolean` instead of `ReactNode`**

We previously used `asChild` to pass the child we wanted to render, but instead, we can use the `children` prop to merge the composed components &mdash; [similar to Radix Themes](https://www.radix-ui.com/primitives/docs/guides/composition).

This means we go from this:

```tsx
<a
  href="https://github.com/namehash/nameguard"
  target="_blank"
  rel="noopener noreferrer"
>
  <Button variant="ghost">Docs</Button>
</a>
```

To this:

```tsx
<Button variant="ghost" asChild>
  <Link variant="ghost" href="https://api.nameguard.io/docs">
    Docs
  </Link>
</Button>
```

2. **`IconButton` now renders children**

This PR also improves simplicity of `IconButton`, which previously required the following props:

* `icon`
* `iconPosition`

These were explicit properties that felt awkward using in our projects. We can instead just pass children and use this button for only buttons that are icons. 

The new proposed API goes from this:

```tsx
<a
  href="https://github.com/namehash/nameguard"
  target="_blank"
  rel="noopener noreferrer"
>
  <IconButton
    icon={
      <GithubIcon className="hidden md:block text-[#0F172A] fill-current" />
  >
    Github
  </IconButton>
</a>
```

To this:

```tsx
 <Button variant="ghost" asChild>
  <Link
    variant="ghost"
    href="https://github.com/namehash/nameguard"
  >
    <GithubIcon className="hidden md:block text-[#0F172A] fill-current" />{" "}
      GitHub
    </Link>
</Button>
```

Which now feels more composable, without the endless props of position. If you want to make something left or right, just add the icon in the children before or after the text, or other elements.

If a button has an icon and text, you use a button and pass it an icon as a child. It's a button with an icon, not an "icon button". They're two different things.

CleanShot 2024-09-12 at 17 00 35@2x

## Other things

* I've updated some instances of the previous behaviour to show the new approach, including the incorrect use of `<a>`.
* If we need to use `next/link`, we should use `next/link` with `asChild` too*

## Todos

- [ ] Make sure all instances of Button/IconButton/Link work as expected.
- [x] Remove `ghost` styling and inherit the styling of the parent. `ghost` was added to remove the underline from links that are rendered as buttons, they should be rendered as buttons and only have their styles, not the links.
- [ ] Add some additional stories (but not for `ghost`, if we don't need it)


packages/namekit-react/src/components/Link.tsx
```diff
@@ -3,7 +3,7 @@ import cc from "classcat";
 
 export interface LinkProps extends React.ComponentPropsWithoutRef<"a"> {
   className?: string;
-  variant?: "primary" | "secondary" | "underline";
+  variant?: "primary" | "secondary" | "ghost" | "underline";
```
lightwalker-eth, line 6: @notrab Whenever we make an update like this please also proactively update the storybook for the related UI component.

The storybook for both `Link` and `Button` are a mess right now too.

CleanShot 2024-09-11 at 18 07 51

For example, I can't customize the variant or the size as expected.

And what are the `rel` and `target` controls for? If these are exposed via Storybook then I should be able to figure out what these do just from the example stories without needing to read any of the code. But there aren't any stories that communicate any ideas about these props. So we should either create those stories or remove these props as controls in Storybook.

CleanShot 2024-09-11 at 18 12 15

The above is also broken. Can you please investigate why? I see that `colorful-text` works on the story for `TruncatedText`.

lightwalker-eth (changes requested):
@notrab Thanks for updates. Reviewed and shared feedback.

apps/nameguard.io/src/components/organisms/Header.tsx
```diff
@@ -1,75 +1,74 @@
...
+import { Button, IconButton, Link } from "@namehash/namekit-react";
 
 export const Header = () => {
   return (
     <header className="sticky bg-white top-0 w-full z-20 border-b border-gray-300 h-[56px] py-[9px] sm:h-[70px] sm:py-4 select-none">
       <div className="max-w-7xl mx-auto items-center justify-between flex flex-row px-6">
         <div className="flex flex-row lg:gap-2 xl:gap-7 justify-between items-center">
           <div className="flex flex-row justify-between items-center gap-1 cursor-pointer">
-            <Link
+            <NextLink
```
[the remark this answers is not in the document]
notrab: So the way `Link` was created initially was that it was just a wrapper on top of `a`, since it's not always a `NextLink` (that has additional helpers). For example, `Link` can be used inside of `nameguard-react` which isn't bound to Next.js

If we know it's an internal link, `NextLink` is a good use here because of the benefits that brings with prefetching.

packages/namekit-react/src/components/Link.tsx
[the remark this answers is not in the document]
notrab: I haven't done this yet because the intended API might not be final, wanted to wait for feedback.

packages/namekit-react/src/components/Link.tsx
notrab: @lightwalker-eth this is the first time I have seen these "controls". I had them disabled when I first implemented Storybook, so I had no idea they were even here.

apps/nameguard.io/src/components/organisms/Header.tsx
```diff
@@ -1,75 +1,74 @@
...
-            </a>
+            <Button variant="ghost" asChild={<NextLink href="/contact" />}>
+              Contact
+            </Button>
 
             <CalendarButton>Discuss an integration</CalendarButton>
           </div>
 
           <div className="lg:hidden">
-            <IconButton
```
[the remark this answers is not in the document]
notrab: @lightwalker-eth I would like to rethink `IconButton` so it's just an "Icon Button". It's not a button with text. If you want that, you use a `Button` with a Slot with the icon left or right. Similar to Radix. If you're happy with that proposal, I can make the change.

FrancoAguzzi (approved):
Since `nameguard-react` was changed it is important to add a changeset file in this PR.

Approving it since it is aligned with what we have discussed earlier.

lightwalker-eth (changes requested):
@notrab Hey, great to see and review this PR. Our UI library is coming together 😄 

Please see review comments.

I also did a search across all files for "<a " in the monorepo and found a few other UI components that should be updated to use `Button`, `Link`, or `IconButton`.

I appreciate if your suggestion is to update the following in a separate PR, but if you suggest that please proactively create a new story in Shortcut to track this goal.

- [ ] `FinancialSupportSection`
- [ ] `BrandAssets`
- [ ] `Footer` (from nameguard.io)
- [ ] `DeveloperResourceItem`
- [ ] `ChecksSection`
- [ ] `ReadySectionProps`
- [ ] `AvatarWithTooltip`
- [ ] `BrandAssets`
- [ ] `Cta` (from namehashlabs.org -- I think we aren't using this at all anymore?)
- [ ] `FinancialSupportSection`
- [ ] `TeamMemberComponent`
- [ ] `PartnerCard`
- [ ] `PartnerCardProps`
- [ ] `JoinOurTeam`
- [ ] `ExternalLinks`
- [ ] `RoadMapElement`

I also found an update needed to the following UI components, but don't worry about them for this PR because we will update / fix the following UI components in a different story (long background info...)

- [X] `ImpersonationReport`

apps/namehashlabs.org/components/2 - molecules/our-supporters-section.tsx
```diff
@@ -85,17 +85,14 @@ const AdditionalSupportersContainer = () => {
...
         href="https://snapshot.org/#/ens.eth/proposal/0x6ba81cd2997288cc49ae1b95921ec8f107e8ffb9733321d53d488e2b30710b86"
+        asChild
       >
-        <IconButton
-          iconPosition="right"
-          icon={<ExternalLinkIcon className="text-white w-5 h-5" />}
-        >
-          View more
+        <IconButton>
+          View more <Link.ExternalIcon />
```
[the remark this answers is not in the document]
notrab: Fixed.

apps/namehashlabs.org/app/ens-referral-program/page.tsx
```diff
@@ -128,16 +127,12 @@ const Item = ({
...
       </div>
       {!!buttonUrl && (
-        <a target="_blank" href={buttonUrl}>
-          <IconButton
-            iconPosition="right"
-            icon={<ExternalLinkIcon className="text-black w-5 h-5" />}
-            variant="secondary"
-            size="medium"
-          >
+        <Button variant="secondary" size="medium" asChild>
```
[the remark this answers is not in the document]
notrab: Resolved, fixed the layout issue also.

apps/namehashlabs.org/app/ens-referral-program/page.tsx
```diff
@@ -3,11 +3,10 @@ import Image from "next/image";
 import cc from "classcat";
 
 import { DocumentTextIcon } from "@heroicons/react/24/solid";
-import { ExternalLinkIcon } from "@/components/1 - atoms";
```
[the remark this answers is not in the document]
notrab: Removed and cleaned up instances.

apps/namehashlabs.org/components/2 - molecules/product-component.tsx
```diff
@@ -63,14 +63,11 @@ export const ProductComponent = ({
...
             </SectionText>
             {buttonUrl && (
-              <a href={buttonUrl} target="_blank">
-                <IconButton
-                  iconPosition="right"
-                  icon={<ExternalLinkIcon className="w-5 h-5" />}
-                >
-                  Join the discussion
+              <Link href={buttonUrl} asChild>
+                <IconButton>
```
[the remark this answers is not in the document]
notrab: Yeah, this is causing the issue of the underline.

packages/nameguard-react/src/components/ChatModal/ChatModal.tsx
```diff
@@ -32,21 +32,12 @@ export const ChatModal = forwardRef(
...
-              rel="noopener noreferrer"
-            >
-              <IconButton
-                iconPosition="right"
-                icon={
-                  <ExternalLinkIcon className="!w-5 !h-5 fill-current !text-white" />
-                }
-              >
+            <Link className="inline-flex" href={xmtpChatUrl} asChild>
+              <IconButton>
```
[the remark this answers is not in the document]
notrab: Fixed

packages/nameguard-react/src/components/Share/Share.tsx
```diff
@@ -77,13 +77,8 @@ export function Share({ name }: ShareProps) {
...
       <Tooltip
         trigger={
-          <IconButton
-            variant="ghost"
-            icon={
-              <ArrowUpTrayIcon className="text-black fill-current w-6 h-6" />
-            }
-            onClick={() => setIsOpen(true)}
-          >
+          <IconButton variant="ghost" onClick={() => setIsOpen(true)}>
```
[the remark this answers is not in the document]
notrab: Yeah, but, the icon button needs the classes changed and padding fixed. So I will need to do another round on that specific component. Doing that should also remove the need for `!p-2` that @eduramme has added in places, because IconButton will have equal padding!

apps/namehashlabs.org/components/3 - organisms/role-page.tsx
```diff
@@ -54,15 +52,13 @@ const JoinOurTeam = () => {
           </Link>
         </p>
       </div>
-      <a
+      <Link
         href="mailto:hello@namehashlabs.org"
         className="px-4 cursor-pointer flex items-center gap-3 py-2 rounded-[8px] bg-white flex-nowrap hover:bg-gray-100 transition-colors duration-200 w-full justify-center "
```
lightwalker-eth, line 57: I think we shouldn't need most / all of these classnames anymore now that this transitioned from an `<a` to one of our new `<Link`?

It's a little difficult to test this right now because we purposefully made the roles data structure empty which means you can't access this page currently.

apps/namehashlabs.org/components/2 - molecules/product-component.tsx
```diff
@@ -62,12 +62,15 @@ export const ProductComponent = ({
             <SectionText className="lg:text-start text-center w-full text-lg leading-7 font-light">
               <Balancer>{subtitle}</Balancer>
             </SectionText>
+
             {buttonUrl && (
-              <Link href={buttonUrl} asChild>
-                <IconButton>
-                  Join the discussion <Link.ExternalIcon />
+              <div className="flex justify-center lg:justify-start">
+                <IconButton asChild>
```
lightwalker-eth, line 68: Something funny is happening here. Perhaps it's because this is using `IconButton` and not `Button`?

Here's how this looks now:

CleanShot 2024-09-21 at 00 08 41

And here's how it looked before:

CleanShot 2024-09-21 at 00 09 06

apps/namehashlabs.org/components/1 - atoms/headline-banner.tsx
```diff
@@ -12,12 +12,12 @@ export const HeadlineBanner = () => {
...
         target="_blank"
-        href="https://twitter.com/NamehashLabs"
+        href="https://x.com/NamehashLabs"
         variant="underline"
         size="small"
       >
-        <p className="hidden sm:block">Follow us on Twitter</p>
-        <p className="block sm:hidden w-max">Follow us</p>
+        <span className="hidden sm:block">Follow us on Twitter</span>
+        <span className="block sm:hidden w-max">Follow us</span>
```
lightwalker-eth, line 20: When the screen width gets small enough, this "Follow Us" variation appears but it isn't getting the correct `<Link` UX.

Additionally, I don't know why, but the logic for the header isn't working correctly:

CleanShot 2024-09-21 at 00 11 02

The hamburger menu is falling off the edge of the screen.

apps/namehashlabs.org/components/2 - molecules/our-supporters-section.tsx
```diff
@@ -85,14 +79,11 @@ const AdditionalSupportersContainer = () => {
       <p className="text-lg leading-6 font-normal text-gray-500">
         and so many others
       </p>
-      <Link
-        href="https://snapshot.org/#/ens.eth/proposal/0x6ba81cd2997288cc49ae1b95921ec8f107e8ffb9733321d53d488e2b30710b86"
-        asChild
-      >
-        <IconButton>
+      <IconButton asChild>
```
lightwalker-eth, line 82: This should be a `Button`

apps/namehashlabs.org/components/2 - molecules/products-section.tsx
```diff
@@ -98,7 +97,7 @@ const Product = ({
                     target={withoutExternalLinkIconInCTA ? undefined : "_blank"}
                   >
```
lightwalker-eth, line 98: This should be a `Button` rather than an `IconButton`.

apps/namehashlabs.org/components/2 - molecules/products-section.tsx
```diff
@@ -98,7 +97,7 @@ const Product = ({
                     target={withoutExternalLinkIconInCTA ? undefined : "_blank"}
```
lightwalker-eth, line 97: Not sure if I'm leaving this comment in the right place, but can you please investigate the UX of this button:

CleanShot 2024-09-21 at 00 21 17

A few issues here:
1. I'm in the Vercel preview environment, but for some reason this link is going to the production website.
2. This link is internal and therefore should be a `NextLink`
3. This link shouldn't open in a new tab.

packages/namekit-react/src/components/IconButton.tsx
```diff
@@ -6,15 +6,28 @@ export interface IconButtonProps extends ButtonProps {
...
       className,
     ]);
 
     return (
-      <Button ref={ref} className={combinedClassName} {...props}>
+      <Button
+        ref={ref}
+        className={combinedClassName}
+        size={size}
+        padding={iconButtonSizeClasses[size]}
```
lightwalker-eth, line 28: Is it really needed to add this `padding` param? Could this same goal be achieved through the values we pass on the `className` field?

apps/nameguard.io/src/components/organisms/Header.tsx
```diff
@@ -52,7 +52,7 @@ export const Header = () => {
           </div>
 
           <div className="lg:hidden">
-            <IconButton className="!p-2" variant="ghost">
+            <IconButton variant="ghost">
               <NGSearchIcon />
```
lightwalker-eth, line 56: It looks to me like the size of the search icon button is smaller than the size of the hambuger menu icon button.

Please compare the hover states.

CleanShot 2024-09-21 at 00 27 41

Additionally, as I reduce the width of the screen, the search field gets too close to the NameGuard logo.

CleanShot 2024-09-21 at 00 29 19

lightwalker-eth (changes requested):
@notrab Thanks for updates. Reviewed and shared feedback :+1:

apps/nameguard.io/src/components/organisms/Header.tsx
notrab: Fixed

apps/namehashlabs.org/components/2 - molecules/product-component.tsx
notrab: @lightwalker-eth It seems we have different styles of the external link icon. We should settle for one, wdyt?

apps/namehashlabs.org/components/3 - organisms/role-page.tsx
notrab: Yeah, so the initiative I took here was to update it to include the correct component, but keep the style implementation, so we can remove the styles if it looked ok. I was going to ask about the role data.

I can add some dummy data to test it out regardless. Will report back!

apps/namehashlabs.org/components/1 - atoms/headline-banner.tsx
notrab: Fixed

apps/namehashlabs.org/components/2 - molecules/products-section.tsx
notrab: Fixed.

apps/namehashlabs.org/components/2 - molecules/header.tsx
```diff
@@ -9,22 +10,22 @@ export const Header = () => {
   return (
     <nav className="lg:px-[112px] px-5 w-full h-20 flex items-center justify-center border-b border-black border-opacity-10 z-20">
```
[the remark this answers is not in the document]
notrab: You're right. The `Link` component was designed to have no hover styles on mobile, because there isn't really hover state, but an active state.

I've reverted that change and applied the same style to all viewports now.

lightwalker-eth (changes requested):
@notrab Thanks for updates. Still a few items to fix. Reviewed and shared feedback :+1:

packages/namekit-react/src/components/Link.tsx
```diff
@@ -30,7 +30,17 @@ const isInternalLink = (href: string | undefined) => {
   return href.startsWith("/") || href.startsWith("#") || !href.includes("://");
```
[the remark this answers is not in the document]
notrab: I'd rather not export from that component library until I find a better home for this. Maybe it's a utility export from the package itself.

packages/namekit-react/src/components/IconButton.tsx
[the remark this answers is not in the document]
lightwalker-eth: @notrab Understand. I think no need to make the `IconButton` super smart to automatically handle all this stuff. What matters is: How does the final UX in the Storybook look on first impression?

Right now it doesn't look good for the reasons reported.

Appreciate if you can suggest how we can fix it. Ex: Maybe use smaller icons or larger icons for different size `IconButton` stories.

apps/nameguard.io/src/components/organisms/Hero.tsx
```diff
@@ -32,14 +32,11 @@ export function Hero() {
...
 
-          <Button
-            variant="primary"
-            size="large"
-            asChild
-            className="hidden lg:block relative z-10"
-          >
-            <Link href="https://api.nameguard.io/docs">View the docs</Link>
-          </Button>
+          <div className="hidden md:block relative z-10">
```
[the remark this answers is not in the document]
notrab: Fixed

lightwalker-eth (approved):
@notrab Thanks for updates! Approved 🚀

PR #396, opened 2024-09-11
feat(nameguard-sdk): throw errors from API

notrab:
Earlier this year we discussed the idea of adding different error types to the SDK for timeouts, lost connection, server errors and input validation.

This PR starts the conversation about adding such support.

Please note;
1. There is no changeset, I need to do the other end of this to update `nameguard-react`

### DIFF
--- a/packages/nameguard-sdk/src/index.ts
+++ b/packages/nameguard-sdk/src/index.ts
@@ -462,14 +462,28 @@ export interface SecurePrimaryNameResult {
   nameguard_result: NameGuardReport | null;
 }
 
-// TODO: Let's apply more formalization to this error class.
-class NameGuardError extends Error {
+export enum NameGuardErrorType {
+  Server = "Server",
+  Network = "Network",
+  Timeout = "Timeout",
+  Abort = "Abort",
+  Input = "Input",
+  Unknown = "Unknown",
+}
+
+export class NameGuardError extends Error {
   constructor(
-    public status: number,
+    public type: NameGuardErrorType,
+    public status?: number,
     message?: string,
   ) {
-    super(message);
+    super(message || `NameGuard ${type} Error`);
+    this.name = "NameGuardError";
   }
+}
+
+function isError(error: unknown): error is Error {
+  return error instanceof Error;
 }
 
 const DEFAULT_ENDPOINT = "https://api.nameguard.io";
@@ -533,13 +547,30 @@ export class NameGuard {
    * @example
    * const nameGuardReport = await nameguard.inspectName('vitalik.eth');
    */
-  public inspectName(
+  public async inspectName(
     name: string,
     options?: InspectNameOptions,
   ): Promise<NameGuardReport> {
     const network_name = this.network;
 
-    return this.rawRequest("inspect-name", "POST", { name, network_name });
+    try {
+      const network_name = this.network;
+
+      return await this.rawRequest("inspect-name", "POST", {
+        name,
+        network_name,
+      });
+    } catch (error) {
+      if (error instanceof NameGuardError) {
+        throw error;
+      }
+
+      throw new NameGuardError(
+        NameGuardErrorType.Input,
+        undefined,
+        `Invalid input: ${isError(error) ? error.message : "Unknown error"}`,
+      );
+    }
   }
 
   // TODO: Document how this API will attempt automated labelhash resolution through the ENS Subgraph.
@@ -562,22 +593,36 @@ export class NameGuard {
    * @param {InspectNameOptions} options The options for the inspection.
    * @returns {Promise<BulkConsolidatedNameGuardReport>} A promise that resolves with a list of `ConsolidatedNameGuardReport` values for each name queried in the bulk inspection. Check the `inspected` field of each report to determine if the name was fully inspected or inspected in a limited way for performance reasons.
    */
-  public bulkInspectNames(
+  public async bulkInspectNames(
     names: string[],
     options?: InspectNameOptions,
   ): Promise<BulkConsolidatedNameGuardReport> {
     if (names.length > MAX_BULK_INSPECTION_NAMES) {
-      throw new Error(
+      throw new NameGuardError(
+        NameGuardErrorType.Input,
+        undefined,
         `Bulk inspection of more than ${MAX_BULK_INSPECTION_NAMES} names at a time is not supported.`,
       );
     }
 
     const network_name = this.network;
 
-    return this.rawRequest("bulk-inspect-names", "POST", {
-      names,
-      network_name,
-    });
+    try {
+      return await this.rawRequest("bulk-inspect-names", "POST", {
+        names,
+        network_name,
+      });
+    } catch (error) {
+      if (error instanceof NameGuardError) {
+        throw error;
+      }
+
+      throw new NameGuardError(
+        NameGuardErrorType.Unknown,
+        undefined,
+        isError(error) ? error.message : "Unknown error",
+      );
+    }
   }
 
   // TODO: We need to have more specialized error handling here for cases such as the lookup in the ENS Subgraph failing.
@@ -601,26 +646,43 @@ export class NameGuard {
     options?: InspectNamehashOptions,
   ): Promise<NameGuardReport> {
     if (!isKeccak256Hash(namehash)) {
-      throw new Error("Invalid Keccak256 hash format for namehash.");
+      throw new NameGuardError(
+        NameGuardErrorType.Input,
+        undefined,
+        "Invalid Keccak256 hash format for namehash.",
+      );
     }
 
     const network = this.network;
 
-    const url = new URL(
-      `inspect-namehash/${network}/${namehash}`,
-      this.endpoint,
-    );
+    try {
+      const url = new URL(
+        `inspect-namehash/${network}/${namehash}`,
+        this.endpoint,
+      );
 
-    const response = await fetch(url);
+      const response = await fetch(url);
 
-    if (!response.ok) {
+      if (!response.ok) {
+        throw new NameGuardError(
+          NameGuardErrorType.Server,
+          response.status,
+          `Failed to inspect namehash ${namehash} using the network ${network}.`,
+        );
+      }
+
+      return await response.json();
+    } catch (error) {
+      if (error instanceof NameGuardError) {
+        throw error;
+      }
+
       throw new NameGuardError(
-        response.status,
-        `Failed to inspect namehash ${namehash} using the network ${network}.`,
+        NameGuardErrorType.Unknown,
+        undefined,
+        isError(error) ? error.message : "Unknown error",
       );
     }
-
-    return await response.json();
   }
 
   // TODO: The main purpose of this function is to pass a tokenId rather than a labelhash. However we need to make changes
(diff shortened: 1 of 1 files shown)


packages/nameguard-sdk/src/index.ts
```diff
@@ -462,16 +462,30 @@ export interface SecurePrimaryNameResult {
...
+  Abort = "Abort",
+  Input = "Input",
+  Unknown = "Unknown",
+}
+
+export class NameGuardError extends Error {
   constructor(
-    public status: number,
+    public type: NameGuardErrorType,
+    public status?: number,
```
lightwalker-eth, line 477: What's the goal for this `status` field?

packages/nameguard-sdk/src/index.ts
```diff
@@ -462,16 +462,30 @@ export interface SecurePrimaryNameResult {
   nameguard_result: NameGuardReport | null;
 }
 
-// TODO: Let's apply more formalization to this error class.
-class NameGuardError extends Error {
+export enum NameGuardErrorType {
```
lightwalker-eth, line 465: Can we please document at least 1 sentence for each of these different error types to explain in more words what they represent?

packages/nameguard-sdk/src/index.ts
```diff
@@ -528,18 +536,29 @@ export class NameGuard {
...
+        name,
+        network_name,
+      });
+    } catch (error) {
+      if (error instanceof NameGuardError) {
+        throw error;
+      }
+
+      throw new NameGuardError(
+        NameGuardErrorType.Input,
```
lightwalker-eth, line 557: This isn't necessarily an input error. Could we add logic here to classify the error type as accurately as possible?

packages/nameguard-sdk/src/index.ts
```diff
@@ -559,25 +578,37 @@ export class NameGuard {
    *
    *
    * @param {string[]} names The list of names for NameGuard to inspect.
-   * @param {InspectNameOptions} options The options for the inspection.
    * @returns {Promise<BulkConsolidatedNameGuardReport>} A promise that resolves with a list of `ConsolidatedNameGuardReport` values for each name queried in the bulk inspection. Check the `inspected` field of each report to determine if the name was fully inspected or inspected in a limited way for performance reasons.
    */
```
lightwalker-eth, line 582: Let's add a `@throws ...` clause to the documentation for each of the relevant functions that communicates the idea that these functions throw a `NameGuardError` if anything goes wrong.

packages/nameguard-sdk/src/index.ts
```diff
@@ -590,37 +621,50 @@ export class NameGuard {
...
+    } catch (error) {
+      if (error instanceof NameGuardError) {
+        throw error;
+      }
 
-    if (!response.ok) {
       throw new NameGuardError(
-        response.status,
-        `Failed to inspect namehash ${namehash} using the network ${network}.`,
+        NameGuardErrorType.Unknown,
```
lightwalker-eth, line 663: We should add handling to each of these network calls that detects specific error scenarios such as server / network / timeout / abort.

We should also look at the API docs for each NameGuard API and use the related HTTP status codes to help return more specific error types.

packages/nameguard-sdk/src/index.ts
```diff
@@ -698,7 +778,9 @@ export class NameGuard {
     options?: SecurePrimaryNameOptions,
   ): Promise<SecurePrimaryNameResult> {
```
lightwalker-eth, line 779: Please also check the `nameguard-js` package and see if there's anything there that needs to be updated. NameGuard JS implements a subclass of the `NameGuard` class.

packages/nameguard-sdk/src/index.ts
```diff
@@ -793,22 +902,47 @@ export class NameGuard {
...
-
-    return await response.json();
   }
 
   public abortAllRequests(): void {
     this.abortController.abort();
-
     this.abortController = new AbortController();
+
+    throw new NameGuardError(
```
lightwalker-eth, line 941: Not sure if this function should throw an error?

lightwalker-eth (changes requested):
@notrab Thanks for updates. Reviewed and shared feedback :+1:

packages/nameguard-sdk/src/index.ts
notrab: Good idea @lightwalker-eth, I've added this, as well as some `@example` doc blocks.

packages/nameguard-sdk/src/index.ts
notrab: he HTTP status code associated with the error, if applicable. I've updated it to include a doc block.

packages/nameguard-sdk/src/index.ts
notrab: Added a doc block for the whole class and constructor now @lightwalker-eth 👍🏻

packages/nameguard-sdk/src/index.ts
notrab: @lightwalker-eth I had this locally I didn't push, because it probably needs more thought about the errors returning from the API too so we can throw the correct errors in code.

Here's what I was thinking:

```ts
if (error instanceof TypeError && error.message.includes("Failed to fetch")) {
      throw new NameGuardError(
        NameGuardErrorType.Network,
        undefined,
        "Network error occurred while inspecting the name"
      );
    }

    if (error instanceof Error && error.message.includes("timeout")) {
      throw new NameGuardError(
        NameGuardErrorType.Timeout,
        undefined,
        "Request timed out while inspecting the name"
      );
    }

    if (error instanceof Error && error.message.toLowerCase().includes("invalid")) {
      throw new NameGuardError(
        NameGuardErrorType.Input,
        undefined,
        `Invalid input: ${error.message}`
      );
    }

    // If we can't classify it more specifically, use Unknown
    throw new NameGuardError(
      NameGuardErrorType.Unknown,
      undefined,
      `An unexpected error occurred: ${error instanceof Error ? error.message : String(error)}`
    );
  }
```

packages/nameguard-sdk/src/index.ts
notrab: What if we did this (keep in mind not everything here regarding response codes is in the openapi spec that I looked through) but we could remove any from the example below that we don't want to add (Rate limiting for example):

``ts
async rawRequest(
  path: string,
  method: string = "GET",
  body: object = {},
  headers: object = {}
): Promise<any> {
  const url = new URL(path, this.endpoint);

  const options: RequestInit = {
    method,
    headers: {
      "Content-Type": "application/json",
      ...headers,
    },
    signal: this.abortController.signal,
  };

  if (method !== "GET") {
    options.body = JSON.stringify(body);
  }

  try {
    const response = await fetch(url, options);

    if (!response.ok) {
      // Handle specific HTTP status codes
      switch (response.status) {
        case 400:
          throw new NameGuardError(
            NameGuardErrorType.Input,
            response.status,
            "Bad request: The server could not understand the request"
          );
        case 401:
          throw new NameGuardError(
            NameGuardErrorType.Authentication,
            response.status,
            "Unauthorized: Authentication is required and has failed or has not been provided"
          );
        case 403:
          throw new NameGuardError(
            NameGuardErrorType.Authorization,
            response.status,
            "Forbidden: The server understood the request but refuses to authorize it"
          );
        case 404:
          throw new NameGuardError(
            NameGuardErrorType.NotFound,
            response.status,
            "Not Found: The requested resource could not be found"
          );
        case 429:
          throw new NameGuardError(
            NameGuardErrorType.RateLimit,
            response.status,
            "Too Many Requests: You have sent too many requests in a given amount of time"
          );
        case 500:
          throw new NameGuardError(
            NameGuardErrorType.Server,
            response.status,
            "Internal Server Error: The server has encountered a situation it doesn't know how to handle"
          );
        default:
          if (response.status >= 500) {
            throw new NameGuardError(
              NameGuardErrorType.Server,
              response.status,
              `Server error: ${response.statusText}`
            );
          } else {
            throw new NameGuardError(
              NameGuardErrorType.Unknown,
              response.status,
              `Unexpected error: ${response.statusText}`
            );
          }
      }
    }

    return await response.json();
  } catch (error) {
    if (error instanceof NameGuardError) {
      throw error;
    } else if (error instanceof TypeError && error.message === "Failed to fetch") {
      throw new NameGuardError(
        NameGuardErrorType.Network,
        undefined,
        "Network error: Unable to reach the server"
      );
    } else if (error instanceof DOMException && error.name === "AbortError") {
      throw new NameGuardError(
        NameGuardErrorType.Abort,
        undefined,
        "Request was aborted"
      );
    } else {
      throw new NameGuardError(
        NameGuardErrorType.Unknown,
        undefined,
        `An unexpected error occurred: ${error instanceof Error ? error.message : String(error)}`
      );
    }
  }
}
```

File: packages/nameguard-sdk/src/index.ts (hunk 7 above)
Comment (notrab) [reply packages/nameguard-sdk/src/index.ts]: At the time of spiking the PR I thought this could be slightly helpful. But I've reverted the changes and added a new doc block.

File: packages/nameguard-sdk/src/index.ts (hunk 5 above)
Comment (lightwalker-eth) [reply packages/nameguard-sdk/src/index.ts:685]: @notrab This looks like a good direction, thank you :+1:

A few suggestions for feedback:

1. What about the timeout case? Can you please simulate a timeout and verify our logic can detect that case correctly?
2. I'm concerned about the `DOMException` referenced in the code above. Can you please suggest a path forward for us where we can be sure all the code here works equally well both in frontend contexts where it runs in the browser and in backend contexts where it runs in node?
3. Suggest we reduce the number of `NameGuardErrorType` values to only those we have real use cases for currently. For example, I understand we never have errors for `NameGuardErrorType.Authentication`, `NameGuardErrorType.Authorization`, or `NameGuardErrorType.RateLimit`. We can introduce new error codes in the future if / when we build related features for that. What do you think? A key motivation for this suggestion is that each error code case introduces the possibility of needing specialized UX handling in the frontend. Maybe I'm being too cautious about that. I'm coming at this from the perspective of trying to avoid introducing too many UI states. But perhaps that's not an issue and we can create just a single generic UI wherever error state reporting into the UI is needed.

File: packages/nameguard-sdk/src/index.ts (hunk 3 above)
Comment (lightwalker-eth) [reply packages/nameguard-sdk/src/index.ts:572]: @notrab Please see our other active comment thread on this PR. My understanding is the other approach you've suggested for inside `rawRequest` would be better than the approach suggested in the message above, and therefore that the approach suggested above could be removed.

If I'm missing something here please let me know :+1:

**Review by lightwalker-eth — COMMENTED** [review]
@notrab Hey just saw your updated comments here. Reviewed and shared feedback :+1:

File: packages/nameguard-sdk/src/index.ts [hunk 8]
```diff
@@ -790,21 +953,92 @@ export class NameGuard {
       options.body = JSON.stringify(body);
     }
 
-    const response = await fetch(url, options);
+    try {
+      const timeoutId = setTimeout(() => {
+        this.abortController.abort();
+      }, 30000); // 30 second timeout
```
notrab, line 959: We could make this configurable if we need.

I think we could attach timeout to the signal instead from looking at [this page](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static). I need to test it.

notrab: I've added this to the constructor now.

PR #399, opened 2024-09-16
feat: [sc-25622] Refine Tooltip Component

FrancoAguzzi:
Story details: https://app.shortcut.com/ps-web3/story/25622


apps/storybook.namekit.io/stories/Namekit/Tooltip.stories.tsx
```diff
@@ -45,6 +71,28 @@ export const LeftPlacement: Story = {
...
+  render: (args) => {
+    return (
+      <div className="nk-h-screen nk-flex nk-items-center">
+        <Tooltip {...args} />
+      </div>
+    );
+  },
 };
 
 export const DefaultMaxTooltipWidth: Story = {
```
lightwalker-eth, line 98: Adding a comment here that GitHub won't let me add several lines lower where I really want to comment.

CleanShot 2024-09-17 at 16 53 23

This creates a maintenance burden because it's easy to forget to update the storybook story if we update this constant later. Better instead to write this as: "This is the default max tooltip width. You can customize the max width if desired. See the "Custom Max Tooltip Width" story for an example of setting a custom max width."

CleanShot 2024-09-17 at 16 56 00

There's no benefit to changing the trigger just for this story (relative to all the others). Better to just keep this the same as all the other stories: "Tooltip Trigger".

CleanShot 2024-09-17 at 16 57 29

It's not good to write content in our stories like this. The problem is that we need to consider how users can customize the control for the `maxTooltipWidth`. Therefore, don't make explicit reference in your writing to a specific value such as "600px" that the user might override through the storybook controls. If you write in this way it can be a source of confusion. You need to think from the perspective of people using and interacting with these stories.

Here's a better way to write this: "This is a customized max tooltip width. You can customize the max width if desired." There's no need to write other ideas here.

apps/storybook.namekit.io/stories/Namekit/Tooltip.stories.tsx
```diff
@@ -45,6 +71,28 @@ export const LeftPlacement: Story = {
...
+  render: (args) => {
+    return (
+      <div className="nk-w-full nk-flex nk-justify-center">
+        <Tooltip {...args} />
+      </div>
+    );
+  },
+};
+
+export const TopPlacement: Story = {
```
lightwalker-eth, line 83: You should move `TopPlacement` to come before `BottomPlacement`. People normally think in order from top, bottom, left, right.

apps/storybook.namekit.io/stories/Namekit/Tooltip.stories.tsx
```diff
@@ -5,18 +5,44 @@ import { Tooltip } from "@namehash/namekit-react/client";
...
+      options: ["top", "right", "bottom", "left"],
+      control: { type: "select" },
+    },
+    maxTooltipWidth: {
+      control: { type: "number" },
     },
   },
+  args: {
+    placement: "top",
+    maxTooltipWidth: 400,
```
lightwalker-eth, line 44: Why are you hardcoding `400` here? This kind of work creates all kinds of problems for maintainability. You should be using the constant we defined.

lightwalker-eth (changes requested):

lightwalker-eth (changes requested):
@FrancoAguzzi Thanks for updates. I've created PR 404 with the suggested enhancements to PR 399: https://github.com/namehash/namekit/pull/404

Once PR 404 is merged into PR 399, I'm happy to approve and merge PR 399 into main :+1:

PR #400, opened 2024-09-16
Add charSplit to ens-utils

djstrong:


Carbon225:
@djstrong I believe this requires a changeset

Carbon225:
@djstrong I don't think you should be updating the versions in package.json manually, it's done automatically during release I think.

Carbon225 (changes requested):
Revert the manual version change and the rest looks fine.

djstrong (commented):
revert version change

Carbon225 (approved):

packages/ens-utils/src/ensname.ts
```diff
@@ -326,6 +326,19 @@ export function getRegistrationPotential(name: ENSName): RegistrationPotential {
   }
 }
 
+/**
+ * Splits a label into its individual characters.
```
lightwalker-eth, line 330:
```suggestion
 * Splits a string into its individual characters.
```

This function is for any string, not just for the subset of strings that are labels.

packages/ens-utils/src/ensname.ts
```diff
@@ -326,6 +326,19 @@ export function getRegistrationPotential(name: ENSName): RegistrationPotential {
...
+ * 
+ * This function uses the spread operator to split the string into an array of its Unicode characters (UTF-32) compatible with ens-normalize.
```
lightwalker-eth, line 332: A few thoughts on this line:

1. Best not to document implementation details. Ex: "This function uses the spread operator to split the string into an array..."
2. Is there a special reason to identify these are UTF-32 characters? Perhaps there's an important idea there? Is this different from the regular / default representation of characters in JavaScript? Or is that splitting different than splitting of UTF-16 characters?
3. Can you help me understand the significance of saying that this is "compatible with ens-normalize"? I imagine there's an important idea there, but need your help to understand it 😄

packages/ens-utils/src/ensname.ts
```diff
@@ -326,6 +326,19 @@ export function getRegistrationPotential(name: ENSName): RegistrationPotential {
...
+ * Each element in the resulting array represents a single Unicode character.
+ * 
+ * @param label
```
lightwalker-eth, line 335:
```suggestion
 * @param text
```

packages/ens-utils/src/ensname.ts
```diff
@@ -326,6 +326,19 @@ export function getRegistrationPotential(name: ENSName): RegistrationPotential {
...
+ * @returns An array of the individual characters within `label`.
```
lightwalker-eth, line 336:
```suggestion
 * @returns An array of the individual characters within `text`.
```

packages/ens-utils/src/ensname.ts
```diff
@@ -326,6 +326,19 @@ export function getRegistrationPotential(name: ENSName): RegistrationPotential {
...
+ */
+export function splitCharacters(label: string): string[] {
```
lightwalker-eth, line 338:
```suggestion
export function splitCharacters(text: string): string[] {
```

packages/ens-utils/src/ensname.ts
```diff
@@ -326,6 +326,19 @@ export function getRegistrationPotential(name: ENSName): RegistrationPotential {
...
+  return [...label];
+}
+
 /**
  * Calculates the number of characters in a label.
```
lightwalker-eth, line 343: Can we also update this function to make it clear that it's about any string in general and not only labels?

lightwalker-eth (changes requested):
@djstrong @Carbon225 Thanks for the updates here. Reviewed and shared feedback mostly around 1 idea :+1:

packages/ens-utils/src/ensname.ts
djstrong: Thank you. I have changed documentation to:
In JavaScript, the ".split()" method of a string may give different results  because it returns an array of UTF-16 code units.
By "compatible with ens-normalize" I meant it uses the same implementation but maybe it is not important.

lightwalker-eth (approved):
Looks good :+1:

PR #404, opened 2024-09-18
Tooltip Enhancements

lightwalker-eth:
CleanShot 2024-09-18 at 23 48 35

This PR:

1. Adds an auto-generated "Docs" page to the Storybook for "Tooltip" that includes code samples for each story. This step was taken to help establish a more clear standard for how we should be building our Storybook stories moving forward.
2. Improves the UX for "Tooltip" in Storybook.
3. Aligns the definition of `TooltipPlacement` with our existing standards.
4. Adds a named constant for `DEFAULT_TOOLTIP_PLACEMENT`.


apps/storybook.namekit.io/.storybook/preview.js
lightwalker-eth, line 1: Not sure why this shows as deleted on GitHub PR review. I properly renamed this file to .tsx in git.

No actions needed here. Just sharing background info.

apps/storybook.namekit.io/.storybook/preview.tsx
```diff
@@ -0,0 +1,51 @@
+import React from "react";
```
lightwalker-eth, line 1: Had to change this file from .js to .tsx and import React, etc.. as I needed to define a custom `docs` param below in the file.

apps/storybook.namekit.io/.storybook/preview.tsx
```diff
@@ -0,0 +1,51 @@
...
+import "../global.css";
+
+const preview: Preview = {
+  parameters: {
+    options: {
+      storySort: {
+        order: ["Nameguard", "Namekit", "ENS Utils", "ENS Webfont"],
+      },
+    },
+    docs: {
```
lightwalker-eth, line 26: Needed to define a custom `docs` param here to hide the `<Primary>` and `<Controls>` sections that are rendered on "Docs" pages by default.

Please see comments below for details.

I already spent several hours on this. Let's please not get stuck on this any more for now and focus on advancing this PR so that we can close it out as soon as possible.

apps/storybook.namekit.io/stories/Namekit/Tooltip.stories.tsx
```diff
@@ -1,24 +1,15 @@
...
-  Tooltip,
-  TooltipPlacement,
-  DEFAULT_MAX_TOOLTIP_WIDTH,
-} from "@namehash/namekit-react/client";
+import { Tooltip, TooltipPlacement } from "@namehash/namekit-react/client";
 
 const meta: Meta<typeof Tooltip> = {
   title: "UI/Tooltip",
   component: Tooltip,
+  tags: ["autodocs"],
```
lightwalker-eth, line 8: This is what tells Storybook to auto generate the "Docs" page.

apps/storybook.namekit.io/stories/Namekit/Tooltip.stories.tsx
```diff
@@ -1,24 +1,15 @@
...
   parameters: {
     layout: "centered",
   },
   argTypes: {
-    trigger: {
```
lightwalker-eth, line 16: It's better to totally hide these `argTypes`. We should only define `argTypes` for args that are proper for a user to directly manipulate in Storybook controls.

Before, it looked like this:

CleanShot 2024-09-18 at 23 33 29

Now it looks like this:

CleanShot 2024-09-18 at 23 34 17

apps/storybook.namekit.io/stories/Namekit/Tooltip.stories.tsx
```diff
@@ -27,10 +18,6 @@ const meta: Meta<typeof Tooltip> = {
       control: { type: "number" },
     },
   },
-  args: {
```
lightwalker-eth, line 30: Removed these `args` because I've learned some new best practices with Storybook.

One key guideline is to focus on making sure you get really clean code examples on the "Docs" page:

CleanShot 2024-09-18 at 23 36 36

If we use `args` here then it doesn't properly auto generate the code for cases where you want to show example code where an optional param is not defined. Ex: `maxTooltipWidth` in the screenshot above.

apps/storybook.namekit.io/stories/Namekit/Tooltip.stories.tsx
```diff
@@ -39,60 +26,88 @@ type Story = StoryObj<typeof Tooltip>;
 
 export const TopPlacement: Story = {
   args: {
-    trigger: <>Tooltip Trigger</>,
```
lightwalker-eth, line 42: I changed the way the `trigger` and `children` args are being defined on all these stories.

The problem is that passing them the way they were defined before doesn't generate the best code examples / UX in Storybook.

A better solution for a case like what we have here is to use the `render` function strategy as I've added to each story.

apps/storybook.namekit.io/stories/Namekit/Tooltip.stories.tsx
```diff
@@ -39,60 +26,88 @@ type Story = StoryObj<typeof Tooltip>;
...
-    children: <>Tooltip content</>,
-    placement: TooltipPlacement.TOP,
+    placement: TooltipPlacement.Top,
+  },
+  render: (args) => {
+    const trigger = <span>Tooltip Trigger</span>;
```
lightwalker-eth, line 32: I changed each of the `trigger` and `children` in each story to be in `<span>` tags rather than `<>` tags. The reason is because when you look at the example code generated in storybook, you see some React implementation details there when you use `<>`.

We want to optimise for really nice auto-generated code examples on the "Docs" pages. Therefore I switched all of these to `<span>`

apps/storybook.namekit.io/stories/Namekit/Tooltip.stories.tsx
```diff
@@ -39,60 +26,88 @@ type Story = StoryObj<typeof Tooltip>;
...
 export const CustomMaxTooltipWidth: Story = {
   args: {
-    trigger: <>Tooltip Trigger</>,
-    children: (
-      <>
-        This is a customized max tooltip width. You can customize the max width
-        if desired.
-      </>
-    ),
-    placement: TooltipPlacement.BOTTOM,
```
lightwalker-eth, line 95: We should make each story as "pure" for communicating a given idea as possible. The idea of a custom tooltip width doesn't need to add any complexity of also specifying a custom placement.

Therefore removed a custom `placement` here.

packages/namekit-react/src/components/Tooltip.tsx
```diff
@@ -12,19 +12,21 @@ type Props = {
 export const DEFAULT_MAX_TOOLTIP_WIDTH = 400;
 
 export const TooltipPlacement = {
-  TOP: "top",
```
lightwalker-eth, line 15: We shouldn't write enum values this way. We should always follow the same standard for how we write them.

packages/namekit-react/src/components/Tooltip.tsx
```diff
@@ -12,19 +12,21 @@ type Props = {
...
 export type TooltipPlacement =
   (typeof TooltipPlacement)[keyof typeof TooltipPlacement];
 
+export const DEFAULT_TOOLTIP_PLACEMENT = TooltipPlacement.Top;
+
 export function Tooltip({
   trigger,
   children,
-  placement = TooltipPlacement.TOP,
+  placement = DEFAULT_TOOLTIP_PLACEMENT,
```
lightwalker-eth, line 29: Defining a named constant is a good strategy for default params. It helps to make code and docs easier to automatically keep updated as things change.

packages/namekit-react/src/components/Tooltip.tsx
```diff
@@ -42,7 +44,7 @@ export function Tooltip({
     <Popover className="nk-flex nk-items-center">
       <Float
         show={open}
-        placement={placement}
+        placement={placement} // our `TooltipPlacement` values are a subset of headlessui's `Placement` values
```
lightwalker-eth, line 47: This was an important idea to document.

lightwalker-eth (commented):
@FrancoAguzzi Adding comments to this PR to help explain the motivation for various changes.

apps/storybook.namekit.io/.storybook/preview.tsx
FrancoAguzzi: I think GitHub has issues when interpreting a file name change

apps/storybook.namekit.io/stories/Namekit/Tooltip.stories.tsx
```diff
@@ -39,60 +26,88 @@ type Story = StoryObj<typeof Tooltip>;
...
-    children: (
-      <>
-        This is the default max tooltip width. You can customize the max width
-        if desired. See the "Custom Max Tooltip Width" story for an example of
-        setting a custom max width.
-      </>
-    ),
-    placement: TooltipPlacement.BOTTOM,
+  render: (args) => {
+    const trigger = <span>Tooltip Trigger</span>;
```
FrancoAguzzi, line 85: Maybe moving this to a DEFAULT constant is a good idea?

FrancoAguzzi (approved):

apps/storybook.namekit.io/stories/Namekit/Tooltip.stories.tsx
lightwalker-eth: @FrancoAguzzi I thought about that too, but I'm happy with a little bit of repetition here because it makes the code so nice and simple.

PR #406, opened 2024-09-18
Resolve tailwind warning

lightwalker-eth:
CleanShot 2024-09-19 at 00 35 03

This PR removes the build warning above.

I'm curious why we need this line at all though? Should we just entirely remove it?


notrab:
@lightwalker-eth I think this is still required, but I'd be curious to know what happens if you remove it.

I believe I added this originally before we were building the unique namespaced styles for Namekit. I don't think we do the same thing with nameguard-react yet, so this is why I think it's needed.

Essentially, `nameguard-react` doesn't export its own styles. So this line in the apps using it, allow the component styles to be detected and built.

We should try to migrate `nameguard-react` to use a prefix `ng-` so it can export its own styles.

PR #408, opened 2024-09-19
Upgrade pnpm to 9.10.0

lightwalker-eth:
CleanShot 2024-09-19 at 15 49 17

This PR advances the following goals:

- [X] Upgrades PNPM to latest version.
- [X] Adds explicit Node.js version dependency to monorepo root (PNPM 9.10.0 requires at least Node.js v18.12.)
- [X] Removes `yup` as a dependency from the root monorepo project (it isn't a dependency here). Moved this dependency where it belongs (at the app / package level, as needed).
- [X] Removes `package-lock.json` and `pnpm-lock.yaml` from non-root directories. We should not use `npm` at all, and therefore we shouldn't have any `package-lock.json` files in the monorepo. Additionally, we should be managing all package dependencies using `pnpm` in the root of the monorepo, and not going into individual subdirectories to introduce additional `pnpm-lock.yaml` files.
- [X] Updates `.gitignore` to help prevent issues like the above from happening again.
- [X] Removes duplicate dependency of `@namehash/namekit-react` inside `apps/nameguard.io/package.json`


apps/nameguard.io/package.json
```diff
@@ -12,7 +12,6 @@
     "@heroicons/react": "2.0.18",
     "@namehash/ens-utils": "workspace:*",
     "@namehash/ens-webfont": "workspace:*",
-    "@namehash/namekit-react": "workspace:*",
```
lightwalker-eth, line 15: We're not removing this dependency. I'm just removing a duplicate reference to this dependency. It's listed twice in this package.json.

notrab (approved):
Looks great!

PR #411, opened 2024-09-21
Refine CI

lightwalker-eth:
# Goals

1. Enhance use of environment variables to move in the direction of improved local development workflows.
   1. Created and documented `api/.env.example`.
   2. Created `api/package.json` with new `deploy:prod` script.
   3. Remove unused environment variables in GitHub Actions
2. Improve consistency in GitHub Actions
   1. Align Node.js version in CI with Node.js version that we use locally via .nvmrc
   2. Ensure we use same Node.js and PNPM versions locally as we do in GitHub actions
   3. More properly install npm dependencies in a CI context using `pnpm install --frozen-lockfile`.
   4. Ensure we use same `ubuntu-latest` on all GitHub Actions.
   5. Apply consistent strategy for installing pnpm / node.js / installing npm dependencies across all GitHub Actions.
3. Remove redundant GitHub Action `.github/workflows/nameguard_api_cicd.yml`.
4. Enhance security in GitHub Actions
   1. Reduce permissions in GitHub Actions to minimum required
5. Enable manual triggering of GitHub Actions through GitHub UI
   1. Added `workflow_dispatch` GitHub Action trigger
6. Enhance performance of GitHub Actions
   1. Utilize caching for pnpm in GitHub Actions.
   2. Remove unnecessary steps in GitHub Actions.
7. Remove failure scenarios
   1. Prevent concurrent executions of GitHub Action to deploy NameGuard Lambda to production
   2. Handle "deployment cancellation" case.
8. Remove GitHub Actions deprecation warnings.
   1. Upgrade to `actions/checkout@v4`
   2. Upgrade to `pnpm/action-setup@v4`
   3. Upgrade to `actions/setup-node@v4`
   4. Upgrade to `actions/setup-python@v5`
9. Other misc updates:
   1. Refine documentation and naming in GitHub Actions.
   2. Added LICENSE file to each package and app in monorepo.

# Screenshots of some of the warnings this PR resolves

CleanShot 2024-09-21 at 13 35 02

CleanShot 2024-09-26 at 01 10 47

CleanShot 2024-09-25 at 19 07 03

CleanShot 2024-09-26 at 00 06 05


api/serverless.yml
```diff
@@ -32,20 +30,18 @@ functions:
     name: oss-nameguard-${self:custom.stage}
     memorySize: 1769
     timeout: 10
-    versionFunctions: true
```
lightwalker-eth, line 35: This was not a valid property

pnpm-workspace.yaml
```diff
@@ -2,3 +2,4 @@ packages:
   - "apps/**"
   - "packages/**"
   - "internal/**"
+  - "api"
```
lightwalker-eth, line 5: Added a `package.json` here. We need pnpm to find it so that the root directory `pnpm-lock.yaml` is built properly.

api/.docker/config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "credsStore": "ecr-login"
```
lightwalker-eth, line 2: Created this file to resolve this warning:

image

lightwalker-eth (commented):
Completed self review :+1:

api/serverless.yml
BanaSeba: You mean that this is `true` by default right?

BanaSeba (approved):
Everything looks alright. If default versioning is OK as mentioned in my comment we can proceed with merge. Otherwise we need to set versioning to false and remove serverless-prune plugin.

BanaSeba (changes requested):
Lambda deployment is not triggered even though it should. It might be caused by invalid YAML formatting

api/serverless.yml
lightwalker-eth: image

This property was causing the error above.

PR #420, opened 2024-09-27
Rename "NameGuard result" to "NameGuard report"

djstrong:


lightwalker-eth (commented):
Code review looks good :+1: Now going to investigate unit test errors.

lightwalker-eth (approved):
@djstrong Thanks :+1: Approving this PR with the optimistic assumption that the CI errors are all due to either:
1. We have to deploy a new production NameGuard Lambda for some of these tests to work.
2. Coinbase is returning errors for offchain names in our unit tests right now, ex: `exampleprimary.cb.id`. These errors are on their side, not ours.

PR #439, opened 2024-10-15
feat(namekit-react): Identity

notrab:
This PR adds a preview for the mechanics of the `<Identity />` component. You can see a preview of it here:

https://github.com/user-attachments/assets/47096a1a-d447-42ff-b040-cba7bce0ba4b

Keep in mind that this is a technical preview, and no styling has been implemented yet, and I won't spend time on the styles until we know we are fetching all the data we want, and the DSL is nice.

One thing I dislike about PR is it now adds as a dependency `@namehash/nameguard`. Instead, I would prefer this to be a peer dependency, and anyone wanting to use this component should install `@namehash/nameguard` AND `@namehash/namekit-react`.

### DIFF
--- a/apps/storybook.namekit.io/stories/Namekit/Identity.stories.tsx
+++ b/apps/storybook.namekit.io/stories/Namekit/Identity.stories.tsx
@@ -1,0 +1,59 @@
+import React from "react";
+import type { Meta, StoryObj } from "@storybook/react";
+import { Identity } from "@namehash/namekit-react";
+
+const meta: Meta<typeof Identity.Root> = {
+  title: "Namekit/Identity",
+  component: Identity.Root,
+  argTypes: {
+    address: { control: "text" },
+    network: {
+      control: {
+        type: "select",
+        options: ["mainnet", "sepolia"],
+      },
+    },
+    className: { control: "text" },
+  },
+};
+
+export default meta;
+
+type Story = StoryObj<typeof Identity.Root>;
+
+const IdentityCard: React.FC<{
+  address: string;
+  network?: "mainnet" | "sepolia";
+}> = ({ address, network }) => (
+  <Identity.Root address={address} network={network}>
+    <Identity.Avatar />
+    <Identity.Name />
+    <Identity.Address />
+    <Identity.NameGuardShield />
+  </Identity.Root>
+);
+
+export const Default: Story = {
+  args: {
+    address: "0x838aD0EAE54F99F1926dA7C3b6bFbF617389B4D9",
+    network: "mainnet",
+    className: "rounded-xl",
+  },
+  render: (args) => <IdentityCard {...args} />,
+};
+
+export const MultipleCards: Story = {
+  render: () => (
+    <>
+      <IdentityCard address="0x838aD0EAE54F99F1926dA7C3b6bFbF617389B4D9" />
+      <IdentityCard address="0x123456789abcdef123456789abcdef123456789a" />
+      <IdentityCard address="0xabcdef123456789abcdef123456789abcdef1234" />
+      <IdentityCard address="0x987654321fedcba987654321fedcba987654321f" />
+      <IdentityCard
+        address="0xfedcba987654321fedcba987654321fedcba9876"
+        network="sepolia"
+      />
+      <IdentityCard address="0x111222333444555666777888999000aaabbbcccd" />
+    </>
+  ),
+};
--- a/packages/namekit-react/package.json
+++ b/packages/namekit-react/package.json
@@ -43,6 +43,7 @@
     "@headlessui/react": "1.7.17",
     "@namehash/ens-utils": "workspace:*",
     "@namehash/ens-webfont": "workspace:*",
+    "@namehash/nameguard": "workspace:*",
     "classcat": "5.0.5"
   },
   "devDependencies": {
--- a/packages/namekit-react/src/components/Identity.tsx
+++ b/packages/namekit-react/src/components/Identity.tsx
@@ -1,0 +1,145 @@
+import React, {
+  createContext,
+  useContext,
+  useState,
+  useEffect,
+  ReactNode,
+} from "react";
+import {
+  createClient,
+  Network,
+  type SecurePrimaryNameResult,
+} from "@namehash/nameguard";
+
+interface IdentityContextType extends SecurePrimaryNameResult {
+  address: string;
+}
+
+const IdentityContext = createContext<IdentityContextType | null>(null);
+
+const useIdentity = () => {
+  const context = useContext(IdentityContext);
+
+  if (!context) {
+    throw new Error("useIdentity must be used within an IdentityProvider");
+  }
+
+  return context;
+};
+
+interface SubComponentProps {
+  className?: string;
+  children?: ReactNode;
+}
+
+interface RootProps {
+  address: string;
+  network?: Network;
+  className?: string;
+  children: ReactNode;
+}
+
+const Root = ({
+  address,
+  network = "mainnet",
+  className,
+  children,
+  ...props
+}: RootProps) => {
+  const [data, setData] = useState<IdentityContextType | null>(null);
+  const [loading, setLoading] = useState(true);
+  const [error, setError] = useState<string | null>(null);
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        setLoading(true);
+
+        const nameguard = createClient({ network });
+
+        const result = await nameguard.getSecurePrimaryName(address, {
+          returnNameGuardReport: true,
+        });
+
+        setData({ ...result, address });
+      } catch (err) {
+        setError(
+          err instanceof Error ? err.message : "An unknown error occurred",
+        );
+      } finally {
+        setLoading(false);
+      }
+    };
+
+    fetchData();
+  }, [address, network]);
+
+  if (loading) return <div>Loading...</div>;
+  if (error) return <div>Error: {error}</div>;
+  if (!data) return null;
+
+  return (
+    <IdentityContext.Provider value={data}>
+      <div className={`namekit-identity ${className}`} {...props}>
+        {children}
+      </div>
+    </IdentityContext.Provider>
+  );
+};
+
+const Avatar = ({ className, children, ...props }: SubComponentProps) => {
+  const { display_name } = useIdentity();
+
+  return (
+    <div className={`namekit-avatar ${className}`} {...props}>
+      <img
+        src={`https://avatar.example.com/${display_name}`}
+        alt={display_name}
+      />
+      {children}
+    </div>
+  );
+};
+
+const Name = ({ className, ...props }: SubComponentProps) => {
+  const { display_name } = useIdentity();
+
+  return (
+    <div className={`namekit-name ${className}`} {...props}>
+      {display_name}
+    </div>
+  );
+};
+
+const Address = ({ className, ...props }: SubComponentProps) => {
+  const { address } = useIdentity();
+
+  return (
+    <div className={`namekit-address ${className}`} {...props}>
+      {address}
+    </div>
+  );
+};
+
+const NameGuardShield = ({ className, ...props }: SubComponentProps) => {
+  const { nameguard_report } = useIdentity();
+
+  return (
+    <div className={`namekit-nameguard-shield ${className}`} {...props}>
+      <div className="namekit-nameguard-rating">
+        Rating: {nameguard_report?.rating}
+      </div>
+      <div className="namekit-nameguard-risk-count">
+        Risks: {nameguard_report?.risk_count}
+      </div>
+    </div>
+  );
+};
+
+export const Identity = {
+  Root,
+  Avatar,
+  Name,
+  Address,
+  NameGuardShield,
+};
--- a/packages/namekit-react/src/index.ts
+++ b/packages/namekit-react/src/index.ts
@@ -13,3 +13,4 @@ export { ENSTextArea } from "./components/ENSTextArea";
 export { Input } from "./components/Input";
 export { ENSInput } from "./components/ENSInput";
 export { Link } from "./components/Link";
+export { Identity } from "./components/Identity";
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -568,6 +568,9 @@ importers:
       '@namehash/ens-webfont':
         specifier: workspace:*
         version: link:../ens-webfont
+      '@namehash/nameguard':
+        specifier: workspace:*
+        version: link:../nameguard-sdk
       classcat:
         specifier: 5.0.5
         version: 5.0.5
@@ -3548,6 +3551,7 @@ packages:
   eslint@8.57.1:
     resolution: {integrity: sha512-ypowyDxpVSYpkXr9WPv2PAZCtNip1Mv5KTW0SCurXv/9iOpcrH9PaqUElksqEB6pChqHGDRCFTyrZlGhnLNGiA==}
     engines: {node: ^12.22.0 || ^14.17.0 || >=16.0.0}
+    deprecated: This version is no longer supported. Please see https://eslint.org/version-support for other options.
     hasBin: true
 
   esniff@1.1.3:
@@ -7591,7 +7595,7 @@ snapshots:
   '@eslint/eslintrc@2.1.4':
     dependencies:
       ajv: 6.12.6
-      debug: 4.3.7
+      debug: 4.3.7(supports-color@8.1.1)
       espree: 9.6.1
       globals: 13.24.0
       ignore: 5.3.2
@@ -7762,7 +7766,7 @@ snapshots:
   '@humanwhocodes/config-array@0.13.0':
     dependencies:
       '@humanwhocodes/object-schema': 2.0.3
-      debug: 4.3.7
+      debug: 4.3.7(supports-color@8.1.1)
       minimatch: 3.1.2
     transitivePeerDependencies:
       - supports-color
@@ -9023,7 +9027,7 @@ snapshots:
       '@typescript-eslint/types': 8.8.1
       '@typescript-eslint/typescript-estree': 8.8.1(typescript@5.6.2)
       '@typescript-eslint/visitor-keys': 8.8.1
-      debug: 4.3.7
+      debug: 4.3.7(supports-color@8.1.1)
       eslint: 8.57.1
     optionalDependencies:
       typescript: 5.6.2
@@ -9049,7 +9053,7 @@ snapshots:
     dependencies:
       '@typescript-eslint/typescript-estree': 8.8.0(typescript@5.6.2)
       '@typescript-eslint/utils': 8.8.0(eslint@8.57.1)(typescript@5.6.2)
-      debug: 4.3.7
+      debug: 4.3.7(supports-color@8.1.1)
       ts-api-utils: 1.3.0(typescript@5.6.2)
     optionalDependencies:
       typescript: 5.6.2
@@ -9082,7 +9086,7 @@ snapshots:
     dependencies:
       '@typescript-eslint/types': 8.8.0
       '@typescript-eslint/visitor-keys': 8.8.0
-      debug: 4.3.7
+      debug: 4.3.7(supports-color@8.1.1)
       fast-glob: 3.3.2
       is-glob: 4.0.3
       minimatch: 9.0.4
@@ -9097,7 +9101,7 @@ snapshots:
     dependencies:
       '@typescript-eslint/types': 8.8.1
       '@typescript-eslint/visitor-keys': 8.8.1
-      debug: 4.3.7
+      debug: 4.3.7(supports-color@8.1.1)
       fast-glob: 3.3.2
       is-glob: 4.0.3
       minimatch: 9.0.4
@@ -9917,19 +9921,11 @@ snapshots:
     dependencies:
       ms: 2.1.3
 
-  debug@4.3.4:
-    dependencies:
-      ms: 2.1.2
-
   debug@4.3.4(supports-color@8.1.1):
     dependencies:
       ms: 2.1.2
     optionalDependencies:
       supports-color: 8.1.1
-
-  debug@4.3.7:
-    dependencies:
-      ms: 2.1.3
 
   debug@4.3.7(supports-color@8.1.1):
     dependencies:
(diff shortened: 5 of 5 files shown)


packages/namekit-react/src/components/Identity.tsx
```diff
@@ -0,0 +1,145 @@
...
+        setLoading(false);
+      }
+    };
+
+    fetchData();
+  }, [address, network]);
+
+  if (loading) return <div>Loading...</div>;
+  if (error) return <div>Error: {error}</div>;
+  if (!data) return null;
```
lightwalker-eth, line 79: Suggest a few updates here:

1. The `Root` and all of its children should always be rendered, even if data is still loading or failed to load. The responsibility for handling these cases should be pushed down into child components. Ex: `Avatar` should know what to do when it is loading or when there was a loading error, etc..
2. ... continuing on from above: Suggest updating the interface that is returned by `useIdentity` so that all of the following are true:
   1. Independent of any loading state, it is always guaranteed to be able to get the following values: `address` and `network` (these props should never be undefined).
   2. There's some distinct field(s) representing the loading state / request error state for the identity.
   3. We shouldn't have the returned value `... extends SecurePrimaryNameResult` (because this value might still be loading, etc..). Therefore we should have this as an optional property of the returned value that is undefined unless the loading state field is some value representing it has been successfully loaded.

lightwalker-eth (commented):
@notrab Very cool :+1: Reviewed and shared feedback 😄

packages/namekit-react/src/components/Identity.tsx
notrab: Addressed

packages/namekit-react/package.json
```diff
@@ -43,6 +43,7 @@
     "@headlessui/react": "1.7.17",
     "@namehash/ens-utils": "workspace:*",
     "@namehash/ens-webfont": "workspace:*",
+    "@namehash/nameguard": "workspace:*",
```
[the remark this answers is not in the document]
notrab: FYI: This hasn't been changed yet.

packages/namekit-react/src/components/Identity.tsx
```diff
@@ -0,0 +1,292 @@
...
+  className?: string;
+  children?: ReactNode;
+}
+
+interface RootProps {
+  address: string;
+  network?: Network;
+  className?: string;
+  children: ReactNode;
+  returnNameGuardReport?: boolean;
```
lightwalker-eth, line 46: @notrab Hey, suggest extending this idea. When someone defines an Identity Root, suggest that they also identify what lookups they want to make for that identity.

For example, suggest we support optional lookups of each of the following:
1. NameGuardReport.
2. EFP
   1. Stats (https://docs.ethfollow.xyz/api/users/stats/ -- you've already implemented this in a way, just suggesting it is restructured to be optional).
   2. Common Followers (https://docs.ethfollow.xyz/api/users/commonfollowers/)
   3. Is current user following this identity? (https://docs.ethfollow.xyz/api/users/followerstate/)
3. ENS Profile, for example, social media handles, etc.. To save some time, we should take advantage of existing services for this. Suggest we use this service for now: https://enstate.rs/docs#tag/single-profile/GET/u/{name_or_address}

A few other important suggestions here:
1. For items 2 and 3 listed under EFP above, we also need to know the address of the current user (as opposed to the address of the identity we want to show data about). Please see my separate comment where I suggest we add wagmi as a dependency. If we do this, we can take advantage of the https://wagmi.sh/react/api/hooks/useAccount hook for this purpose. Then, based on the value returned by the `useAccount` hook, we could have some intelligent logic inside the EFP related components: If the current user is known, we also work to load the items 2 and 3 listed under EFP above. If we don't know who the current user is, then we don't load items 2 and 3 listed above, since that context only makes sense in relation to knowing who the current user is.
2. Strongly suggest we don't try to do all this data fusion ourselves between all these different API calls on our frontend. Instead, suggest we move this data fusion API calls inside our own backend for a number of reasons. Let's keep it maximally simple for now and just do the most basic implementation of this API inside the backend of one of our existing open source Next.js apps that are in our monorepo (such as the NameGuard app). We can do a more advanced version of this API in a separate future PR that would create a new app for api.namekit.io where we could move this backend logic in the future).
   1. Therefore, the API call to NameGuard would move to our backend and not happen in the frontend here. Please ask me if any questions. I'm trying to simplify the frontend so that it only needs to make 1 API call per Identity.
   2. One theoretical downside to moving all this work to the backend is increased latency for showing partial results. Ex: If data needs to be loaded via 4 external API calls, and the backend waits for all 4 API responses before returning its own response, the frontend might have to wait longer if 1 of those 4 API responses is meaningfully slower than others. However, I don't suggest we worry about that for now. If we identity this as an issue / opportunity for the future we can pick from a number of techniques where our backend API can stream data to the frontend as it loads.
4. It seems to me it would be a nice DX to make it so the Identity context / hook would include info on which items were requested to be loaded from the API. In other words, it seems nice if these components could separately identify the case of we never requested to load the ENS Profile vs there being no ENS profile for a given identity. Does that make sense?

packages/namekit-react/src/components/Identity.tsx
```diff
@@ -0,0 +1,292 @@
...
+  });
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        setData((prev) => ({ ...prev, loadingState: "loading" }));
+
+        const nameguard = createClient({ network });
+
+        const result = await nameguard.getSecurePrimaryName(address, {
```
lightwalker-eth, line 71: Suggest that we add ReactQuery as a dependency / peer dependency for NameKit React.

The web3 community has generally adopted ReactQuery as an unofficial standard. There's a ton of various libraries and packages already using it or built on it.

For example, strongly suggest having a look at the following:

1. wagmi. https://wagmi.sh/ The same team that built wagmi built viem https://viem.sh/. Pretty much everyone is using both libraries in their web3 apps. viem is a more low-level library that's suitable for use on a backend or frontend. wagmi is a frontend library that wraps viem. wagmi uses ReactQuery for all their network requests.
2. onchainkit. https://onchainkit.xyz/ This is built as a layer on top of wagmi.
3. rainbowkit. https://www.rainbowkit.com/ This is also built as a layer on top of wagmi.
4. Etc..

In fact, suggest that we not only add ReactQuery as a dependency, but we also add wagmi as a dependency. We shouldn't try to reinvent the whole universe. Pretty much everyone who might use NameKit React is also going to be using wagmi, so we better make it convenient for everything to work together.

packages/namekit-react/src/components/Identity.tsx
```diff
@@ -0,0 +1,292 @@
...
+        setData((prev) => ({
+          ...prev,
+          loadingState: "error",
+          error:
+            err instanceof Error ? err.message : "An unknown error occurred",
+        }));
+      }
+    };
+
+    const fetchFollowersData = async () => {
```
lightwalker-eth, line 90: Suggest we put some special care into all the names we use for things. For example, if EFP calls this idea "stats", then suggest we use similar language in our own code.

packages/namekit-react/src/components/Identity.tsx
```diff
@@ -0,0 +1,292 @@
...
+const ENSProfileLink = ({ className, ...props }: SubComponentProps) => {
+  const { identityData, loadingState } = useIdentity();
+
+  if (loadingState !== "success" || !identityData?.display_name) {
+    return null;
+  }
+
+  return (
+    <a
+      href={`https://app.ens.domains/${identityData.display_name}`}
```
lightwalker-eth, line 272: We need some strategy for people who use NameKit React to customize the link target for viewing a detailed profile. This needs to support a variety of cases, including calling a function to a traditional href, customizing opening in a new tab or not, etc..

Additionally, the way this link is built may vary. For example, in some apps the link to a profile is based on an address. In other apps, the link is based on an ENS Name (where possible) and only uses a link with an address as a fallback if the address has no primary name.

For example, consider an app such as EFP. When someone clicks on a profile link on their app, they don't want it to open to the official ENS App. They want it to open to the profile page on their own app.

I worked on some ideas for how a goal like this in the following PR. Open to other suggested strategies for achieving this dynamic logic: Check the logic around the file packages/nameguard-react/src/utils/openreport.ts it might give some inspiration? https://github.com/namehash/namekit/pull/282/files#diff-4b7ca1b4f908b4cf07782ef287606daa30f6c17f8181e6d7a3ab82eb09137dbd

lightwalker-eth (changes requested):
@notrab Hey, great to see the progress you've made here. Shared some feedback for our overall strategy on next steps :+1: Appreciate your advice!

packages/namekit-react/src/components/Identity.tsx
```diff
@@ -115,31 +185,108 @@ const Address = ({ className, ...props }: SubComponentProps) => {
...
+      target="_blank"
+      rel="noopener noreferrer"
+      className={`namekit-ens-profile-link ${className}`}
+      {...props}
+    >
+      <ENSLogo />
```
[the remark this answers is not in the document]
notrab: Agreed

packages/namekit-react/src/components/Identity.tsx
```diff
@@ -115,31 +185,108 @@ const Address = ({ className, ...props }: SubComponentProps) => {
...
+
+const ENSProfileLink = ({ className, ...props }: SubComponentProps) => {
+  const { identityData, loadingState } = useIdentity();
+
+  if (loadingState !== "success" || !identityData?.display_name) {
+    return null;
+  }
+
+  return (
+    <a
```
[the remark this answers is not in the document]
notrab: In theory, yes. But I want to make more progress on the core implementation before moving to UI.

packages/namekit-react/src/components/Identity.tsx
notrab: @lightwalker-eth this is what I have in mind that I would like to experiment this week...

## 1. Standard implementation
This shows the default link:

```tsx
<Identity.Root address={address}>
  <Identity.ENSProfileLink />
</Identity.Root>
```

## 2. Custom NameKit Config

```tsx
const efpConfig = {
  profileLinks: {
    getProfileTarget: ({ address, ensName }) => 
      `/profile/${ensName || address}`,
    openInNewTab: false,
    // Optionally use Next.js Link or other routing component
    LinkComponent: NextLink,
  }
};

<NameKitConfigProvider config={efpConfig}>
  <Identity.Root address={address}>
    <Identity.ENSProfileLink />
  </Identity.Root>
</NameKitConfigProvider>
```

Optionally, a click handler similar to what you suggested:

```tsx
const handlerConfig = {
  profileLinks: {
    getProfileTarget: ({ address, ensName }) => 
      (e: React.MouseEvent) => {
        e.preventDefault();
        // Custom handling logic
        openProfileModal(address, ensName);
      }
  }
};
```

This allows us to move towards the "React Query" context provider for data too.

lightwalker-eth: @notrab Hey, thanks for putting this together and sharing it 🙌 🚀 

A few suggestions:

1. `getProfileTarget` am I correct to understand this is supposed to return a string representing a URL? If so, suggest renaming this to `getProfileURL`.
   1. I also see your separate click handler example, but confused how everything fits together here. Because here the `getProfileTarget` doesn't return a string, it returns a function that takes a `React.MouseEvent` param.
2. `openInNewTab` and `LinkComponent` I like the general direction you're working towards here for supporting more customizability. But maybe we use a different strategy? What do you think of modifying the approach there to instead have a function called `getProfileLink` that takes in a `url` (as might be generated by `getProfileURL` and then returns a DOM element tree for turning that `url` into the HTML for a link.
   1. This strategy not only would make customizing link targets / link components no problem, but I believe it should also give a solution for the "click handler" case.
   2. Maybe instead of `getProfileLink` taking as input a `url` as a param, it should instead take the same param type as `getProfileTarget` / `getProfileURL`. Then, if `getProfileLink` wants to build a link with a URL in it, it can call `getProfileURL` itself to do that. Alternatively, if `getProfileLink` wants to do some javascript-based "click handler" it can then more easily define that logic if the param passed to `getProfileLink` were an address or an ensName.
3. For now, maybe we exclusively build support for getting the profiles using address, not using ensName. I'm worried supporting both will slow us down right now. It's definitely nice to ultimately add support for both in the future, but suggest we come back to that later. For example, in the future we might define some special interface for what gets passed into `Identity.Root` where it is either a name or an address and not both. This same interface could then be passed around to other downstream components such as the callback functions defined in the config, etc.. But suggest we keep things more simple for now and only focus on identities that are anchored in an address and not an ensName.
4. Suggest that we define some default handlers for the profile link configs. The default should result in opening to view the profile of the provided address on the ens.domains website in a new tab. Ex: https://app.ens.domains/0xf81bc66316A3f2A60Adc258F97F61dFcBdd23Bb1
5. It seems nice to make it so that `Identity.ENSProfileLink` just defines a link, not the anchor elements for the link`. As I understand we should allow any children to be passed into `Identity.ENSProfileLink` which would then become the anchor for the link. As I understand, just doing `<Identity.ENSProfileLink />` should be similar to doing `<a />` in that there shouldn't be anything for a user to see or interact with until you define the children beneath that.
6. Whatever interface we decide on for setting the config for profile links, it might be nice to allow a custom config for a specific `Identity.ENSProfileLink` instance to be passed in. By default people wouldn't do that, so then it would fallback to the default config set in `NameKitConfigProvider`.
7. Perhaps good to rename `Identity.ENSProfileLink` to just `Identity.ProfileLink`. I like making "ENS" implied rather than explicit. For example, use `name` instead of `EnsName`, etc.. If we don't do this we'll have "ENS" too much all over our code.

Appreciate your advice :+1:

notrab: @lightwalker-eth I've applied most of this feedback in a new commit. I agree with it all.

I also updated the stories to match the new requirements too. I hope I didn't miss anything, it took some refactoring/rethinking to make it all work better 🙏🏻

packages/namekit-react/src/components/Identity.tsx
notrab: @lightwalker-eth this is fantastic feedback, and the ideas are awesome.

I do prefer a DX whereby there's less configuration. So if you use the `Identity.Followers` then it knows to fetch followers (like it does now). The only thing not following this is `nameGuardReport`, it could be messy but I think we could export another component that is `Identity.NameGuardReport` and if that's used at all, the initial query returns the report too. If that makes sense.

However, to be more explicit with the "options" available as you suggested, we could do something like this:

```ts
interface IdentityLookupOptions {
  nameGuardReport?: boolean;
  efpStats?: boolean;
  efpCommonFollowers?: boolean;
  efpFollowState?: boolean;
  ensProfile?: boolean;
}
```

Which means we can do something like this:

```tsx
<Identity.Root
  address="0x..."
  network="mainnet"
  lookupOptions={{
    nameGuardReport: true,
    efpStats: true,
    ensProfile: true,
  }}
>
  <Identity.Avatar />
  <Identity.Name />
  <Identity.Address />
  <Identity.NameGuardShield />
  <Identity.ProfileLink>View Profile</Identity.ProfileLink>
  <Identity.Followers />
</Identity.Root>
```

apps/storybook.namekit.io/stories/Namekit/Identity.stories.tsx
```diff
@@ -35,29 +35,147 @@ const IdentityCard: React.FC<{
...
+    getProfileURL: (address: string) => `/profiles/${address}`,
+    getProfileLink: (address: string, children: React.ReactNode) => (
+      <a href={`/profiles/${address}`}>{children}</a>
+    ),
+  },
+};
+
+const CustomAppIdentityCard: React.FC<{ address: string }> = ({ address }) => (
+  <NameKitProvider config={customAppConfig}>
+    <Identity.Root address={address}>
```
lightwalker-eth, line 59: Suggest we make a few updates here to put our "Identity" components on a stronger foundation. Specifically: an "identity reference" is more than just an address. It is a combination of a chain and an address.

1. Suggest defining an interface for `ImplicitIdentityReference` that combines both an address and a chain, but where the chain param is optional. The purpose of this interface is to optimize DX for the average case where all addresses within an app are on the same chain. It should be noted how this isn't always the case though, and increasingly won't be in the future the way things are headed.
2. Suggest defining an interface for `IdentityReference` that extends `ImplicitIdentityReference` but makes the chain param required.
3. Suggest defining a little utility function that can be used within the context of a NameKitProvider that automatically converts any `ImplicitIdentityReference` to an explicit `IdentityReference` using the configured chain as the default.
4. There's a bunch of functions here that currently only take an "address" param. These should change so that they instead take a param of type `IdentityReference` or `ImplicitIdentityReference`. Inside the implementation of these functions, we can call the utility function described above that takes either an `IdentityReference` or `ImplicitIdentityReference` as a param and always returns an explicit `IdentityReference` for use in all our internal logic.
5. Functions such as `getProfileURL` and `getProfileLink` should take an `IdentityReference` as a param. The default implementation of these functions should then consider the chain parameter and not only the address. For example. If the chain param is mainnet, then our existing strategy is good. If the chain param is anything else (such as sepolia, etc..), then there isn't an existing profile service we can easily link to for this. So instead suggest making our default implementation of all the profile link generator functions trigger some alert box or something like that for the case that the chain is anything but mainnet. Once we build the backend APIs into NameKit for generating profile data, we can come back to this and implement some improved default behaviour for showing profiles on chains other than mainnet.

Here's an acceptance criteria for everything described above: If I use our Identity components to show the identity for a chain other than mainnet (such as sepolia) I shouldn't get a profile link to view the profile of that address on mainnet as that wouldn't actually be the correct profile / identity.

Please ask me if any questions :+1:

lightwalker-eth (changes requested):
@notrab Thanks for updates :+1: Reviewed and shared feedback

apps/storybook.namekit.io/stories/Namekit/Identity.stories.tsx
```diff
@@ -35,29 +35,147 @@ const IdentityCard: React.FC<{
...
+      <Identity.Avatar />
+      <Identity.Name />
+      <Identity.ProfileLink>
+        <button>View Profile</button>
```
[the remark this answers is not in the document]
notrab: As I mentioned before, I'm avoiding the UI side of things until we nail the DX. Swapping out components is easy, but if we need to make changes to the UI components, it could cause conflicts so I've opted not to focus on that yet.

apps/storybook.namekit.io/stories/Namekit/Identity.stories.tsx
```diff
@@ -35,29 +35,147 @@ const IdentityCard: React.FC<{
...
+    </Identity.ProfileLink>
     <Identity.Followers />
   </Identity.Root>
 );
 
+const customAppConfig = {
+  profileLinks: {
+    getProfileURL: (address: string) => `/profiles/${address}`,
+    getProfileLink: (address: string, children: React.ReactNode) => (
+      <a href={`/profiles/${address}`}>{children}</a>
```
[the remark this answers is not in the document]
notrab: Same comment as before.

packages/namekit-react/src/components/Identity.tsx
```diff
@@ -243,42 +268,40 @@ const Followers = ({ className, ...props }: SubComponentProps) => {
...
+const ProfileLink = ({
+  className,
+  children,
+  config: instanceConfig,
+  ...props
+}: ProfileLinkProps) => {
+  const { loadingState, address } = useIdentity();
+  const globalConfig = useNameKitConfig();
+
+  if (loadingState !== "success") {
```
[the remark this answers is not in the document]
notrab: Good point. I remove the check for that, it's a habit of always checking and rendering fallbacks.

packages/namekit-react/src/components/Identity.tsx
notrab: @lightwalker-eth I made some changes based on what you said about the idea of a class for the `ProfileLinkGenerator`.

```ts
export class ProfileLinkGenerator {
  private baseURL: string;
  private name: string;

  constructor(name: string, baseURL: string) {
    this.name = name;
    this.baseURL = baseURL;
  }

  getName(): string {
    return this.name;
  }

  getProfileURL(address: string): string {
    return `${this.baseURL}${address}`;
  }
}
```

This is basic for now, but acts as a way to set a base URL, and we could even export some default ones, such as:

```ts
export const ENSProfileLink = new ProfileLinkGenerator(
  "ENS",
  "https://app.ens.domains/",
);
```

Then this can be used wherever:

```ts
interface ProfileLinkProps {
  config?: ProfileLinkGenerator;
  className?: string;
  children?: React.ReactNode;
  onClick?: (e: React.MouseEvent) => void;
}

const ProfileLink: React.FC<ProfileLinkProps> = ({ config, children, onClick }) => {
  const identity = useIdentity();
  const nameKitConfig = useNameKitConfig();

  const linkConfig = config || nameKitConfig.profileLinks?.[0] || DEFAULT_PROFILE_LINKS[0];

  if (!identity) {
    console.warn("ProfileLink used outside of Identity context");
    return null;
  }

  const url = linkConfig.getProfileURL(identity.address);

  return (
    <a
      href={url}
      target="_blank"
      rel="noopener noreferrer"
      className="namekit-profile-link"
      onClick={onClick}
    >
      {children || linkConfig.getName()}
    </a>
  );
};
```

This also allows ProfileLinks to be controlled individually, or part of the context provider down. Giving more flexibility.

packages/namekit-react/src/components/Identity.tsx
```diff
@@ -12,21 +12,46 @@ import {
...
+export const ENSProfileLink = new ProfileLinkGenerator(
+  "ENS",
+  "https://app.ens.domains/",
+);
+
+const DEFAULT_PROFILE_LINKS = [ENSProfileLink];
+
 interface NameKitConfig {
-  profileLinks?: ProfileLinkConfig;
+  profileLinks?: ProfileLinkGenerator[];
```
lightwalker-eth, line 41: Similar to the feedback above. Can you help me understand the goal for representing this as an array? In my thinking there should be exactly one, not 0 or many.
```suggestion
  profileLinkGenerator: ProfileLinkGenerator;
```

packages/namekit-react/src/components/Identity.tsx
```diff
@@ -12,21 +12,46 @@ import {
 } from "@namehash/nameguard";
 import cc from "classcat";
 
-export interface ProfileLinkConfig {
-  getProfileURL: (address: string) => string;
-  getProfileLink: (address: string, children: ReactNode) => JSX.Element;
+export class ProfileLinkGenerator {
+  private baseURL: string;
```
lightwalker-eth, line 16: I don't think this should have a "baseURL". This assumes too much about how the URLs will be constructed, especially when considering how they should consider both chain and address.

packages/namekit-react/src/components/Identity.tsx
```diff
@@ -12,21 +12,46 @@ import {
...
+  private name: string;
+
+  constructor(name: string, baseURL: string) {
+    this.name = name;
+    this.baseURL = baseURL;
+  }
+
+  getName(): string {
```
lightwalker-eth, line 24: Can you help me understand the goal for `getName`? It will help a lot to document ideas in comments.

packages/namekit-react/src/components/Identity.tsx
```diff
@@ -268,40 +293,69 @@ const Followers = ({ className, ...props }: SubComponentProps) => {
...
+      rel="noopener noreferrer"
+      className="namekit-profile-link"
+      onClick={onClick}
+    >
+      {children || linkConfig.getName()}
+    </a>
+  );
+};
+
+interface ProfileLinksProps {
```
lightwalker-eth, line 334: Let's have a chat to help align on goals here. I think we're drifting away from solving the problems that people are actually looking to solve.

lightwalker-eth (changes requested):
@notrab Hey thanks for updates. Reviewed and shared feedback.

PR #453, opened 2024-10-29
feat: [sc-25878] Implement Checkbox Component in NameKit React

eduramme:
Story details: https://app.shortcut.com/ps-web3/story/25878


packages/namekit-react/tailwind.config.ts
```diff
@@ -11,6 +11,7 @@ const config: Config = {
       },
     },
   },
+  plugins: [require("@tailwindcss/forms")({ strategy: "class" })],
```
lightwalker-eth, line 14: A few questions here:

1. Given how you've moved this dependency into `namekit-react`, should we update this documentation in the README.md file in the root of our monorepo? https://github.com/namehash/namekit/blob/85c83708b058ea383457faba1001a1de309967bf/README.md?plain=1#L518-L522 For example, exactly who needs to follow this instruction? Is it anyone building an app that adds namekit-react as a dependency? Or does this instruction only need to be followed if you import specific UI components from namekit-react? Etc.. Appreciate if you could write out a lot of details about this to make it more clear. I don't understand this currently.
2. Can you help me understand more specifically why we need this plugin at all? I don't mind adding a plugin if it only impacts code directly in our own package. What I don't like is when we do things that cause unstandard complexity for everyone else who might use our package. For example, could we effectively replace the use of this plugin by just manually adding more classnames to the related form UI components?

packages/nameguard-react/src/components/SettingsModal/SettingsModal.tsx
```diff
@@ -73,12 +73,10 @@ export const SettingsModal = () => {
                     <div className="space-y-4">
                       <div className="relative flex items-start">
                         <div className="flex h-6 items-center">
-                          <input
+                          <Checkbox
```
lightwalker-eth, line 76: A few UX suggestions here:

CleanShot 2024-10-30 at 23 13 28

In the screenshot above, let's please think about 3 regions related to each checkbox:
1. The checkbox
2. The space between the checkbox and the label
3. The label

Currently, the logic is working like this:

1. Checkbox:
   - Hover: show hover state on checkbox
   - Hover cursor: Show "clicky hand"
   - Click: toggle checkbox selected state

2. The space between the checkbox and the label:
   - Hover: Nothing
   - Hover cursor: Nothing
   - Click: Nothing

3. The label
   - Hover: show hover state on checkbox
   - Hover cursor: Nothing
   - Click: toggle checkbox selected state

Please note how all 3 of these are working differently currently.

I believe all 3 should work exactly the same. All 3 should follow these rules:
- Hover: show hover state on checkbox
- Hover cursor: Show "clicky hand"
- Click: toggle checkbox selected state

packages/namekit-react/src/components/Checkbox.tsx
```diff
@@ -0,0 +1,15 @@
+import cc from "classcat";
+import React, { InputHTMLAttributes } from "react";
+
+export const Checkbox = (props: InputHTMLAttributes<HTMLInputElement>) => {
```
lightwalker-eth, line 4: I shared other comments on this PR related to making the label associated with the <input element somehow part of the code for this new `<Checkbox ...` component.

What do you think about making it so that we make it so that the content of the label is passed as children to `<Checkbox ...`? The goal in sharing this suggestion is to try to keep apps that use our `Checkbox` component the freedom to fully customize the styling or other details of the label without making the params for `Checkbox` really complicated. Maybe this is a flawed idea I'm sharing? Appreciate your advice :+1:

lightwalker-eth (changes requested):
@eduramme Hey appreciate these updates. Reviewed and shared feedback :+1:

apps/storybook.namekit.io/stories/Namekit/Checkbox.stories.tsx
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import { Meta, StoryObj } from "@storybook/react";
+import { Checkbox } from "@namehash/namekit-react";
+
+const meta: Meta<typeof Checkbox> = {
+  title: "UI/Checkbox",
+  component: Checkbox,
+  argTypes: {
+    disabled: { control: "boolean" },
+  },
```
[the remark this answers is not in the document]
eduramme: @lightwalker-eth Done

apps/storybook.namekit.io/stories/Namekit/Checkbox.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+};
+
+export const DisabledChecked: Story = {
+  args: {
+    checked: true,
+    disabled: true,
+  },
+};
+
```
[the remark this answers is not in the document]
eduramme: @lightwalker-eth done

packages/namekit-react/package.json
```diff
@@ -55,6 +55,8 @@
     "react-dom": "18.3.1",
     "tailwindcss": "3.4.13",
     "tsup": "8.3.0",
+    "@types/node": "22.7.4",
```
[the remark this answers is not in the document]
eduramme: @lightwalker-eth humm, not really. I only added it to resolve an error, but it isn't not showing up anymore. I'm glad to remove it 😃

packages/namekit-react/src/components/Checkbox.tsx
```diff
@@ -0,0 +1,15 @@
...
+  return (
+    <input
+      type="checkbox"
+      {...props}
+      className={cc([
+        "nk-form-checkbox nk-h-4 nk-w-4 nk-rounded nk-border-[1.25px] nk-border-gray-300 nk-text-black nk-ring-0 nk-cursor-pointer checked:nk-ring-0 focus:nk-ring-0 focus:nk-ring-transparent focus:nk-outline-none focus-visible:nk-border-gray-400 active:nk-ring-0 enabled:hover:nk-border-gray-400 disabled:nk-cursor-not-allowed disabled:nk-opacity-50 disabled:nk-bg-gray-100 disabled:checked:nk-bg-gray-300",
```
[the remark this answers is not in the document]
eduramme: @lightwalker-eth makes sense

packages/nameguard-react/src/components/SettingsModal/SettingsModal.tsx
eduramme: @lightwalker-eth Nice! done

packages/namekit-react/src/components/Checkbox.tsx
eduramme: @lightwalker-eth Sounds good! Done it. The only thing that might be confusing is that we're not directly telling the user it will be rendered as a label, but I think it’s fine since it simplifies things

packages/namekit-react/tailwind.config.ts
eduramme: 1. This instruction only needs to be followed if the user imports our `Input` component from `namekit-react`. Other components, like our checkbox, are already compatible with the `@tailwindcss/forms` reset, so no additional configuration is needed for them. However, if the user imports the `Input` component and applies a global CSS reset (such as `@tailwindcss/forms` or similar), it may cause unintended styling issues with the `Input` component.
<br>To avoid these issues, users who import our `Input` component should follow the instructions to configure `@tailwindcss/forms` with the class strategy, or alternatively, we could add the reset class directly to the `Input` component to ensure it displays correctly in any setup.

2. The `@tailwindcss/forms` plugin is basically a CSS reset focused on form elements. Browsers come with their own default styles for things like checkboxes, inputs, and textareas, which can make it tough to get a consistent look with Tailwind. This plugin wipes those defaults, making it easier to style form components the way we want.
<br>We added `@tailwindcss/forms` to `namekit-react` because our checkbox component already depends on that reset to look right. But our `Input` component doesn’t account for this reset yet. So, if someone using our package applies a global CSS reset on form elements, it can make our `Input` look weird.
<br>One way to handle this and get rid of the warning is to add the `@tailwindcss/forms` reset class directly to our `Input` component. That way, even if the user has a global CSS reset, our `Input` won’t be affected.
<br>This isn’t just a `@tailwindcss/forms` issue, it could happen with any CSS reset. We just noticed it specifically with `@tailwindcss/forms` because we were using it globally on Nameguard, which made the `Input` component look weird in forms there.


To wrap things up, this is a very specific issue, it only occurs if the user imports the `Input` component from `namekit-react` and has a CSS reset applied globally to form elements. Given how specific this is, I believe we have three options:

A: Remove the comment and let the user figure it out independently if that rare scenario happens.
B: Add the `@tailwindcss/forms` reset to our `Input` component.
C: Include a more generic warning, such as:
"This package includes custom styles that may be affected by other CSS resets or frameworks (e.g., Tailwind CSS forms). If you encounter unexpected styling issues, consider checking for conflicts with any global or opinionated CSS resets in your project."

packages/namekit-react/src/components/Checkbox.tsx
```diff
@@ -1,15 +1,40 @@
 import cc from "classcat";
 import React, { InputHTMLAttributes } from "react";
 
-export const Checkbox = (props: InputHTMLAttributes<HTMLInputElement>) => {
+export const Checkbox = ({
```
lightwalker-eth, line 4: @eduramme Hey can you please refine the logic for the click area / special mouse cursor makes sense if there's no text associated with the checkbox?

For example, in the Default story, there's a huge area that toggles the checkbox that shouldn't.

CleanShot 2024-11-06 at 16 59 40

packages/namekit-react/tailwind.config.ts
eduramme: That's what happens with our contact form, for example
image

lightwalker-eth (changes requested):
@eduramme Thanks for improvements. Reviewed and shared a few suggestions :+1:

packages/namekit-react/tailwind.config.ts
eduramme: @lightwalker-eth I’ve added a few reset classes to our Input component and removed the warning about Tailwind forms.

https://github.com/namehash/namekit/pull/453/commits/870fe0bb54a1bfa86b0f2ca596d794831978d3a5

packages/namekit-react/src/components/Checkbox.tsx
eduramme: @lightwalker-eth done

packages/nameguard-react/src/components/SettingsModal/SettingsModal.tsx
```diff
@@ -71,112 +71,75 @@ export const SettingsModal = () => {
...
+                        name="attempt-ens-normalization"
+                        checked={localSettings.attemptEnsNormalization}
+                        onChange={(e) =>
+                          setLocalSettings((prev) => ({
+                            ...prev,
+                            attemptEnsNormalization: e.target.checked,
+                          }))
+                        }
+                      >
+                        <p>Attempt normalization</p>
```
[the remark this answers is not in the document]
eduramme: The children prop is not restricted to a single node. It can be a single node, multiple nodes, or even a string. This flexibility allows for more dynamic and composable UI structures. Having two node elements as children of the Checkbox component is perfectly fine.

packages/namekit-react/src/components/Checkbox.tsx
```diff
@@ -1,15 +1,40 @@
...
+          id={id}
+          disabled={disabled}
+          {...props}
+          className={cc([
+            "nk-form-checkbox disabled:nk-cursor-default nk-h-4 nk-w-4 nk-rounded nk-border-[1.25px] nk-border-gray-300 nk-text-black nk-cursor-pointer checked:nk-ring-0 focus:nk-ring-0 focus:nk-ring-transparent focus:nk-outline-none focus-visible:nk-border-gray-400 enabled:hover:nk-border-gray-400 disabled:nk-opacity-50 disabled:nk-bg-gray-100 disabled:checked:nk-bg-gray-300",
+            className,
+          ])}
+        />
+      </div>
+      <div
```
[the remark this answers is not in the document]
eduramme: @lightwalker-eth Any styles the user applies to the label will replace the default ones we set. Adding classes to the checkbox’s children can make things easier, especially if the user just passes a simple string (which is probably what most people will do).

Also, keep in mind that the label and checkbox are separate HTML elements. If the user wants, they can create a custom setup like this:

```
<Checkbox id="checkbox-id" />
<label htmlFor="checkbox-id" />
```

The main benefit of including the label inside the checkbox component is that it’s already styled, so the user doesn’t have to think much about it (it just works out of the box).

Makes sense?

packages/nameguard-react/src/components/SettingsModal/SettingsModal.tsx
```diff
@@ -71,112 +71,75 @@ export const SettingsModal = () => {
...
+                        checked={localSettings.trimWhitespace}
+                        onChange={(e) =>
+                          setLocalSettings((prev) => ({
+                            ...prev,
+                            trimWhitespace: e.target.checked,
+                          }))
+                        }
+                      >
+                        <p>Trim whitespace</p>
+                        <p className="text-gray-500 text-sm leading-5 font-normal">
```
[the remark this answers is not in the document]
eduramme: image

Sounds good. We will have that empty space, that will be covered by some padding or margin. Do you think it's a bad idea to create a "description" prop for the Checkbox component?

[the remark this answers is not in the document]
eduramme: This way, we could ensure that text is alined with the label

lightwalker-eth (approved):
@eduramme Looks good :+1: Thank you for the updates!

PR #462, opened 2024-11-07
feat: [sc-25834] Remove headless-ui tooltip dependency

eduramme:
Story details: https://app.shortcut.com/ps-web3/story/25834


notrab (approved):
Approving but the title says it’s removing react-tooltip, even though it adds it. I expect this is just a mistake for headless-ui tooltip.

PR #474, opened 2024-12-03
feat: add groupedByCategory endpoint to NameGraph class

FrancoAguzzi:
This PR addresses the addition of the following items:

- `/grouped-by-category` NameGraph endpoint
- All interfaces related to it

The interfaces were mirrored from the following references:
- http://100.24.45.225/docs
- https://github.com/namehash/name-generator/blob/07e7117031ad5189f7babf158a112507234829ce/conf/prod_config_new.yaml#L84
- https://github.com/namehash/namehash/blob/main/lib/api/generate-names.ts

This PR is blocking the creation of an endpoint for `/suggestions-by-category` as these share the same deps


packages/namegenerator-sdk/src/index.ts
```diff
@@ -0,0 +1,144 @@
+import fetch from "cross-fetch";
+
+const DEFAULT_ENDPOINT = "http://100.24.45.225/";
+
+export const NameGeneratorGroupingCategory = {
```
FrancoAguzzi, line 5: @lightwalker-eth 

This definition is not required: in the docs, this is defined as "string". I wanted to list all possible grouping categories by using this strategy, which can be removed if you think it will be messy to update this in the future based on NameGenerator API updates.

FrancoAguzzi: Captura de Tela 2024-12-03 às 14 06 46

For reference

packages/namegraph-sdk/package.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "@namehash/namegraph-sdk",
+  "description": "A lightweight JavaScript client for the namegraph API.",
```
lightwalker-eth, line 3:
```suggestion
  "description": "A lightweight JavaScript client for the NameGraph API.",
```

lightwalker-eth (approved):
@FrancoAguzzi Hey looks good. Let's ship it. Approved 🚀

PR #492, opened 2024-12-13
feat: [sc-25979] Migrate NameGuard from serverless to Terraform deployment

BanaSeba:
Story details: https://app.shortcut.com/ps-web3/story/25979


apps/api.nameguard.io/package.json
```diff
@@ -1,29 +0,0 @@
...
-    "NameKit",
-    "NameGuard",
-    "NameHash"
-  ],
-  "scripts": {
-    "deploy:prod": "DOCKER_CONFIG=.docker serverless deploy --stage prod",
-    "deploy:staging": "DOCKER_CONFIG=.docker serverless deploy --stage staging"
-  },
-  "dependencies": {
-    "serverless": "3.39.0"
```
lightwalker-eth, line 24: @BanaSeba suggest undoing the delete of this file. And then navigating to this directory and running `pnpm remove serverless`. We need PNPM to remove this dependency. The lock file in the root of the monorepo should be getting updated in this PR too.

apps/api.nameguard.io/package.json
```diff
@@ -1,29 +0,0 @@
...
-  },
-  "devDependencies": {
-    "serverless-prune-plugin": "2.0.2"
```
lightwalker-eth, line 27: @BanaSeba Please see comment above. We also need to ask PNPM to remove this dependency too. We don't remove it just by deleting this file.

lightwalker-eth (changes requested):
@BanaSeba Hey great to see this. Reviewed and shared feedback :+1:

apps/api.nameguard.io/package.json
BanaSeba: Updated pnpm as well

apps/api.nameguard.io/package.json
BanaSeba: Updated pnpm as well

lightwalker-eth (approved):
@BanaSeba Excited to say goodbye to Serverless and these outdated dependencies 🚀 Thank you! 🙌

PR #499, opened 2024-12-19
✨ Integrate NameGraph SDK API call into Explore Collections page and make collections exploring page work

FrancoAguzzi:


apps/namegraph.dev/tailwind.config.ts
```diff
@@ -1,25 +1,47 @@
...
 const config: Config = {
-    darkMode: ["class"],
-    content: [
+  darkMode: ["class"],
+  content: [
     "./pages/**/*.{js,ts,jsx,tsx,mdx}",
     "./components/**/*.{js,ts,jsx,tsx,mdx}",
     "./app/**/*.{js,ts,jsx,tsx,mdx}",
   ],
+  safelist: [
```
lightwalker-eth, line 10: Can you help me understand the goal here?

apps/namegraph.dev/components/mini-apps/explore-collections/suggestion-category-header.tsx
```diff
@@ -0,0 +1,22 @@
+import Skeleton from "@/components/skeleton";
+import { NameGraphFetchTopCollectionMembersResponse } from "@namehash/namegraph-sdk/utils";
+
+interface SuggestionCategoryHeaderProps {
+  category: NameGraphFetchTopCollectionMembersResponse | null;
```
lightwalker-eth, line 5: Please always document what the `null` case represents

apps/namegraph.dev/components/mini-apps/explore-collections/recursive-related-collection-pills.tsx
```diff
@@ -0,0 +1,52 @@
...
+  recursiveRelatedCollections,
+}: RecursiveRelatedCollectionPillsProps) => {
+  if (recursiveRelatedCollections.length === 0) {
+    return null;
+  }
+
+  const RelatedCollectionPillsSkeleton = (
+    <div className="ml-1 flex flex-wrap items-center text-sm mt-3 gap-2">
+      <p className="font-light min-w-[128px]">Check out related collections:</p>
+      {Array(3).map((idx) => (
```
lightwalker-eth, line 20: Avoid magic numbers. Always define a named constant.

packages/namegraph-sdk/src/utils.ts
```diff
@@ -71,29 +71,25 @@ export type NameGraphSuggestion = {
...
-};
-
 export type NameGraphRelatedCollectionResponse = {
   collection_id: string;
   collection_title: string;
   collection_members_count: number;
 };
 
 export type NameGraphGroupedByCategoryResponse = {
-  categories: NameGraphCategory[];
```
lightwalker-eth, line 85: Can you please confirm 100% that you validated that these API interface changes are correct?

packages/namegraph-sdk/src/utils.ts
```diff
@@ -71,29 +71,25 @@ export type NameGraphSuggestion = {
...
-  all_tokenizations: [];
+  categories: NameGraphFetchTopCollectionMembersResponse[];
+  all_tokenizations: string[][];
 };
 
 export type NameGraphFetchTopCollectionMembersResponse = {
   suggestions: NameGraphSuggestion[];
   name: string;
   type: string;
-  collection_id: string;
```
lightwalker-eth, line 93: Can you please confirm 100% that you validated that these API interface changes are correct?

apps/namegraph.dev/components/mini-apps/explore-collections/suggestion-category.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+    return () => {
+      window.removeEventListener(
+        "resize",
+        lodash.debounce(setCategoryDimensions, 100),
+      );
+    };
+  }, []);
+
+  const customizedPillsColors = [
```
lightwalker-eth, line 129: Each time I scroll down the colors are changing. This is confusing to users because they naturally then try to figure out if there's some symbolic meaning for a given color. As I understand, there's no symbolic meaning to these colors. Therefore suggest to completely remove the "dynamic rainbow" effect. If there's a goal for this please help me understand what it is. Thanks.

apps/namegraph.dev/lib/utils.ts
```diff
@@ -1,6 +1,126 @@
...
 export function cn(...inputs: ClassValue[]) {
   return twMerge(clsx(inputs));
 }
+
+interface LabelAndModeQueryParams {
+  label: string;
+  mode: "full";
+}
+
+export interface NameGraphRelatedCollection {
```
lightwalker-eth, line 19: Is there a reason this isn't in the SDK? Is it a part of the API?

apps/namegraph.dev/lib/utils.ts
```diff
@@ -1,6 +1,126 @@
...
+  collection_id: string;
+  collection_title: string;
+  collection_members_count: number;
+}
+
+export interface NameGraphSuggestionCategory {
```
lightwalker-eth, line 25: Is there a reason this isn't in the SDK? Is it a part of the API?

apps/namegraph.dev/lib/utils.ts
```diff
@@ -1,6 +1,126 @@
...
+  type: NameGraphGroupingCategory;
+  // NameGraph API guarantees "name" will always be a normalized name
+  name: string;
+  // NameGraph API guarantees "suggestions" will never be an empty list
+  suggestions: NameGraphSuggestion[];
+  related_collections: NameGraphRelatedCollection[];
+}
+
+// This array is used for checking when a category is not a "related" category
+const NameGraphSuggestionCategoryTypes = Object.values(
```
lightwalker-eth, line 35: Seems this idea should be moved into the SDK?

apps/namegraph.dev/lib/utils.ts
```diff
@@ -1,6 +1,126 @@
...
+  NameGraphGroupingCategory,
+  NameGraphSuggestion,
+} from "@namehash/namegraph-sdk/utils";
+import { createNameGraphClient } from "@namehash/namegraph-sdk";
 
 export function cn(...inputs: ClassValue[]) {
   return twMerge(clsx(inputs));
 }
+
+interface LabelAndModeQueryParams {
```
lightwalker-eth, line 14: I'm confused here. It looks like we're smashing together two separate goals that should be seprated.

1. The definition of an interface associated with a NameGraph API query.
2. Specific constants that are sent in our implementation of this query within our example app.

As i understand it, (1) above should live in the SDK. (2) should live inside of this app, but it should not be achieved by defining a NEW interface. Instead it should be achieved by making reference to the interface defined in (1).

apps/namegraph.dev/lib/utils.ts
```diff
@@ -1,6 +1,126 @@
...
+  input: LabelAndModeQueryParams,
+): Promise<NameGraphSuggestion[]> => {
+  if (input.label.includes("."))
+    throw new Error("Invalid label for generating name suggestions");
+
+  // Related collections params
+  const enable_learning_to_rank = true;
+  const max_names_per_related_collection = 10;
+  // Determines the number of collections returned in category.related_collections
+  const max_recursive_related_collections = 3;
```
lightwalker-eth, line 57: Please note: I shared another comment in another file how you need to use named constants and not magic numbers. You defined a magic number 3 somewhere and that named constant should be the same as this named constant. They are the same idea!

apps/namegraph.dev/lib/utils.ts
```diff
@@ -1,6 +1,126 @@
...
+ */
+const NameGraphClient = createNameGraphClient();
+
+/**
+ * Main function for fetching name suggestions from the NameGraph API.
+ */
+export const generateNamesByQuery = async (
+  input: LabelAndModeQueryParams,
+): Promise<NameGraphSuggestion[]> => {
+  if (input.label.includes("."))
```
lightwalker-eth, line 50: Why are we using this strategy? Why aren't we implementing the same strategy as before where if the input includes "." that we internally process that so as to only use the childmost label for our query?

Here we should identify back to the user the exact label that we're using to generate ideas with. Please note the related UX for this that we implemented here: https://www.namerank.io/?name=example.name.with.multiple.labels

CleanShot 2024-12-20 at 17 29 48

1. The users raw input query, which may include any number of labels.
2. We show back to the user the label we're actually going to use in the query.
3. We show the result of the query.

Please study the UX and implementation referenced above. We should follow the same pattern here.

lightwalker-eth (approved):
@FrancoAguzzi Hey, appreciate the updates here. Exciting to see. I've approved this PR so that you can merge it, but please carefully review my feedback and either resolve it in this PR, or create a separate PR that resolves it. Thank you!

apps/namegraph.dev/tailwind.config.ts
FrancoAguzzi: Yes! Safelist makes it possible to generate programatically the bg colors used by the name ideas pills. 

It seems like Tailwind runs in a different "time-space" (😅) than programatic JS, which can be attested by removing the "safelist" from Tailwind config: when we do so, the bg-[#SOMECOLOR] is not applied. Safelisting in Tailwind means including this class in Tailwind's bundle upfront, guaranteeing the classes inside "safelist" array will have corresponding Tailwind styling config present in the CSS bundle. 🙂 Does this help?

packages/namegraph-sdk/src/utils.ts
FrancoAguzzi: Yes! 💯

packages/namegraph-sdk/src/utils.ts
FrancoAguzzi: This is aligned!

apps/namegraph.dev/components/mini-apps/explore-collections/suggestion-category.tsx
FrancoAguzzi: I used to like how the colors would change, subjectively representing the numerous possibilities of results our NameGraph can result in. I did the requested change, though, as I understand the new perspective shared!

apps/namegraph.dev/lib/utils.ts
FrancoAguzzi: This was actually a duplicate, nice you find it

apps/namegraph.dev/lib/utils.ts
FrancoAguzzi: Same as above

apps/namegraph.dev/lib/utils.ts
FrancoAguzzi: As this is a representation of the same NameGraph SDK present utility, but, with a different data structure, this is very much related to clients. I say this because apps might want to use this is different ways and mapping from one data structure to another one may be a responsibility of the client side, I think, so I wouldn't migrate this, what do you think?

apps/namegraph.dev/lib/utils.ts
FrancoAguzzi: Now following the same pattern of using the childmost label for search. Mini-app is also displaying both the complete and the query-used strings, alongside with the results!

PR #511, opened 2024-12-30
Fix tests - Subgraph stopped resolving some names

djstrong:


lightwalker-eth (approved):
@djstrong Thank you :+1:

PR #512, opened 2024-12-30
Get log probability color

djstrong:


apps/namerank.io/app/indicator.tsx
```diff
@@ -16,9 +16,26 @@ export const Indicator = ({ value }: IndicatorProps) => {
     }
   };
 
+  const getLogProbabilityColor = (val: number): [number, number, number] => {
```
lightwalker-eth, line 19: @notrab Hey, could you please help to advance next steps here? The big picture is that this function is attempting to generate a RGB hex code that can be used to dynamically determine the color of an indicator based on a `log_probability` value. I asked @djstrong to focus on the math side of this logic so that we could then pass this goal off to you for refinement to align with TypeScript / Tailwind / CSS best practices. Please let us know if any questions!

@notrab Requested enhancements on this for you to lead include:
1. Rename the `val` param to something approximating `log_probability` to be more self documenting
2. Refine the return type to better align with best practices for how a color code might be returned.

lightwalker-eth (changes requested):
@djstrong Hey great to see this!

@notrab Hey, could you please take the lead on the next steps for this PR? Thank you!

apps/namerank.io/app/indicator.tsx
```diff
@@ -1,24 +1,37 @@
...
-      return "bg-yellow-500";
-    } else {
-      return "bg-red-500";
-    }
-  };
+export const Indicator = ({ log_probability }: IndicatorProps) => {
+  const normalizedPercentage = calculateNormalizedPercentage(log_probability);
+  const [r, g, b] = getRGBColor(normalizedPercentage);
+
+  const width = Math.max(50, 128 * normalizedPercentage);
```
lightwalker-eth, line 29: Lets define a named constant for 50 and 128 to make this code more self documenting. As I understand they represent min and max width of the indicator?

Suggest we make the min a bit less. Maybe only 20?

lightwalker-eth (changes requested):
@notrab Overall looks good! Just 1 change request

apps/namerank.io/app/indicator.tsx
notrab: Done!

lightwalker-eth (approved):
@notrab Thanks for update! Approved 🚀

PR #513, opened 2024-12-30
feat: [sc-25983] Prepare NameRank for transition into Monorepo

Byczong:
Story details: https://app.shortcut.com/ps-web3/story/25983

For now, `namerank` is placed under `apps/api.namerank.io/`.

Later on, the python package from this folder should be extracted and placed under `packages/namerank-python/`. This will require path changes in `.github/workflows` and `.pre-commit-config.yaml`.


.github/workflows/namerank-api-lambda-deploy.yml
```diff
@@ -0,0 +1,138 @@
...
+          ALCHEMY_URI_SEPOLIA: ${{ secrets.ALCHEMY_URI_SEPOLIA }}
+          ENS_SUBGRAPH_URL_MAINNET: ${{ secrets.ENS_SUBGRAPH_URL_MAINNET }}
+          ENS_SUBGRAPH_URL_SEPOLIA: ${{ secrets.ENS_SUBGRAPH_URL_SEPOLIA }}
+          AWS_REGION: ${{ secrets.AWS_REGION }}
+          CERTIFICATE_NAME: ${{ secrets.CERTIFICATE_NAME }}
+          HOSTED_ZONE_NAME: ${{ secrets.HOSTED_ZONE_NAME }}
+        run: |
+          if [[ ${{ github.ref }} == 'refs/heads/main' ]]; then
+          STAGE="prod"
+          DOMAIN_NAME=${{ secrets.PROD_DOMAIN_NAME }}
```
djstrong, line 68: @BanaSeba all these files are moved from namerank repo. I guess `secrets.PROD_DOMAIN_NAME` and `secrets.STAGING_DOMAIN_NAME` should be changed?

apps/api.namerank.io/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM --platform=linux/arm64 public.ecr.aws/lambda/python:3.11
```
Carbon225, line 1: It would make sense to move the responsibility of setting `arm64` to the deployment script.

apps/api.namerank.io/pyproject.toml
```diff
@@ -0,0 +1,61 @@
...
+fastapi = "^0.115.4"
+mangum = {version = "^0.17.0"}
+httpx = "^0.25.0"
+python-dotenv = "^1.0.0"
+pyahocorasick = "^2.0.0"
+setuptools = "^75.1.0"
+
+
+[tool.poetry.extras]
+namerank = ["mangum"]
```
Carbon225, line 27: In nameguard this extra is called "lambda". Shouldn't it be called the same here? So you would do `pip install namerank[lambda]` instead of `pip install namerank[namerank]`.

apps/api.namerank.io/LICENSE
```diff
@@ -0,0 +1 @@
+TODO
```
lightwalker-eth, line 1: Please copy the exact same LICENSE file we've added to other apps here :+1: Thanks

apps/api.namerank.io/Dockerfile
BanaSeba: Are we ever using any other platform than arm64? I would expect that any testing/deployment for this lambda will be done only for arm64

.github/workflows/namerank-api-lambda-deploy.yml
BanaSeba: I've added env variables to Github actions and created PR with updating this script to use them: https://github.com/namehash/namekit/pull/522

lightwalker-eth: @BanaSeba Should we also make these environment variables / secret names more specific? Worried that we might be using the same values for both NameGuard and NameRank, when they should be different.

lightwalker-eth (changes requested):
@Byczong Hey awesome to see this step 🚀 Reviewed and shared some feedback 👍 @BanaSeba Also a question for you in my feedback.

.github/workflows/namerank-api-lambda-deploy.yml
BanaSeba: I've added a NAMERANK prefix before each of those specific variables. For now, this should be sufficient

apps/api.namerank.io/pyproject.toml
Byczong: I changed it, but also made the `mangum` optional, as the extra would be obsolete otherwise.

This caused the tests from `test_api.py` to fail in the ci, so I changed it to `poetry install --extras "lambda"` in the workflow.

lightwalker-eth (approved):
@Carbon225 @Byczong @djstrong @BanaSeba Looks awesome 🚀 Approved!

PR #515, opened 2024-12-31
feat: [sc-25972] Refine NameRank SDK

Carbon225:
Story details: https://app.shortcut.com/ps-web3/story/25972

### DIFF
--- a/packages/namerank-sdk/src/index.ts
+++ b/packages/namerank-sdk/src/index.ts
@@ -1,5 +1,6 @@
 import fetch from "cross-fetch";
 import { Network, NameGuardReport } from "@namehash/nameguard";
+import { InspectorResult } from "./labelinspector";
 
 const ETH_TLD = "eth";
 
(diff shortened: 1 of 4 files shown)


packages/namerank-sdk/src/index.ts
```diff
@@ -9,24 +9,345 @@ export enum LabelStatus {
...
+  /** The normalization status of the label */
   status: LabelStatus;
+
+  /** The probability of the label */
   probability: number;
+
+  /** The natural logarithm of the probability */
   log_probability: number;
+
+  /** The number of words in the label */
```
lightwalker-eth, line 325: Can we please document more details about this `word_count` field? For example, is it sometimes 0 or undefined? What happens if we can't find any words? And when we do return a positive `word_count` value, how does this value relate to the `top_tokenization` field?

packages/namerank-sdk/src/index.ts
```diff
@@ -9,24 +9,345 @@ export enum LabelStatus {
...
   word_count: number;
+
+  /** The most likely tokenization of the label */
   top_tokenization?: string[];
+
+  /** All possible tokenizations of the label */
   tokenizations: Record<string, any>[];
```
lightwalker-eth, line 332: Let's please also define an interface for the value of each of these records. We should avoid the use of `any`.

lightwalker-eth (changes requested):
@Carbon225 Hey great to see the updates here. Reviewed in detail and shared feedback :+1: @djstrong FYI

packages/namerank-sdk/src/index.ts
```diff
@@ -9,24 +9,345 @@ export enum LabelStatus {
   Unknown = "unknown",
 }
 
+// Character and grapheme types can potentially evolve independently.
+
+export enum CharacterType {
```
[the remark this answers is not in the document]
djstrong: None of these are present in NameGuard SDK.

djstrong:
@lightwalker-eth I have addressed all issues.

notrab:
@djstrong @lightwalker-eth we should add a changesets entry so the types are released as a minor update.

packages/namerank-sdk/src/index.ts
```diff
@@ -81,7 +459,7 @@ export class NameRank {
    *
    * @param {string} name The name for NameRank to inspect.
    * @param {InspectNameOptions} options The options for the inspection.
-   * @returns {Promise<NameRankReport>} A promise that resolves with the `NameRankReport` of the name.
+   * @returns {Promise<NameRankResponse>} A promise that resolves with the `NameRankResponse` of the name.
    * @example
    * const nameRankReport = await namerank.inspectName('vitalik.eth');
```
lightwalker-eth, line 464:
```suggestion
   * const nameRankResponse = await namerank.inspectName('vitalik.eth');
```

packages/namerank-sdk/src/index.ts
```diff
@@ -9,24 +9,399 @@ export enum LabelStatus {
...
+
+export type InspectorResult = InspectorResultNormalized | InspectorResultUnnormalized;
+
+/** 
+ * Represents a single tokenization result with its probability scores
+ */
+export interface Tokenization {
+  /** The tokens that make up this tokenization */
+  tokens: string[];
+  /** The probability score for this tokenization, between 0 and 1 */
```
lightwalker-eth, line 318: Suggest we reuse the more formal definition of `probability` from `NLPLabelAnalysis`.

packages/namerank-sdk/src/index.ts
```diff
@@ -9,24 +9,399 @@ export enum LabelStatus {
...
+  probability: number;
+  /** The natural log of the probability score, <= 0 */
```
lightwalker-eth, line 320: Suggest we also update this to match `log_probability` in `NLPLabelAnalysis`.

packages/namerank-sdk/src/index.ts
```diff
@@ -9,24 +9,399 @@ export enum LabelStatus {
...
+  log_probability: number;
+}
+
 export interface NLPLabelAnalysis {
-  inspection: any;
+  /** The result of the label inspection */
```
lightwalker-eth, line 325: Could we also document here that `inspection` will be of type `InspectorResultNormalized` if `status` is `Normalized` and otherwise will be of type `InspectorResultUnnormalized`?

What if `status` is `Unknown`?

packages/namerank-sdk/src/index.ts
```diff
@@ -9,24 +9,399 @@ export enum LabelStatus {
...
+   */
   log_probability: number;
+
+  /**
+   * The minimum number of words across all valid tokenizations of the label that contain no gaps.
+   * Will be 0 if no valid tokenization without gaps is found. For example, labels containing only
+   * numbers or special characters may have a word_count of 0.
+   * Note: this is not the number of words in the top_tokenization, but the minimum number of words 
+   * across all valid tokenizations without gaps.
+   * Must be greater than or equal to 0.
```
lightwalker-eth, line 353:
```suggestion
   * Always a non-negative integer.
```

packages/namerank-sdk/src/index.ts
```diff
@@ -9,24 +9,399 @@ export enum LabelStatus {
...
+  /**
+   * Score indicating how interesting/memorable the name is. For single labels, returns the score directly. For 2-label names (e.g., "nick.eth"), returns the 
+   * score for the first label ("nick"). For 3 or more labels, returns 0. If the label is not inspected, 
+   * this field will be 0. The score ranges from 0.0 to 1.0 inclusive, where 0.0 indicates least interesting 
+   * and 1.0 indicates most interesting.
+   */
   interesting_score: number;
+
+  /**
+   * The result of the NLP analysis on the label. If the label is not inspected, this field will be undefined.
```
lightwalker-eth, line 395: Could we refine this to make it very explicit what it means for the label not to be inspected? What field / field value would define this case?

packages/namerank-sdk/src/index.ts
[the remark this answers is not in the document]
lightwalker-eth: @djstrong Ah, I see. Am I correct to assume that these ideas are actually related to the LabelInspector? Assuming so, what do you think about creating a separate labelinspector.ts file that you could move those definitions into, and then importing those definitions back into this file.

My goal here is to keep this file more "pure" to NameRank.

I don't suggest we create a separate typescript package for a LabelInspector SDK in this PR. This would be a nice goal for a separate PR, where we might move this labelinspector.ts file into a separate package, but for now just creating a new and separate file in this current package is ok :+1:

Appreciate your advice!

lightwalker-eth (changes requested):
@djstrong Hey super updates! Thank you. Reviewed and shared feedback :+1:

lightwalker-eth (approved):
@djstrong Great updates! Reviewed and approved 🚀

PR #520, opened 2025-01-02
feat: use default namerank api endpoint

notrab:
Additionally removes `console.log` inside the client. Logging is useful, so we should add the ability to do better logging in all our SDKs with some kind of `debug` flag.


lightwalker-eth (approved):
@notrab Looks good, thanks :+1:

PR #523, opened 2025-01-04
feat: [sc-25941] Custom API Root Path

Goader:


packages/nameguard-sdk/src/index.ts
```diff
@@ -606,10 +606,13 @@ export class NameGuard {
 
     const network = this.network;
 
-    const url = new URL(
-      `inspect-namehash/${network}/${namehash}`,
-      this.nameguardEndpoint,
-    );
+    // Create base URL and ensure it ends with a trailing slash
```
lightwalker-eth, line 609: A few suggestions here:

1. Move the relevant URL logic into a separate dedicated function that's designed to be easy to unit test, and then writing some unit tests for it to verify it does what we expect.
2. If our goal is to ensure something is true about the string for the NameGuard endpoint we use, then suggest we just do that once in the constructor and not repeat it with each API call.

CleanShot 2025-01-07 at 14 26 07

In other words, it seems we should move the call to this new utility function into the constructor and remove it from here.

packages/nameguard-sdk/src/index.ts
```diff
@@ -775,7 +778,15 @@ export class NameGuard {
     body: object = {},
     headers: object = {},
   ): Promise<any> {
-    const url = new URL(path, this.nameguardEndpoint);
+    // Create base URL and ensure it ends with a trailing slash
```
lightwalker-eth, line 781: Please see my related comment above on logic for building these nameGuardEndpoint values.

packages/nameguard-sdk/src/index.ts
```diff
@@ -775,7 +778,15 @@ export class NameGuard {
...
+    const baseUrl = new URL(this.nameguardEndpoint.toString());
+    if (!baseUrl.pathname.endsWith('/')) {
+      baseUrl.pathname += '/';
+    }
+
+    // Combine base URL with path, removing any leading slash from path
+    const url = new URL(path.replace(/^\/+/, ''), baseUrl);
```
lightwalker-eth, line 788: regex make me nervous when they aren't inside unit tests 😅

packages/namerank-sdk/src/index.ts
```diff
@@ -208,7 +207,14 @@ export class NameRank {
     body: object = {},
     headers: object = {},
   ): Promise<any> {
-    const url = new URL(path, this.namerankEndpoint);
+    // Create base URL and ensure it ends with a trailing slash
```
lightwalker-eth, line 210: Please see my related feedback on the nameguard-sdk.

packages/namerank-sdk/src/index.ts
```diff
@@ -187,10 +187,9 @@ export class NameRank {
     options?: InspectNameOptions,
   ): Promise<NameRankResponse> {
     const network_name = this.network;
-    // TODO: Temp fix until `api.namerank.io` is updated to return namerank api without subfolder
-    return this.rawRequest("namerank/inspect-name", "POST", {
+    return this.rawRequest("inspect-name", "POST", {
       name,
-      network_name,
+      // network_name,
```
lightwalker-eth, line 192: Can you share more background info why we're commenting this out?

lightwalker-eth (changes requested):
@djstrong @Goader Thanks for updates :+1: Reviewed and shared feedback!

packages/namerank-sdk/src/index.ts
djstrong: There is no `network_name` parameter in NameRank. Jamie added this, I guess by accident.

djstrong: @lightwalker-eth NameRank is using NameGuard where this parameter exists, but NameRank does not try to resolve labelhashes so it is not used. Maybe  I am missing something.

packages/nameguard-sdk/src/index.ts
djstrong: The problem is affecting only NameRank but solution was copied to NameGuard too.
If we provide an endpoint with subfolder without trailing slash then the subfolder was removed. So in the constructor is added only checking if this trailing slash exists.
There is no problem with slash at the beginning of endpoint names as we are controlling the names.

packages/nameguard-sdk/src/index.ts
djstrong: I have removed this code.

lightwalker-eth (approved):
@djstrong Hey looks great, reviewed and approved 🚀

PR #531, opened 2025-01-07
make NameRank top app

djstrong:


lightwalker-eth (approved):
@djstrong Nice solution, thanks 🚀

PR #536, opened 2025-01-09
rename namerank to nameai

notrab:


lightwalker-eth (approved):
@notrab Thanks for lockfile fix 🚀

PR #547, opened 2025-01-13
feat(nameai): Leaderboard Demo

notrab:
I built this to show off the `interesting_score` from NameAI:

CleanShot 2025-01-13 at 10 18 13

There's still some tweaks needed, but lmkwyt!

### DIFF
--- a/apps/nameai.dev/app/leaderboard/client.tsx
+++ b/apps/nameai.dev/app/leaderboard/client.tsx
@@ -1,0 +1,41 @@
+"use client";
+
+import { useState } from "react";
+import { nameai } from "@namehash/nameai";
+
+import { Form } from "../form";
+import { List } from "./list";
+
+interface NameItem {
+  name: string;
+  interestingScore: number;
+}
+
+export function Client({ initialNames }: { initialNames: NameItem[] }) {
+  const [names, setNames] = useState(initialNames);
+
+  const handleSubmit = async (name: string) => {
+    try {
+      const result = await nameai.inspectName(name);
+      const newItem = {
+        name,
+        interestingScore: result.nameai.interesting_score,
+      };
+      setNames((prevNames) => {
+        const updatedNames = [...prevNames, newItem];
+        return updatedNames.sort(
+          (a, b) => b.interestingScore - a.interestingScore,
+        );
+      });
+    } catch (error) {
+      console.error("Error adding name:", error);
+    }
+  };
+
+  return (
+    <>
+      <Form onSubmit={handleSubmit} />
+      <List names={names} />
+    </>
+  );
+}
--- a/apps/nameai.dev/app/leaderboard/list.tsx
+++ b/apps/nameai.dev/app/leaderboard/list.tsx
@@ -1,0 +1,26 @@
+"use client";
+
+interface NameItem {
+  name: string;
+  interestingScore: number;
+}
+
+export function List({ names }: { names: NameItem[] }) {
+  return (
+    <ul className="space-y-2">
+      {names?.map((item, index) => (
+        <li
+          key={item.name}
+          className="flex items-center justify-between bg-white p-3 rounded shadow"
+        >
+          <span className="font-bold">{item.name}</span>
+          <div className="flex items-center gap-2">
+            <span className="text-sm text-gray-600">
+              {item.interestingScore.toFixed(4)}
+            </span>
+          </div>
+        </li>
+      ))}
+    </ul>
+  );
+}
--- a/apps/nameai.dev/app/leaderboard/page.tsx
+++ b/apps/nameai.dev/app/leaderboard/page.tsx
@@ -1,0 +1,52 @@
+import { Suspense } from "react";
+import { nameai } from "@namehash/nameai";
+import { Skeleton } from "../../components/skeleton";
+import { Client } from "./client";
+
+interface NameItem {
+  name: string;
+  interestingScore: number;
+}
+
+const defaultNames = ["vitalik", "ethereum", "web3", "blockchain", "defi"];
+
+async function getInitialNames(): Promise<NameItem[]> {
+  const loadedNames = await Promise.all(
+    defaultNames.map(async (name) => {
+      const result = await nameai.inspectName(name);
+      return {
+        name,
+        interestingScore: result.nameai.interesting_score,
+      };
+    }),
+  );
+  return loadedNames.sort((a, b) => b.interestingScore - a.interestingScore);
+}
+
+export default async function LeaderboardPage() {
+  const initialNames = await getInitialNames();
+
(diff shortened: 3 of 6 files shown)


lightwalker-eth (commented):
@notrab So cool to see this 🤩 

Some general suggestions:

1. Could you please update the homepage so that it doesn't have any demo apps directly embedded into it, but instead includes links to our two separate demo apps. This will also achieve the goal of having a link from the homepage to the new demo app implemented in this PR.
2. Let's please rename words such as "leaderboard" or "rank" to "sort" or "AI sort". Suggesting this language because it balances achieving our goals for the demo while also mitigating potential anger people might have at the idea of all names not being considered equally valuable. Therefore, this should be the "sort" or "AI sort" demo, rather than "leaderboard" demo.

apps/nameai.dev/app/leaderboard/client.tsx
```diff
@@ -0,0 +1,41 @@
+"use client";
+
+import { useState } from "react";
+import { nameai } from "@namehash/nameai";
+
+import { Form } from "../form";
+import { List } from "./list";
+
+interface NameItem {
```
lightwalker-eth, line 9:
```suggestion
interface LabelItem {
```

As I understand, these will actually all be labels, not names?

I appreciate how our use of language is a bit confusing here. For example, at a UI level or branding / marketing level we might use "name" in a number of places because a lot more people will understand what a "name" is, rather than a "label". However, at a technical level, if something is actually always a single label, suggest we call it "label" rather than "name".

apps/nameai.dev/app/leaderboard/client.tsx
```diff
@@ -0,0 +1,41 @@
...
+  name: string;
```
lightwalker-eth, line 10:
```suggestion
  label: string;
```

apps/nameai.dev/app/leaderboard/client.tsx
```diff
@@ -0,0 +1,41 @@
...
+  interestingScore: number;
```
lightwalker-eth, line 11: @notrab FYI: sharing this comment and link to communicate how we'll work to achieve greater consistency in the "sort" language I suggested in another comment in a separate future PR: https://app.shortcut.com/ps-web3/story/26097/nameai-rename-interesting-score-to-sort-score

apps/nameai.dev/app/leaderboard/list.tsx
```diff
@@ -0,0 +1,26 @@
+"use client";
+
+interface NameItem {
```
lightwalker-eth, line 3: Please see related feedback I shared in a different file. Perhaps we move this definition to only a single place?

apps/nameai.dev/app/leaderboard/list.tsx
```diff
@@ -0,0 +1,26 @@
...
+  name: string;
+  interestingScore: number;
+}
+
+export function List({ names }: { names: NameItem[] }) {
```
lightwalker-eth, line 8: Please see my other comment for how these are actually labels, rather than names.

apps/nameai.dev/app/leaderboard/page.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+      <Client initialNames={initialNames} />
+    </div>
+  );
+}
+
+function LeaderboardSkeleton() {
+  return (
+    <div className="space-y-2">
+      {[...Array(5)].map((_, index) => (
```
lightwalker-eth, line 47: Could we define a named constant for the count of loading skeletons?

apps/nameai.dev/app/leaderboard/page.tsx
```diff
@@ -0,0 +1,52 @@
+import { Suspense } from "react";
+import { nameai } from "@namehash/nameai";
+import { Skeleton } from "../../components/skeleton";
+import { Client } from "./client";
+
+interface NameItem {
```
lightwalker-eth, line 6: Please see my other related comments about this in other files.

apps/nameai.dev/app/leaderboard/page.tsx
```diff
@@ -0,0 +1,52 @@
...
+  name: string;
+  interestingScore: number;
+}
+
+const defaultNames = ["vitalik", "ethereum", "web3", "blockchain", "defi"];
```
lightwalker-eth, line 11:
```suggestion
const defaultLabels = ["vitalik", "ethereum", "web3", "blockchain", "defi"];
```

apps/nameai.dev/app/leaderboard/page.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+async function getInitialNames(): Promise<NameItem[]> {
```
lightwalker-eth, line 13: Suggest renaming names to labels

apps/nameai.dev/app/leaderboard/page.tsx
```diff
@@ -0,0 +1,52 @@
...
+  return loadedNames.sort((a, b) => b.interestingScore - a.interestingScore);
+}
+
+export default async function LeaderboardPage() {
+  const initialNames = await getInitialNames();
+
+  return (
+    <div className="max-w-3xl mx-auto px-6">
+      <h1 className="text-2xl font-bold mb-4">
+        Sort Names by Interesting Score
```
lightwalker-eth, line 32: I'm happy to continue using "names" and not "labels" at a UI level / branding level :+1:

lightwalker-eth, line 32:
```suggestion
        Sort Names by AI
```

apps/nameai.dev/app/leaderboard/page.tsx
```diff
@@ -0,0 +1,52 @@
...
+      </h1>
+      <p className="mb-4">
+        We are helping people sort lists of names based on how interesting they
```
lightwalker-eth, line 35:
```suggestion
        Sort lists of names by AI-predicted relevance.
```

lightwalker-eth (changes requested):
@notrab Awesome to see this! 🚀 Reviewed and shared feedback :+1:

apps/nameai.dev/app/leaderboard/client.tsx
notrab: Noted, and updated.

apps/nameai.dev/app/leaderboard/page.tsx
notrab: This isn't actually used currently so didn't refactor, will do so.

lightwalker-eth (approved):
@notrab Hey great to see your updates :+1: Approved 🙌

PR #549, opened 2025-01-15
✨ Improve namegraph.dev

FrancoAguzzi:


packages/namegraph-sdk/src/index.ts
```diff
@@ -118,7 +119,8 @@ export class NameGraph {
     },
   ): Promise<NameGraphSuggestion[]> {
     const max_sample_size = 5;
-    const seed = options?.seed || 0;
+    const random_seed = Number(Number(Math.random() * 10).toFixed(0));
+    const seed = options?.seed || random_seed;
```
lightwalker-eth, line 123: As I understand, the client SDK shouldn't generate a random seed itself if none is provided. Instead, the client SDK should pass no `seed` at all and then the NameGraph backend will automatically generate a random seed.

This is the responsibility of the backend, not the client SDK.

Please test if this works ok and if any questions or issues please send messages with details to Krzysztof in our dev team channel so he can offer suggestions. Thanks!

packages/namegraph-sdk/src/index.ts
```diff
@@ -149,12 +151,15 @@ export class NameGraph {
...
     },
   ): Promise<NameGraphSuggestion[]> {
-    const method = "left-right-shuffle-with-unigrams";
+    const default_method = ScrambleMethod["left-right-shuffle-with-unigrams"];
+    const method = options?.method || default_method;
     const n_top_members = 25;
     const max_suggestions = 10;
-    const seed = options?.seed || 0;
+    const random_seed = Number(Number(Math.random() * 10).toFixed(0));
+    const seed = options?.seed || random_seed;
```
lightwalker-eth, line 162: Please see my other comment. As I understand the client SDK should never be generating random numbers itself.

lightwalker-eth (approved):
@FrancoAguzzi Hey great to see these updates! 🚀 Reviewed and approved. I also shared 2 comments with feedback suggestions but recommend you fix those in a separate future PR :+1:

PR #557, opened 2025-01-17
Add returnNameGuardReport for getSecurePrimaryName

djstrong:


notrab:
@djstrong I updated the example site to include the nameguard-react styles, so the double icon should now be fixed.

CleanShot 2025-01-24 at 15 22 26@2x

notrab (approved):

PR #559, opened 2025-01-17
Update changeset config

djstrong:


lightwalker-eth (approved):
@djstrong Great fix, thank you :+1:

PR #561, opened 2025-01-20
feat(nameai.dev): add metadata and favicon

notrab:


apps/nameai.dev/app/sort/page.tsx
```diff
@@ -40,3 +45,22 @@ export default async function SortPage() {
     </div>
   );
 }
+
+const title = "AI Sort";
```
notrab, line 49: @lightwalker-eth feel free to suggest changes here and we can merge once happy.

apps/nameai.dev/app/sort/page.tsx
```diff
@@ -40,3 +45,22 @@ export default async function SortPage() {
...
+const description = "Bring names more likely to be interesting to the top.";
```
notrab, line 50: @lightwalker-eth feel free to suggest changes here and we can merge once happy.

apps/nameai.dev/app/tokenization/page.tsx
```diff
@@ -34,3 +40,32 @@ export default async function TokenizePage(props: TokenizePageProps) {
     </div>
   );
 }
+
+export async function generateMetadata(props: Props): Promise<Metadata> {
+  const searchParams = await props.searchParams;
+  const label = searchParams?.label || "";
+
+  if (!label) return notFound();
```
notrab, line 48: This is actually incorrect. I will change the behaviour here so that the page has different metadata IF there is a label in the query string, otherwise, show a fallback value.

lightwalker-eth (approved):
@notrab Looks good 🚀

PR #569, opened 2025-01-23
Unknown status is not implemented yet

djstrong:


lightwalker-eth (approved):
@djstrong Seems like a good step for now. Thanks :+1:

PR #570, opened 2025-01-23
feat: [SC-26202] ✨  Build Name Details Page draft and most important …

FrancoAguzzi:
…features


apps/namegraph.dev/app/collections/[id]/explore-collection-page.tsx
```diff
@@ -360,14 +362,15 @@ export const ExploreCollectionPage = ({ id }: { id: string }) => {
                                   collectionMembers[
                                     params.page
                                   ]?.suggestions.map((suggestion) => (
-                                    <div
+                                    <Link
+                                      href={`/name/${buildENSName(suggestion.name.replace(" ", "")).name}`}
```
Comment (lightwalker-eth) [inline apps/namegraph.dev/app/collections/[id]/explore-collection-page.tsx:366]: We never want to repeat logic like this in many places. There should be a single helper function definition with this logic inside of it that is called from wherever we need to apply these rules.

apps/namegraph.dev/app/collections/page.tsx
```diff
@@ -347,9 +341,14 @@ export default function ExploreCollectionsPage() {
             <h1 className="text-sm text-gray-500">
               Collection search results for
             </h1>
-            <h2 className="text-3xl font-bold mb-5 leading-9 truncate">
-              {params.search ? params.search : "______"}
-            </h2>
+            <div className="mt-1 mb-3">
+              <Link
+                href={`/name/${buildENSName(params.search.replace(" ", "")).name}`}
```
lightwalker-eth, line 346: Please see my other comment about never repeating logic like this in multiple places.

apps/namegraph.dev/app/layout.tsx
```diff
@@ -84,6 +85,15 @@ export default function RootLayout({
               <div className="flex flex-row items-center justify-between md:gap-5 h-[40px]">
                 <div className="hidden items-center justify-center lg:flex gap-2">
                   <div className="hidden items-center justify-center xl:flex gap-2">
+                    {/* 
+                      This section is currently commented out as
+                      NameGraph API does not support dotted queries.
+                      
+                      e.g. something.anything
```
lightwalker-eth, line 92: @FrancoAguzzi The NameGraph is not built to support "dotted queries" but that doesn't mean you should be commenting this out.

We should be building this feature now. We don't need to wait on any new features in the NameGraph API.

It is the responsibility of the client application (in this case, namegraph.dev) to:
* Convert from a name (or anything else) into a label when forming an API call to the NameGraph API.
* Take the response from the NameGraph API (that always only contains labels) and add whatever suffix is desired based on logic of the client application. Ex: the namegraph.dev website should dynamically determine what suffix should be added based on the selected suffix.

apps/namegraph.dev/components/collections/collection-card.tsx
```diff
@@ -28,24 +28,30 @@ export const CollectionCard = ({
...
           by {collection.owner}
         </p>
         <div className="relative">
           <div className="flex gap-2">
             {collection.top_names.map((tag) => (
-              <span
+              <Link
                 key={tag.namehash}
-                className="max-h-[28px] w-max bg-gray-100 text-sm px-2 py-1 bg-muted rounded-full"
+                href={`/name/${buildENSName(tag.name.replace(" ", "")).name}`}
```
lightwalker-eth, line 45: Please see my other comments about never repeating logic like this.

apps/namegraph.dev/components/preferred-sufix-context.tsx
```diff
@@ -0,0 +1,43 @@
+import { useState, createContext } from "react";
+import { useCookies } from "react-cookie";
+
+export const PREFERRED_SUFIX_KEY = "PREFERRED_SUFIX";
```
lightwalker-eth, line 4: There's a number of places in the code where we are currently writing "sufix". All of those should change to "suffix" with 2 f's.

apps/namegraph.dev/components/sufix-select.tsx
```diff
@@ -0,0 +1,61 @@
...
+} from "./ui/select";
+import { PreferredSufixContext, Sufixes } from "./preferred-sufix-context";
+
+const availableSufixes: Record<
+  Sufixes,
+  { shortName: string; longName: string }
+> = {
+  [Sufixes.ETH]: {
+    shortName: "$ USD",
+    longName: "United States Dollar (USD · $)",
```
lightwalker-eth, line 20: Seems we will want to modify the `shortName` / `longName` logic here?

apps/namegraph.dev/components/sufix-select.tsx
```diff
@@ -0,0 +1,61 @@
+"use client";
+
+import { DEFAULT_PREFERRED_CURRENCY } from "@namehash/ens-utils";
```
lightwalker-eth, line 3: Suggest we remove any import of `DEFAULT_PREFERRED_CURRENCY`

apps/namegraph.dev/app/name/[name]/name-details-page.tsx
```diff
@@ -0,0 +1,804 @@
...
+  totalItems?: Record<NameRelatedCollectionsTabs, number | undefined>;
+}
+
+interface CollectionsData {
+  sort_order: NameGraphSortOrderOptions;
+  related_collections: NameGraphCollection[];
+}
+
+const NameRelatedCollectionsTabs = {
+  ByMembership: "ByMembership",
```
Comment (lightwalker-eth) [inline apps/namegraph.dev/app/name/[name]/name-details-page.tsx:40]: For me this tab isn't loading its count. It should show its count even before I click on anything.

CleanShot 2025-01-26 at 18 26 55

apps/namegraph.dev/app/name/[name]/name-details-page.tsx
```diff
@@ -0,0 +1,804 @@
...
+    >(undefined);
+
+  const [otherCollections, setOtherCollections] = useState<
+    NameGraphCollection[] | undefined
+  >(undefined);
+
+  interface QueryNameRelatedCollectionsParams {
+    activeTab: NameRelatedCollectionsTabs;
+    orderBy?: NameGraphSortOrderOptions;
+    page: number;
```
Comment (lightwalker-eth) [inline apps/namegraph.dev/app/name/[name]/name-details-page.tsx:83]: We need to improve the loading state UX when I change from one page to another.

CleanShot 2025-01-26 at 18 28 28

It shouldn't feel like everything suddenly disappears when I change pages. This is an important UX to optimize.

apps/namegraph.dev/app/name/[name]/name-details-page.tsx
```diff
@@ -0,0 +1,804 @@
...
+                    );
+                  },
+                )}
+              </Tabs>
+            </div>
+            <div className="flex flex-col space-y-2"></div>
+          </div>
+        </div>
+      </div>
+      {otherCollections ? (
```
Comment (lightwalker-eth) [inline apps/namegraph.dev/app/name/[name]/name-details-page.tsx:771]: There's been a misunderstanding. We should not be using `otherCollections` in this way. We should keep a panel that is strictly a list of collections for the "other collections" list. This is NOT a list of names. It is a list of collection cards only.

A separate idea from the "Other Collections" panel is the "Related Names" panel. The implementation of the "Related Names" panel should be using a different NameGraph API call.

I just went to https://namehash.io/name/lightwalker.eth but it looks like we removed the "Related names" panel that was implemented previously. Maybe we just commented that out? Please feel welcome to ask me questions if anything here isn't clear.

apps/namegraph.dev/app/name/[name]/name-details-page.tsx
```diff
@@ -0,0 +1,804 @@
...
+    NameGraphCollection[] | undefined
+  >(undefined);
+
+  interface QueryNameRelatedCollectionsParams {
+    activeTab: NameRelatedCollectionsTabs;
+    orderBy?: NameGraphSortOrderOptions;
+    page: number;
+  }
+
+  const queryCollections = (params: QueryNameRelatedCollectionsParams) => {
```
Comment (lightwalker-eth) [inline apps/namegraph.dev/app/name/[name]/name-details-page.tsx:86]: If I click the "By membership" tab, no results load for me.

CleanShot 2025-01-26 at 18 39 02

lightwalker-eth (changes requested):
@FrancoAguzzi Hey exciting to see these updates. Reviewed and shared feedback :+1:

packages/namegraph-sdk/src/index.ts
```diff
@@ -59,14 +59,15 @@ export class NameGraph {
 
   public suggestionsByCategory(
     label: string,
+    withoutRelatedCollections = false,
```
lightwalker-eth, line 62: This is the wrong strategy. You should instead add an optional param for `maxRelatedCollections` and have it default to `DEFAULT_MAX_RELATED_COLLECTIONS`. You should then define a const for `DEFAULT_MAX_RELATED_COLLECTIONS` and set it equal to `6`.

If you don't want any related collections then you should explicitly pass a value of `0` for `maxRelatedCollections`.

apps/namegraph.dev/components/collections/name-with-default-suffix.tsx
```diff
@@ -0,0 +1,32 @@
+import { useEffect, useState } from "react";
+import {
+  availableSuffixes,
+  DEFAULT_PREFFERED_SUFFIX,
+  PREFERRED_SUFFIX_KEY,
+  Suffixes,
+} from "../suffix-select";
+
+export const NameWithDefaultSuffix = ({ name }: { name: string }) => {
```
lightwalker-eth, line 9: Would it be better to rename both this UI component and this file to just `NameWithCurrentSuffix`? I don't understand why we use "Default" in the name here, since this only uses the default if no suffix is explicitly set.

apps/namegraph.dev/components/collections/collection-card.tsx
```diff
@@ -28,24 +29,30 @@ export const CollectionCard = ({
...
           by {collection.owner}
         </p>
         <div className="relative">
           <div className="flex gap-2">
             {collection.top_labels.map((tag) => (
-              <span
+              <Link
                 key={tag.label}
-                className="max-h-[28px] w-max bg-gray-100 text-sm px-2 py-1 bg-muted rounded-full"
+                href={`/name/${buildENSName(tag.label.replace(" ", "")).name}`}
```
lightwalker-eth, line 46: You need to define a helper function for this and stop putting logic inline in your HTML.

apps/namegraph.dev/app/collections/page.tsx
lightwalker-eth: @FrancoAguzzi Not sure why this comment was marked as resolved? The issue here is not resolved. You need to define a helper function for this and stop putting logic inline in your HTML.

apps/namegraph.dev/app/collections/[id]/explore-collection-page.tsx
```diff
@@ -360,14 +369,17 @@ export const ExploreCollectionPage = ({ id }: { id: string }) => {
                                   collectionMembers[
                                     params.page
                                   ]?.suggestions.map((suggestion) => (
-                                    <div
+                                    <Link
+                                      href={`/name/${getNormalizedAndTrimmedName(suggestion)}`}
```
Comment (lightwalker-eth) [inline apps/namegraph.dev/app/collections/[id]/explore-collection-page.tsx:373]: Suggest that all links to name detail pages always are a fully qualified name. For example, if the suggestion was "franco" and the current suffix is ".eth" then the fully qualified name would be "franco.eth".

When someone loads a name detail page, the route should always include the fully qualified name.

Therefore, on the name detail page, there's no need to use the user's suffix selection to determine what name is being viewed.

However, if the user is on a name detail page and changes the suffix selection, one new side effect should happen: if the user was previously on the name detail page for {label}{previous suffix}, they should be redirected to the name detail page for {label}{new suffix}.

Please let me know if any questions :+1:

One of the goals here is to make it easy for people to share links to specific names, and those links should always open to the correct fully qualified name, no matter who is viewing that link or what cookies might be set.

apps/namegraph.dev/app/collections/[id]/explore-collection-page.tsx
```diff
@@ -273,6 +276,12 @@ export const ExploreCollectionPage = ({ id }: { id: string }) => {
     });
   }, [collection]);
 
+  const getNormalizedAndTrimmedName = (
+    suggestion: NameGraphSuggestion,
+  ): string => {
+    return buildENSName(suggestion.label.replace(" ", "")).name;
```
Comment (lightwalker-eth) [inline apps/namegraph.dev/app/collections/[id]/explore-collection-page.tsx:282]: Surprised to see this. I thought all labels returned by the NameGraph are guaranteed to be ENS Normalized already. This would also mean they should be guaranteed to never have a space, either.

Please feel welcome to confirm this with Krzysztof.

apps/namegraph.dev/app/name/[name]/name-details-page.tsx
```diff
@@ -0,0 +1,882 @@
...
+const NameRelatedCollectionsTabs = {
+  ByMembership: "ByMembership",
+  ByConcept: "ByConcept",
+} as const;
+
+type NameRelatedCollectionsTabs =
+  (typeof NameRelatedCollectionsTabs)[keyof typeof NameRelatedCollectionsTabs];
+
+export const NameDetailsPage = ({ name }: { name: string }) => {
+  const normalizedName = ens_normalize(decodeURI(name));
```
Comment (lightwalker-eth) [inline apps/namegraph.dev/app/name/[name]/name-details-page.tsx:53]: If the value returned by `decodeURI` is not equal to the value returned by `ens_normalize` then we should throw an error that we only show name details pages for names that are already normalized.

lightwalker-eth (changes requested):
@FrancoAguzzi Hey great to see all the updates here :+1: Thanks for your efforts! Reviewed and shared feedback.

apps/namegraph.dev/app/name/[name]/name-details-page.tsx
Comment (FrancoAguzzi) [reply apps/namegraph.dev/app/name/[name]/name-details-page.tsx:53]: @lightwalker-eth isn't it better UX to show invalid name NFT Avatar and to expose related collections and ideated names that are alike the search done?

apps/namegraph.dev/app/collections/[id]/explore-collection-page.tsx
Comment (FrancoAguzzi) [reply apps/namegraph.dev/app/collections/[id]/explore-collection-page.tsx:282]: Confirmed and removed trimming logics from suggested labels

apps/namegraph.dev/app/name/[name]/name-details-page.tsx
Comment (lightwalker-eth) [reply apps/namegraph.dev/app/name/[name]/name-details-page.tsx]: @FrancoAguzzi I'm open to that. Thank you :+1:

apps/namegraph.dev/app/collections/[id]/explore-collection-page.tsx
```diff
@@ -370,12 +364,12 @@ export const ExploreCollectionPage = ({ id }: { id: string }) => {
                                     params.page
                                   ]?.suggestions.map((suggestion) => (
                                     <Link
-                                      href={`/name/${getNormalizedAndTrimmedName(suggestion)}`}
+                                      href={`/name/${buildENSName(suggestion.label).name}`}
```
Comment (lightwalker-eth) [inline apps/namegraph.dev/app/collections/[id]/explore-collection-page.tsx:367]: Please see my other comment about this.

apps/namegraph.dev/app/collections/[id]/explore-collection-page.tsx
```diff
@@ -449,12 +443,12 @@ export const ExploreCollectionPage = ({ id }: { id: string }) => {
                                   {sampledNameIdeas?.map((suggestion) => {
                                     return (
                                       <Link
-                                        href={`/name/${getNormalizedAndTrimmedName(suggestion)}`}
+                                        href={`/name/${buildENSName(suggestion.label).name}`}
```
Comment (lightwalker-eth) [inline apps/namegraph.dev/app/collections/[id]/explore-collection-page.tsx:446]: Please see my other comment about this.

lightwalker-eth (approved):
@FrancoAguzzi Nice! Some pretty huge upgrades here 🚀 Reviewed and shared feedback. Suggest that we merge this in right away and then work on the feedback in a separate future PR :+1: Thanks!

apps/namegraph.dev/components/collections/collection-card.tsx
```diff
@@ -43,10 +43,10 @@ export const CollectionCard = ({
             {collection.top_labels.map((tag) => (
               <Link
                 key={tag.label}
-                href={`/name/${buildENSName(tag.label.replace(" ", "")).name}`}
+                href={`/name/${buildENSName(tag.label).name}`}
```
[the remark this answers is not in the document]
FrancoAguzzi: @lightwalker-eth I've been trying for some time something that is unsuccessful so I will explain the approach that I took which is successful:

1. It is almost impossible to align strings between: title, search bar and URL params. This behaves this way because we need to constantly deal with string splitting if we want these 3 aligned, which is painful once we should deal with searches like "franco.camelo". This example exposes how we cannot simply split the string in its first dot (.) and change its suffix. We rather need to always have clarity of what is the user search term and the selected suffix. 
2. The successful approach I built easily clarifies this separation: in the URL we have "search" param and "tld" param. In the title, we show both the search + tld. Finally, in the search bar we show only the search. This guarantees we have all customizations aligned while code does not need to handle lots of logics regarding where to split user's search term and neither needs to align three places with the same value at each new search. 

Understanding that each one of the title, selected tld and search bar have its own responsibilities is a kick-start run out of the mental-challenge that results in having these three storing the same value (title = "franco.camelo.box" & search bar = "franco.camelo.box" & selected tld = "franco.camelo.box")

Summed to this, storing title = "franco.camelo.box" & search bar = "franco.camelo" & selected tld = ".box" saves us from lots of computation work in user's browser and in our codebase. Here is the exposure of this approach!

Captura de Tela 2025-02-01 às 11 16 01

Do you agree with using this oriented approach?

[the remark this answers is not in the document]
FrancoAguzzi: @lightwalker-eth when it comes to URL sharing, this simplified approach works great!

Check these out!

https://www.loom.com/share/c8328a009a7647c5833806573698ea07?sid=c95ced9b-e74e-4c6a-8bb3-1d2a7b7da8fb

https://www.loom.com/share/bde1efb2a76d48ca9b3feb1edc72f338?sid=8e8a24ba-c9b3-4780-9afa-f1c4a9397411

apps/namegraph.dev/app/collections/page.tsx
```diff
@@ -341,11 +342,11 @@ export default function ExploreCollectionsPage() {
             </h1>
             <div className="mt-1 mb-3">
               <Link
-                href={`/name/${buildENSName(params.search.replace(" ", "")).name}`}
+                href={`/name/${getLabelWithoutWhitespace(params.search)}`}
                 className="!text-3xl font-bold mb-5 leading-9 truncate"
               >
                 {params.search ? (
-                  <NameWithDefaultSuffix name={params.search} />
+                  <NameWithCurrentSuffix name={params.search} />
```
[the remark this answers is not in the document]
FrancoAguzzi: Fixed

PR #582, opened 2025-01-29
Notrab/sc 26233/priority refinements for nameai app launch

notrab:


Y3drk:
Hi @notrab!

I've spotted this throughout several of our pages and wonder if that's done on purpose or is it a bug/misshap. 

Namely in the **footer**, in the `Made with <3 by NameHash Labs` the hyperlink always leads to itself (ex. to namegraph.dev on namegraph.dev) instead of to [namehashlabs.org](namehashlabs.org). 

It also occurs for [nameai.dev](https://nameai.dev/).

namehash_footer_bug

I see that you perform fixes before launch so I though I'll let you know, just in case.

apps/nameai.dev/app/quicksearch.tsx
```diff
@@ -0,0 +1,126 @@
...
+interface QuickSearchProps {
+  handleClick: (example: string) => void;
+}
+
+export function QuickSearch({ handleClick }: QuickSearchProps) {
+  const sliderRef = useRef<HTMLDivElement>(null);
+  const [isAtStart, setIsAtStart] = useState<boolean>(true);
+  const [isAtEnd, setIsAtEnd] = useState<boolean>(false);
+
+  const examples = [
```
lightwalker-eth, line 16: This is working great! Suggest that we make it so that the list of examples can be dynamically injected in depending on the example app. For now let's use the same examples across both example apps. But I'll follow up later with a request to use some different examples on the AI Sort app. The examples here are better optimized for the Tokenization app.

lightwalker-eth (approved):
@notrab Hey this looks awesome! Great work 🚀 Just 1 small suggestion but can be done in a separate PR :+1:

notrab:
> Hi @notrab!
> 
> I've spotted this throughout several of our pages and wonder if that's done on purpose or is it a bug/misshap.
> 
> Namely in the **footer**, in the `Made with <3 by NameHash Labs` the hyperlink always leads to itself (ex. to namegraph.dev on namegraph.dev) instead of to [namehashlabs.org](namehashlabs.org).
> 
> It also occurs for [nameai.dev](https://nameai.dev/).
> 
> namehash_footer_bug
> 
> I see that you perform fixes before launch so I though I'll let you know, just in case.

Hey @Y3drk 

Thanks for noticing that. I copied this footer from another app layout, so I assume this issue is on other apps too. I've pushed a fix to the `nameai.dev` footer. Hopefully we can create a shared component for these at some point.

PR #587, opened 2025-02-02
fix: [SC-26202] :bug: Mount namegraph.dev name details page URL corre…

FrancoAguzzi:
…ctly


lightwalker-eth (approved):
@FrancoAguzzi Looks good 🙌 😄

PR #606, opened 2025-02-05
feat: [sc-26105] Add first/last name tokenizer to NameAI

Byczong:
Story details: https://app.shortcut.com/ps-web3/story/26105

todo:
- [ ] downloader
- [ ] better performance tests
- [ ] limit data (?)

### DIFF
--- a/apps/api.nameai.dev/nameai/all_tokenizer.py
+++ b/apps/api.nameai.dev/nameai/all_tokenizer.py
@@ -109,6 +109,24 @@ class AllTokenizer:
                     continue
                 automaton.add_word(word, word)
 
+        # with open(get_resource_path(self.config.tokenization.person_first_names), encoding='utf-8') as f:
+        #     for line in f:
+        #         word = line.strip().lower()
+        #         if len(word) <= 2:
+        #             continue
+        #         if word in should_be_tokenized:
+        #             continue
+        #         automaton.add_word(word, word)
+
+        # with open(get_resource_path(self.config.tokenization.person_last_names), encoding='utf-8') as f:
+        #     for line in f:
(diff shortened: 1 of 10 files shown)


apps/api.nameai.io/tests/test_tokenizer.py
```diff
@@ -228,8 +239,136 @@ def test_all_tokenizer_quality2():
...
+            for tokens, log_prob in tokenized_labels:
+                if log_prob >= math.log(1e-10):
+                    failures.append(f'Expected very low score for non-name {input_label}, got {log_prob}')
+
+        if failures:
+            print('\n=== PersonNameTokenizer Quality Test Failures [non_names] ===')
+            for failure in failures:
+                print(failure)
+            print(f'\nTotal failures: {len(failures)} out of {len(quality_tests)} test cases')
+            assert False, 'Some tokenization quality tests failed. See above for details.'
```
Goader, line 331: Are these tests simply adding a probability score check compared to those from `test_nlp_inspector.py`?

Goader:
It all seems good to me. One thing that is bothering me is maintaining 2 separate implementations of the same functionality. I would think of possibly substituting this functionality in NameGraph by using the implementation from here? @djstrong

apps/api.nameai.io/tests/test_tokenizer.py
Byczong: In `test_tokenizer.py` separate tokenizers are tested (`AllTokenizer` and `PersonNamesTokenizer`).
In `test_nlp_inspector.py` the tokenizations come from both tokenizers (merging in done in `NLPInspector`).

So these tests are for different levels of the tokenization pipeline.

PR #607, opened 2025-02-05
prevent form submissions

notrab:


lightwalker-eth (approved):
@notrab Thanks for fix!

PR #608, opened 2025-02-05
Rename nameai.dev to nameai.io

djstrong:


apps/api.nameai.dev/pyproject.toml
```diff
@@ -1,10 +1,10 @@
 [tool.poetry]
 name = "nameai"
-version = "0.1.6"
+version = "0.1.7"
 description = "NameHash NameAI API"
 authors = ["NameHash Team <devops@namehash.io>"]
 maintainers = ["NameHash Team <devops@namehash.io>"]
-homepage = "https://www.nameai.dev/"
+homepage = "https://www.nameai.io/"
 repository = "https://github.com/namehash/nameai"
```
lightwalker-eth, line 8: @djstrong  Ah I think we should update this

lightwalker-eth (approved):
@djstrong Really appreciate this! 💪 One detail caught my eye during review but that can be for a different PR 😄

apps/api.nameai.dev/pyproject.toml
djstrong: Fixed here: https://github.com/namehash/namekit/pull/611

PR #611, opened 2025-02-06
Rename nameai.dev to nameai.io

djstrong:


lightwalker-eth (approved):
@djstrong Great work 🚀

PR #612, opened 2025-02-06
Fix query string

djstrong:


lightwalker-eth (approved):
@djstrong Nice 😄 Works from my testing :+1:

PR #613, opened 2025-02-06
feat: [SC-26255] ✨  Integrate NameAi panel into namegraph.dev Name De…

FrancoAguzzi:
…tails page


lightwalker-eth (approved):
@FrancoAguzzi Thanks for these updates!

PR #618, opened 2025-02-12
chore: [sc-26307] Expose selected parameters in NameGraph SDK

Y3drk:
# Expose selected parameters in NameGraph SDK

As discussed on Slack and in the [Google doc](https://docs.google.com/document/d/12dX0Dxb7VBu9tmCXNV8VvVZlWgiYGx242p8l7W_k8Ak/edit?usp=sharing) I exposed several NameGraph API parameters that were previously "hidden" from the SDK user.

## This PR comprises
- [x] The parameters marked green in the doc have been exposed as parameters in their methods ensuring that:
    - [x] They are **optional**, so we ensure backwards compatability
    - [x] Their previous values are now their **defaults**
    - [x] They are of the **exact type** necessary for the API (meaning we parse/round floats to ints) &rarr; *adjustments to this assumption discussed below*
    - [x] The values from clients are **validated** (limited to certain range of values, like only positive or lesser than etc.) to ensure no errors in SDK requests (max/min values are aligned with the python package) &rarr; *adjustments to this assumption discussed below*
    - [x] The **docstrings** of methods are updated accordingly
- [x] The **new test cases** were created to cover these new parameters
- [ ] The **changeset** was included

## To discuss
* [`Validation`] - What if the value is outside of the accepted range? Should the SDK accept it or throw an error? If we choose the former, should it inform the user in any way (logs?) &rarr; ***UPDATE**: For now we let the API handle such errors since it keeps the SDK's code clean and simple*
* [`Validation`] If we ensure that SDK input is of `number` type, will the API ensure that it's converted to `int`/`float` w\e or should the SDK do it? (afaik Pydantic should do it for us, but I might be mistaken) &rarr; ***UPDATE**: I was indeed mistaken and the API throws an error in such cases. We've decided to keep it that way.*


packages/namegraph-sdk/src/index.ts
```diff
@@ -1,502 +1,530 @@
...
+            max_per_type?: number,
+            max_labels_per_related_collection?: number,
+            max_suggestion_per_grouping_category?: number,
+            min_suggestion_per_grouping_category?: number,
+        },
+    ): Promise<NameGraphGroupedByCategoryResponse> {
+        const categoriesQueryConfig: TypedNameGraphGroupingCategoriesParams = {
+            [NameGraphGroupingCategory.related]: {
+                enable_learning_to_rank: DEFAULT_ENABLE_LEARNING_TO_RANK,
+                max_labels_per_related_collection: options?.max_labels_per_related_collection || 10,
```
Byczong, line 98: Should we extract the default values as constants to `utils.ts`? This might be a bad idea if they are completely out of context though...

packages/namegraph-sdk/src/index.ts
```diff
@@ -1,502 +1,530 @@
...
+        return this.rawRequest("fetch_top_collection_members", "POST", payload);
+    }
+
+    /**
+     * Generates scrambled variations of collection tokens.
+     *
+     * @param {string} collection_id - The ID of the collection to scramble tokens from
+     * @param {number} options.seed - Random seed for reproducible scrambling
+     * @param {ScrambleMethod} options.method - Method to use for scrambling. Possible values: "left-right-shuffle", "left-right-shuffle-with-unigrams", "full-shuffle"
+     * @param {ScrambleMethod} options.n_top_members - Number of collection's top members to include in scrambling
```
Byczong, line 208: Here and below the type should be `number`, not `ScrambleMethod` I guess.

packages/namegraph-sdk/src/index.ts
```diff
@@ -1,502 +1,530 @@
...
+        };
+
+        if (method !== "GET") {
+            options.body = JSON.stringify(body);
+        }
+
+        const response = await fetch(url, options);
+
+        if (!response.ok) {
+            throw new NameGraphError(
```
Byczong, line 494: Considering, we aren't validating the requests in here, and we delegate this responsibility to the API code, I think it would be nice to include more information about the error - especially the validation errors (422).

Example validation error response bodies:

(tested using https://api.namegraph.dev/docs)

```json
{
  "detail": [
    {
      "type": "json_invalid",
      "loc": [
        "body",
        260
      ],
      "msg": "JSON decode error",
      "input": {},
      "ctx": {
        "error": "Expecting ',' delimiter"
      }
    }
  ]
}
```

```json
{
  "detail": [
    {
      "type": "greater_than_equal",
      "loc": [
        "body",
        "max_sample_size"
      ],
      "msg": "Input should be greater than or equal to 1",
      "input": -1,
      "ctx": {
        "ge": 1
      },
      "url": "https://errors.pydantic.dev/2.5/v/greater_than_equal"
    }
  ]
}
```

I think we could just display `response["detail"]["msg"]` if it exists.

Byczong (changes requested):
Some of my comments are rather suggestions, but I think the last 2 require changes.

After this is done, LGTM :))

packages/namegraph-sdk/src/index.ts
```diff
@@ -1,502 +1,530 @@
...
+
+        return this.rawRequest(`grouped_by_category`, "POST", payload);
+    }
+
+    /**
+     * Gets name suggestions by category with configurable related collections.
+     *
+     * @param {string} label - The ENS label to get suggestions for, cannot contain dots (.). If enclosed in double quotes assuming label is pre-tokenized
+     * @param {number} maxRelatedCollections - Maximum number of related collections to return (default: 6)
+     * @param {number} options.max_per_type - collection diversity parameter based on collection types, adds penalty to collections with the same type as other collections. If null, then no penalty will be added
```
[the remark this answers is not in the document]
Y3drk: Good idea, I'll go through the docstrings and add the defaults in missing places.

packages/namegraph-sdk/src/index.ts
```diff
@@ -1,502 +1,530 @@
 import fetch from "cross-fetch";
 import {
-  DEFAULT_ENABLE_LEARNING_TO_RANK,
```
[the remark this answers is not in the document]
Y3drk: Agreed. I actually forgot to check if we already use any formatters, but I don't think that's the case.

packages/namegraph-sdk/src/index.ts
```diff
@@ -1,502 +1,530 @@
...
+            },
+        };
+
+        return this.rawRequest(`grouped_by_category`, "POST", payload);
+    }
+
+    /**
+     * Gets name suggestions by category with configurable related collections.
+     *
+     * @param {string} label - The ENS label to get suggestions for, cannot contain dots (.). If enclosed in double quotes assuming label is pre-tokenized
```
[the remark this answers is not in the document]
Y3drk: Honestly when making those I tried to stick to what the python package says and what I know for sure is true, so If you're certain I can definitely add this.

And by `comment` I assume you mean the description for `label`?

packages/namegraph-sdk/src/index.ts
Y3drk: For `top_n_members` definitely. I'll correct it.

packages/namegraph-sdk/src/index.ts
Y3drk: In the previous PR I did that for all defaults that repeat at least once (they appear at least twice in the code), but left the ones that appear only once... I am honestly not sure what to do about them, since before there were a lot more of these so-called `magic numbers`  and I've already reduced the amount of them quite a bit.

I agree that they are kinda confusing but would it actually be better if we extracted them without any explanation?

I plan to leave it as it is. Maybe we can discuss it more sometime in the future?

packages/namegraph-sdk/src/index.ts
Y3drk: I think we could do this as a separate task/PR?

I feel like I don't know enough about how the python package and/or in the other SKDs work to make such decision on my own... Another item for future discussion?

packages/namegraph-sdk/src/index.ts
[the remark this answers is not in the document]
Byczong: Yes, the splitting for pre-tokenized input is done [here](https://github.com/namehash/namegraph/blob/master/namegraph/preprocessor.py#L52).

As we are using default `sep` argument for `str.split` it uses whitespace as separators.

packages/namegraph-sdk/src/index.ts
Byczong: Ok, I only raise this point because of the "to discuss" points concerning validation in the PR's description.

packages/namegraph-sdk/src/index.ts
Y3drk: Also, in some cases (I've noticed it at least for `max_per_type` and `max_related_collections`) the default values differ between methods, making the unification even harder and possibly messier.

djstrong (approved):

PR #633, opened 2025-02-27
fix: [sc-26332] Investigate & Fix Avatar Loading on NHL Landing Page

eduramme:
Story details: https://app.shortcut.com/ps-web3/story/26332


apps/namehashlabs.org/components/3 - organisms/our-supporters/profile-randomization.ts
```diff
@@ -0,0 +1,56 @@
...
+  // Fisher-Yates shuffle
+  for (let i = result.length - 1; i > 0; i--) {
+    const j = Math.floor(random() * (i + 1));
+    [result[i], result[j]] = [result[j], result[i]];
+  }
+
+  return result;
+}
+
+export function getRandomProfiles(
```
lightwalker-eth, line 23: Please update this function so that if the length of `profiles` is <= `maxCount` that you don't randomize the profiles at all and instead you just return back the provided profiles in their original ordering.

The goal of this change is to make it so that all categories with 4 or fewer supporters always have their supports appear in the same order as defined in `ens-profiles.ts`.

lightwalker-eth (changes requested):
@eduramme Hey overall looks great! Just 1 change request :+1:

lightwalker-eth (approved):
@eduramme Looks good, thank you!

PR #644, opened 2025-03-16
Improve NameGraph app

FrancoAguzzi:


apps/namegraph.dev/.env
```diff
@@ -0,0 +1 @@
+NEXT_PUBLIC_ALCHEMY_URL=https://eth-mainnet.g.alchemy.com/v2/e9uQaE4xz5SO3kWF2noImxFj6lrqxMFl
```
lightwalker-eth, line 1: We should remove this file from Git. Also suggest to turn shut down this app in Alchemy since this API key has now been leaked to GitHub.

lightwalker-eth (approved):
@FrancoAguzzi Hey super updates here 🚀 Nice work 🙌

apps/namegraph.dev/.env
FrancoAguzzi: But thank you, the API key was deleted 👍🏼

PR #704, opened 2025-08-12
chore: add ensawards.org to allowed origins

Y3drk:
Addittion of ensawards.org to request origins accepted by the page's CORS policy


Y3drk:
> ### ⚠️ No Changeset found

Since the changes regard only the namehashlabs.org landing page I didn't include changeset on purpose, since it's not one of our packages

